PR #4, opened 2022-12-07
Add Zod schemas, validate data from NAC API

floatplane:
ptal @stevekuznetsov 

- This seems to be generally working in a local build, but I'm sure it's possible to trigger errors - I haven't clicked on every zone of every avalanche center 😅. Setting up Sentry will help for tracking down and squashing validation bugs
- types/nationalAvalancheCenter/schemas.ts has the meat of the changes


stevekuznetsov:
@floatplane do you want to work out the kinks with the preview action in this PR, even though it's not really related to the changes you're making?

types/nationalAvalancheCenter/index.ts
```diff
@@ -0,0 +1,43 @@
+export * from './enums';
+export * from './schemas';
+
+// TODO(brian) move these two methods somewhere else, they're not types-related
+import {DangerLevel} from './enums';
+
+export const dangerIcon = (level: DangerLevel): string => {
```
stevekuznetsov, line 7: IIRC these got embedded as assets into the repo - maybe can delete it

app.config.ts
```diff
@@ -1,15 +1,21 @@
 import {ExpoConfig, ConfigContext} from '@expo/config';
 
-export default ({config}: ConfigContext): ExpoConfig => {
```
stevekuznetsov, line 3: What kind of PR check would have caught this? Surprised the build didn't barf ...

floatplane: I just ran `yarn tsc` at the root to see what the Typescript compiler had to say. I'm surprised that the build didn't barf too! 

This seems to be by "design"? 

https://stackoverflow.com/questions/58459262/why-does-my-react-native-app-build-successfully-despite-typescript-compiler-erro

floatplane:
> @floatplane do you want to work out the kinks with the preview action in this PR, even though it's not really related to the changes you're making?

I will try!

floatplane:
I kind of can't believe it was that simple 😄

PR #5, opened 2022-12-08
Create a script for CI checks, run Typescript compilation in CI

floatplane:
Update the CI action to check for Typescript errors


stevekuznetsov:
Any specific reason to prefer one check to many? We don't have a quota for checks, and IIRC they run in containers so the overhead is minimal - could be nice to see which of the N things we're checking for has failed

floatplane:
> Any specific reason to prefer one check to many? We don't have a quota for checks, and IIRC they run in containers so the overhead is minimal - could be nice to see which of the N things we're checking for has failed

I think it's nice to be able to run a single check from the local machine and know that it's the same command that CI uses. Debugging actions seems to be a bit of a PITA so I'd generally be in favor of keeping their invocations simple. But I can split these into chunks (once I figure out why the `prettier` and `tsc` packages don't seem to be installed 🤷‍♂️)

floatplane:
^ we also shouldn't be running the preview/publish action if the CI action fails, so have to fix the dependencies there

stevekuznetsov:
Gotcha, I was moreso envisioning a `yarn` command for each action, but either sounds good. When we start getting into crazy conditional shell scripts we can reconsider ;)

floatplane:
well, looks like i got to accidentally test my dependency change 🙈. time to run prettier!

.github/workflows/publish.yaml
```diff
@@ -4,7 +4,12 @@ on:
     branches:
       - main
 jobs:
+  ci:
+    uses: ./.github/workflows/ci.yaml # use the callable ci job to run CI checks
+
   build:
+    # CI ideally *shouldn't* fail post-merge, but you could imagine criss-crossing PRs that break tests
```
stevekuznetsov, line 11: IIRC there's a setting in the repo somewhere that requires tests not only to pass for each PR against the branch they target, but also for them to test *against the current HEAD* - we can turn that on for a saner experience

stevekuznetsov:
LGTM - merge when you're happy with it!

PR #14, opened 2022-12-14
Setup ESLint and fix all the warnings and errors so far 

floatplane:
Most common warnings:
- unused imports
- using `let` instead of `const`
- using explicit types unnecessarily (i.e. `const num: number = 4`)

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -22,6 +22,8 @@ import {TelemetryStationMap} from './components/TelemetryStationMap';
 import {TelemetryStationData} from './components/TelemetryStationData';
 
 Sentry.init({
+  // we're overwriting a field that was previously defined in app.json, so we know it's non-null:
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   dsn: Constants.expoConfig.extra!.sentry_dsn,
   enableInExpoDevelopment: true,
   debug: true, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
@@ -58,7 +60,7 @@ const AvalancheCenterTabScreen = ({route}: AvalancheCenterProps) => {
         name={'avalancheCenterStack'}
         component={AvalancheCenterStackScreen}
         initialParams={{center_id: center_id, date: date}}
-        options={({route}) => ({title: center_id})}
+        options={() => ({title: center_id})}
       />
       <AvalancheCenterTab.Screen
         name={'avalancheCenterTelemetryStack'}
@@ -183,10 +185,12 @@ type StackParamList = {
   };
 };
 
-const Stack = createNativeStackNavigator<StackParamList>();
-
+// TODO(brian): dig into this. This is somehow suppressing typescript errors in
+// other files - if you comment it out, compilation fails.
 declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace ReactNavigation {
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
     interface RootParamList extends StackParamList {}
   }
 }
@@ -234,6 +238,7 @@ const App = () => {
     );
   } catch (error) {
     Sentry.Native.captureException(error);
+    throw error;
   }
 };
 
--- a/app.config.ts
+++ b/app.config.ts
@@ -5,6 +5,9 @@ export default ({config}: ConfigContext): Partial<ExpoConfig> => {
   // we keep in the environment and out of code. For cloud builds, the secrets
   // are stored in Expo and supplied via process.env. For local builds, the
   // secrets can be stored in a .env file and loaded via direnv.
+
+  // we're overwriting fields that were previously defined in app.json, so we know they're non-null:
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
   config.ios!.config!.googleMapsApiKey = process.env.IOS_GOOGLE_MAPS_API_KEY;
   config.android!.config!.googleMaps!.apiKey = process.env.ANDROID_GOOGLE_MAPS_API_KEY;
   config.hooks!.postPublish![0]!.config!.authToken = process.env.SENTRY_API_TOKEN;
--- a/components/AvalancheCenterForecastZonePolygons.tsx
+++ b/components/AvalancheCenterForecastZonePolygons.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import {ActivityIndicator, Text, View} from 'react-native';
 import {Region} from 'react-native-maps';
 
 import {useMapLayer} from '../hooks/useMapLayer';
@@ -13,7 +12,7 @@ export interface AvalancheCenterForecastZonePolygonsProps {
 }
 
 export const AvalancheCenterForecastZonePolygons: React.FunctionComponent<AvalancheCenterForecastZonePolygonsProps> = ({center_id, date, setRegion}) => {
-  const {isLoading, isError, data: mapLayer, error} = useMapLayer(center_id);
+  const {isLoading, isError, data: mapLayer /*, error */} = useMapLayer(center_id);
   if (isLoading) {
     // TODO(skuznets): without the zones, we don't know where on the map to put these loading/error elements ... ?
     return <></>;
--- a/components/AvalancheDangerIcon.tsx
+++ b/components/AvalancheDangerIcon.tsx
@@ -14,6 +14,7 @@ interface size {
 }
 
 export const AvalancheDangerIcon: React.FunctionComponent<AvalancheDangerIconProps> = ({style, level}: AvalancheDangerIconProps) => {
+  /* eslint-disable @typescript-eslint/no-var-requires */
   const sizes: Record<DangerLevel, size> = {
     [DangerLevel.GeneralInformation]: Image.resolveAssetSource(require('../assets/danger-icons/0.png')),
     [DangerLevel.None]: Image.resolveAssetSource(require('../assets/danger-icons/0.png')),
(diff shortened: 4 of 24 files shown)


components/AvalancheProblemLikelihoodLine.tsx
```diff
@@ -16,19 +16,14 @@ export const likelihoodText = (input: AvalancheProblemLikelihood): string => {
 };
 
 const likelihoodToRange = (likelihood: AvalancheProblemLikelihood): SeverityNumberLineRange => {
-  let index: number = 4;
-  switch (likelihood) {
-    case AvalancheProblemLikelihood.AlmostCertain:
-      index -= 1;
```
floatplane, line 22: eslint didn't like fall-through here (most lint tools don't in my experience 😄)

App.tsx
```diff
@@ -183,10 +185,12 @@ type StackParamList = {
   };
 };
 
-const Stack = createNativeStackNavigator<StackParamList>();
-
+// TODO(brian): dig into this. This is somehow suppressing typescript errors in
+// other files - if you comment it out, compilation fails.
```
stevekuznetsov, line 189: Yeah the navigation bits are black magic as far as I'm concerned

app.config.ts
```diff
@@ -5,6 +5,9 @@ export default ({config}: ConfigContext): Partial<ExpoConfig> => {
   // we keep in the environment and out of code. For cloud builds, the secrets
   // are stored in Expo and supplied via process.env. For local builds, the
   // secrets can be stored in a .env file and loaded via direnv.
+
+  // we're overwriting fields that were previously defined in app.json, so we know they're non-null:
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
```
stevekuznetsov, line 10: What null checks are we missing by the way? I thought the `!` annotation in the lines below was doing the assertion for us?

components/AvalancheCenterForecastZonePolygons.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import {ActivityIndicator, Text, View} from 'react-native';
```
stevekuznetsov, line 3: Nice! Does the check also automatically remove these for us?

components/AvalancheCenterForecastZonePolygons.tsx
```diff
@@ -13,7 +12,7 @@ export interface AvalancheCenterForecastZonePolygonsProps {
 }
 
 export const AvalancheCenterForecastZonePolygons: React.FunctionComponent<AvalancheCenterForecastZonePolygonsProps> = ({center_id, date, setRegion}) => {
-  const {isLoading, isError, data: mapLayer, error} = useMapLayer(center_id);
+  const {isLoading, isError, data: mapLayer /*, error */} = useMapLayer(center_id);
```
stevekuznetsov, line 15: Yikes - maybe let's log these at a minimum or send them to Sentry - this is one of those weird ones that can't really be shown since there's no place to put them ...

components/AvalancheDangerIcon.tsx
```diff
@@ -47,7 +48,8 @@ export const AvalancheDangerIcon: React.FunctionComponent<AvalancheDangerIconPro
       return <Image style={s} source={require('../assets/danger-icons/5.png')} />;
     },
   };
-  let actualStyle: ImageStyle = {...style};
+  /* eslint-enable @typescript-eslint/no-var-requires */
+  const actualStyle: ImageStyle = {...style};
```
stevekuznetsov, line 52: What does it mean for this to be `const` if we're setting sub-fields of the thing in the following lines?

components/AvalancheForecast.tsx
```diff
@@ -42,7 +42,7 @@ export const AvalancheForecast: React.FunctionComponent<AvalancheForecastProps>
         navigation.setOptions({title: thisZone.name});
       }
     }
-  }, [forecast]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [forecast, forecast_zone_id, navigation]);
```
stevekuznetsov, line 45: Please let's not do this unless we have a good reason - these are super tricky and the lint rule for this particular thing in my experience is really not correct.

components/AvalancheProblemLikelihoodLine.tsx
stevekuznetsov: Wild. Poorly supported feature of the language?

components/AvalancheForecast.tsx
stevekuznetsov: I guess to expand on these - I spent time carefully choosing what went into the dependency array since it's so critical for correctness & performance and at least in the past I've found that following this linter is a good way to get infinite re-renders and ill defined behavior. I'm of course not an expert but if we change these I would just hope that the change is motivated by carefully considering the hook vs a linter rule applied broadly. Perhaps in some of these cases `useCallback` or `useMemo` might be needed as well for effects capturing functions, and as in this case changing `forecast_zone_id` would cause a full re-render for the component anyway.

components/AvalancheProblemLikelihoodLine.tsx
floatplane: I think accidental fallthrough in switch statements has been a concern since I first learned C, and that was back in the K&R days 👴. Like yeah, it's valid but it's often not what people intend! In this case it was easy enough to rewrite it.

Looks like C++17 lets you annotate this explicitly now, cool! https://en.cppreference.com/w/cpp/language/attributes/fallthrough

app.config.ts
floatplane: My read is that the ESLint warning here is for the use of `!`, because in an ideal world you shouldn't need it (either your types should guarantee that the thing isn't null, or you should have a conditional that's checking that it's not null).

But this is a weird case, given that the structure of the config is determined by `app.json`, so I suppressed the error.

components/AvalancheCenterForecastZonePolygons.tsx
floatplane: It does not :(. It's possible that an editor plugin might!

components/AvalancheDangerIcon.tsx
floatplane: `const` means that you can't assign `actualStyle` to a _different_ object, but you can still modify fields on the object. 

typescript also has a [`readonly` keyword for properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#readonly-properties), but that doesn't apply here.

and of course you can use `Object.freeze()` to create an object that's truly unmodifiable, if you need something more than `const`.

components/AvalancheForecast.tsx
floatplane: Hmm, I'm pretty surprised to hear that - we had this rule enabled on a fairly large React codebase at Stripe and I don't recall seeing it suppressed at all! I'm sure there can be weird cases, but if the callback is firing more than it should be, then it might mean the component is rendering more than it should be as well, since these inputs are ultimately derived from props & state.

I can leave these in for now and create issues to look more deeply at each of them later, but I personally would rather have the warning than not ([Dan Abramov agrees FWIW](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528) 😄 ).

components/AvalancheProblemLikelihoodLine.tsx
stevekuznetsov: I'm used to explicit fallthrough in Go!

app.config.ts
stevekuznetsov: OK, that makes sense - thanks for explaining

components/AvalancheForecast.tsx
stevekuznetsov: Sounds good! Happy to remove them if we're pretty certain it's a good thing. I vaguely remember struggling with cyclical re-render because I followed the linter rules in the summer, but it's more than likely I was just messing something else up entirely. If you have not had poor experiences with this in the past, it's more likely just me being a React noob - when you made the changes in this PR did anything jump out at you as obviously weird for this lint rule?

components/AvalancheProblemLikelihoodLine.tsx
floatplane: ah cool, I haven't learned to speak Go yet but it's on my list 😄

components/AvalancheForecast.tsx
floatplane: I'd say this one felt unsafe to me, thinking through how the callback would behave if `forecast_zone_id` or the `navigation` object changed on render pass 2, and then `forecast` changed on render pass 3. At that point the callback would execute with stale references in the closure and probably not do what you want.

I think in practice that would never happen and it's probably a wash, but generally I think it's great to have the warning. It makes you consider everything that the callback is closing over.

components/AvalancheCenterForecastZonePolygons.tsx
floatplane: Created https://github.com/stevekuznetsov/avalanche-forecast/issues/15 to look into this

floatplane: (we are currently logging validation errors to Sentry but network errors would also show up on this path, and we probably don't want those to go to Sentry)

components/AvalancheForecast.tsx
floatplane: Opened https://github.com/stevekuznetsov/avalanche-forecast/issues/16 to track following up on this

floatplane:
@stevekuznetsov LMK if you feel OK about merging this in its current state!

components/AvalancheCenterForecastZonePolygons.tsx
stevekuznetsov: Sounds good!

components/AvalancheForecast.tsx
stevekuznetsov: Sounds good!

stevekuznetsov:
Merge away!

PR #18, opened 2022-12-16
Fix safe area issues

floatplane:
better:
image
still might be doubling up on headers but we'll fix that shortly
image


floatplane:
I'm going to close this PR, I have a better fix in another branch

PR #19, opened 2022-12-16
Home

floatplane:
- stub in a new home view
- avalanche center picker is still alive, in a `__DEV__`-only tab called "Debug"
- forecast viewing is *broken* - I'll be fast-following with that
- there seems to be some extra padding at the bottom of the tab navigator...will debug that later as well 🤔 

image

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -1,6 +1,6 @@
 import React, {useEffect, useState} from 'react';
 
-import {AppStateStatus, DevSettings, Platform, StyleSheet, Text} from 'react-native';
+import {AppStateStatus, DevSettings, Platform, StyleSheet, Text, View} from 'react-native';
 import {NavigationContainer} from '@react-navigation/native';
 import {createNativeStackNavigator, NativeStackScreenProps} from '@react-navigation/native-stack';
 import {createBottomTabNavigator} from '@react-navigation/bottom-tabs';
@@ -90,9 +90,9 @@ type AvalancheCenterProps = NativeStackScreenProps<HomeStackParamList, 'avalanch
 const MapScreen = ({route}: AvalancheCenterProps) => {
   const {center_id, date} = route.params;
   return (
-    <SafeAreaView style={styles.container}>
+    <View style={{...styles.container}}>
       <AvalancheForecastZoneMap centers={[center_id]} date={date} />
-    </SafeAreaView>
+    </View>
   );
 };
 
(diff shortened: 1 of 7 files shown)


App.tsx
```diff
@@ -32,169 +33,118 @@ Sentry.init({
...
-      <SelectorStack.Screen name="avalancheCenterSelection" component={AvalancheCenterSelectorScreen} options={{title: 'Select an Avalanche Center'}} />
-      <SelectorStack.Screen name="avalancheCenterHome" component={AvalancheCenterTabScreen} options={({route}) => ({title: route.params.center_id})} />
-    </SelectorStack.Navigator>
-  );
-};
+// TODO(brian): do we need this? I'm guessing this kept things working while
+// developing in the offseason. Also TBD whether this should be part of the
+// route; are there use cases for supporting forecasts from different dates than
+// "today"?
+const defaultDate = formatISO(new Date('2022-03-01'));
```
floatplane, line 43: @stevekuznetsov LMK if you have thoughts here!

App.tsx
```diff
@@ -32,169 +33,118 @@ Sentry.init({
...
-
-// TODO(brian): dig into this. This is somehow suppressing typescript errors in
-// other files - if you comment it out, compilation fails.
-declare global {
-  // eslint-disable-next-line @typescript-eslint/no-namespace
-  namespace ReactNavigation {
-    // eslint-disable-next-line @typescript-eslint/no-empty-interface
-    interface RootParamList extends StackParamList {}
-  }
-}
```
floatplane, line 197: figured out the magic here - now using explicitly exported `HomeStackParamList` from `routes.ts` and then calling `useNavigation<HomeStackParamList>`

App.tsx
```diff
@@ -236,7 +220,31 @@ const App = () => {
         <QueryClientProvider client={queryClient}>
           <SafeAreaProvider>
             <NavigationContainer>
-              <AvalancheCenterSelectionScreen />
+              <SafeAreaView style={styles.container}>
```
floatplane, line 223: AFAICT we only need *one* SafeAreaView at the top-level

floatplane:
I'm going to merge and follow up with any requested changes

App.tsx
stevekuznetsov: Yeah - I was hoping that we would allow a date picker in some sort of view - need design input but a calendar should suffice for now. Default to "today".

components/AvalancheCenterSelector.tsx
```diff
@@ -68,24 +68,29 @@ const AvalancheCenterCard: React.FunctionComponent<AvalancheCenterCardProps> = (
...
-}
-
-export const AvalancheCenterSelector: React.FunctionComponent<AvalancheCenterSelectorProps> = ({date}: AvalancheCenterSelectorProps) => {
+export const AvalancheCenterSelector = () => {
   return (
     <SectionList
       style={styles.container}
       sections={center_idsByType}
       keyExtractor={item => item}
-      renderItem={({item}) => <AvalancheCenterCard center_id={item} date={date} />}
```
stevekuznetsov, line 88: Why'd we lose the date?

floatplane: After selecting a center, we now clear the nav stack and go back to the very first view (`Home`) which references the top-level `defaultDate`

PR #22, opened 2022-12-20
*: add observations client, simple pages to view them

stevekuznetsov:
CONTRIBUTING.md: record what's needed for a dev env

Without the Sentry data, a bunch of errors pop up on every app open.

---

App.tsx: use shared state for center, date

This way, choosing a new center in the dev picker influences where
observations, weather data, etc are loaded from.

---

types/nac: update a field to nullable

We don't use this yet, but apparently it's nullable.

---

*: add observations client, simple pages to view them

This commit adds in:

 - a GraphQL schema dumped from introspecting the endpoints the NAC
   provides
 - auto-generated TypeScript types for the above scehma
 - semi-generated axios- and react-query-based clients
 - a simple scroll-view for observations
 - a page to inspect a specific observation

---

cc @floatplane 

The views are very rudimentary to the point of almost being useless, was mostly just focused on the plumbing here.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -204,7 +204,7 @@ const App = () => {
                   initialRouteName="Home"
                   screenOptions={({route}) => ({
                     headerShown: false,
-                    tabBarIcon: ({focused, color, size}) => {
+                    tabBarIcon: ({color, size}) => {
                       if (route.name === 'Home') {
                         return <AntDesign name="search1" size={size} color={color} />;
                       } else if (route.name === 'Observations') {
--- a/App.tsx
+++ b/App.tsx
@@ -22,6 +22,8 @@ import {useOnlineManager} from './hooks/useOnlineManager';
 import {TelemetryStationMap} from './components/TelemetryStationMap';
 import {TelemetryStationData} from './components/TelemetryStationData';
 import {TabNavigatorParamList, HomeStackParamList} from './routes';
+import {Observations} from './components/Observations';
+import {Observation} from './components/Observation';
 
 Sentry.init({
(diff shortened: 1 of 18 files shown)


App.tsx
```diff
@@ -219,29 +199,23 @@ const App = () => {
           <SafeAreaProvider>
             <NavigationContainer>
               <SafeAreaView style={styles.container}>
-                <Tab.Navigator initialRouteName="Home" screenOptions={{headerShown: false}}>
```
stevekuznetsov, line 222: I'm now realizing if we had the center picker navigate to a view that has this whole thing as a child, then we are going to have a consistent route data passing paradigm .. ?

stevekuznetsov: Hm no I can't imagine adding new tabs to one global navigator with the same name but different parameters would be a good idea.

stevekuznetsov:
Looks like Brian's out until January, I will just merge this in and we can iterate after the new year.

hooks/observations-fetcher.ts
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import axios from 'axios';
+
+import {ClientContext, ClientProps} from '../clientContext';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const useFetch = <TData, TVariables>(query: string, options?: RequestInit['headers']): ((variables?: TVariables) => Promise<TData>) => {
```
floatplane, line 8: @stevekuznetsov can we rename this file to `useFetch.ts`?

Issue #23, opened 2022-12-22
Fail Gracefully When Zod Errors Encountered


stevekuznetsov:
If at all possible, we should try to fail more gracefully when a schema does not match what we expected. Today it seems like we just don't show the page at all.

cc @floatplane

stevekuznetsov:
I believe now this will lead to the error view "oh no!" if the `useQuery` fails.

PR #26, opened 2022-12-22
observations: improve cards for overviews

stevekuznetsov:
types: loosen zod restrictions

Even more things we see in the wild... Might want to write a script that
validates every center, map, forecast etc in a batch.

---

observations: improve cards for overviews

---


types/nationalAvalancheCenter/schemas.ts
```diff
@@ -53,18 +53,20 @@ export const forecastPeriodSchema = z.nativeEnum(ForecastPeriod);
...
-  thumbnail: z.string(),
-});
+export const mediaLinksSchema = z
+  .object({
+    large: z.string().optional(),
+    medium: z.string().optional(),
+    original: z.string().optional(),
+    thumbnail: z.string().optional(),
+  })
+  .or(z.string()); // when this field is not populated, it's an empty string, not a null
```
stevekuznetsov, line 63: @floatplane would like your input on how to best handle this ... what a mess.

floatplane: My current best thought is to fix up issues post-fetch, but pre-validation: https://github.com/stevekuznetsov/avalanche-forecast/blob/c27a6dee43269e0097a3526d5b8678f4796773dc/hooks/useAvalancheForecast.ts#L23-L33

For this particular case, I guess I'd look for empty strings and turn them into nulls to make it easier to deal with?

floatplane:
> Even more things we see in the wild... Might want to write a script that
validates every center, map, forecast etc in a batch.

Been thinking that we should do something like:
1. script to download a day's worth of data from all avalanche centers - run this once to get an initial set of test inputs
1. write automated tests to validate that the data passes validation
1. update test data as changes are rolled out on the NAC side

Setting up some automated testing and having tests run on PRs would be more comforting :)

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -47,24 +47,46 @@ export const avalancheProblemLikelihoodSchema = z.nativeEnum(AvalancheProblemLik
...
+        code: z.ZodIssueCode.invalid_type,
+        message: 'Avalanche problem size must be numeric.',
+        expected: 'number',
+        received: 'string',
+      });
+      return z.NEVER;
+    }
+    return parsed;
+  })
+  .transform(val => numberToProblemSize(val));
```
floatplane, line 69: ah, this is slick - did not know about `transform`

stevekuznetsov:
@floatplane looked briefly at doing that this morning but using React hooks inside of a "simple" JS script run by `yarn` seemed impossible, and I was hoping to use the automation to get confidence in our hooks ...

hooks/useAvalancheForecast.ts
```diff
@@ -44,8 +32,9 @@ export const useAvalancheForecast = (center_id: string, forecast_zone_id: number
             url,
           },
         });
+      } else {
+        return parseResult.data;
       }
-      return productSchema.parse(data);
```
floatplane, line 48: I think we should be returning `parseResult.error` from the first branch of the `if` statement

types/nationalAvalancheCenter/schemas.ts
floatplane: ignore this, I didn't realize you had already gone down a different road with `transform`

hooks/useAvalancheForecast.ts
stevekuznetsov: I don't know if it's just my editor, but boy are these types fun:

```
TS2345: Argument of type '() => Promise<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; } | ZodError<...>>' is not assignable to parameter of type 'QueryFunction<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; }, QueryKey>'.   Type 'Promise<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; } | ZodError<...>>' is not assignable to type '{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; } | Promise<...>'.     Type 'Promise<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; } | ZodError<...>>' is not assignable to type 'Promise<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; }>'.       Type '{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; } | ZodError<...>' is not assignable to type '{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; }'.         Type 'ZodError<{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: string | { large?: string; medium?: string; original?: string; thumbnail?: string; }; caption?: string; }[]; ... 14 more ...; forecast_zone?: { ...; }[]; }>' has no properties in common with type '{ status?: ProductStatus.Published; id?: number; media?: { type?: MediaType; id?: number; url?: any; caption?: string; }[]; expires_time?: string; published_time?: string; product_type?: ProductType; ... 11 more ...; forecast_zone?: { ...; }[]; }'.
```

Need to find the right type to parameterize the `useQuery`?

stevekuznetsov: Ah, I see - it's the difference between returning the error and throwing it. The other call (`const {data} = await axios.get(url);`) throws an `AxiosError` - and we have error-handling code in the calling components as well, so WDYT about not using `safeParse` here at all?

I think that would help us with #23 - perhaps we can add helpers for the Sentry bits since we want to associate extra data with these exceptions?

floatplane:
> @floatplane looked briefly at doing that this morning but using React hooks inside of a "simple" JS script run by yarn seemed impossible, and I was hoping to use the automation to get confidence in our hooks ...

- I think it'd be better to use a real test framework like Jest (https://docs.expo.dev/guides/testing-with-jest/) as opposed to rolling a one-off JS script
- once you've got tests set up, you can always test the React component that uses the hook (does it render as expected?) and not get lost in lower level details of "how do I test a hook directly?". or you can create a simple test component that uses the hook internally

want to divide and conquer on this?

hooks/useAvalancheForecast.ts
stevekuznetsov: NVM - we can just `throw` here and keep the flows the same.

stevekuznetsov:
@floatplane sure - I poked around those test frameworks and walked away from them not really knowing how to make them useful for us. Some thoughts:

- I would not want someone's PR to be blocked because some errant data got put onto the NAC platform async
- I had wanted to write tests for the hook itself to make them simpler - two cases:
  - if we don't control the data coming in, what kind of useful test can we write at the component level other than "it rendered"?
  - if we _do_ control the data coming in, we risk not actually testing the API, so I guess in my mind I would have had two sets of tests (does the API conform to expectations, do our components work)

I had played with the [React Native testing library](https://github.com/callstack/react-native-testing-library) but before we had real designs to test end-to-end, I didn't want to add tests for behavior we'd be removing - would love your input on which parts of the design we think are good targets for tests.

types/nationalAvalancheCenter/schemas.ts
stevekuznetsov: Sorry! Forgot to mark this resolved

floatplane:
lgtm! :shipit:

PR #28, opened 2022-12-26
Make some of the app startup a bit safer

floatplane:
- Only init Sentry if it's configured correctly; otherwise warn
- Only add items to DevSettings if the package can be loaded (resolves https://sentry.io/organizations/nwac/issues/3823627003/?project=4504317393436672&query=is%3Aunresolved&referrer=issue-stream)


floatplane:
@stevekuznetsov I hope this cleans up some of the weirdness you observed when you started working on the app last week

PR #31, opened 2022-12-27
Resolve imports from top of source tree, remove many relative imports

floatplane:
This should make moving code around easier.


floatplane:
@stevekuznetsov I want to give you a heads up on this one and make sure you're cool with it. this image shows the differences:
- references in the current directory remain relative, so it's easy to move around a directory of files as a package
- references outside the current directory are changed to be rooted at the top of the source tree

image

stevekuznetsov:
I'll be honest I just let the IDE handle these (and they get auto-collapsed in my editor even) so let's go with whatever you think is best.

PR #32, opened 2022-12-28
Start styling the home screen

floatplane:
The "Danger Scale" isn't quite finished (still needs to be moved), but it's close.

Screenshot_20221228-124922_avalanche-forecast

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -5,14 +5,13 @@ import MapView, {Region} from 'react-native-maps';
 import {useNavigation} from '@react-navigation/native';
 import {FontAwesome5} from '@expo/vector-icons';
 
-import Color from 'color';
 import {parseISO} from 'date-fns';
 
 import {AvalancheDangerForecast, DangerLevel, Feature, ForecastPeriod} from 'types/nationalAvalancheCenter';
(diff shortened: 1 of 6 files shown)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -82,21 +81,39 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
           {isReady && centers.map(center_id => <AvalancheCenterForecastZonePolygons key={center_id} center_id={center_id} setRegion={setRegion} date={date} />)}
         </MapView>
         <View style={styles.legend}>
-          <Text style={styles.legendTitle}>Avalanche Danger Scale</Text>
+          <View style={styles.legendHeader}>
+            <Text style={styles.legendTitle}>Danger Scale</Text>
+            <FontAwesome5 name="info-circle" size={16} color="blue" />
```
stevekuznetsov, line 86: Just to confirm I'm reading this right - as of this commit we don't do anything when someone interacts with this thing, right?

stevekuznetsov (commented):
Looks great! Just one question

components/AvalancheForecastZoneMap.tsx
floatplane: Correct, behavior hasn't been specified yet ([figma discussion])(https://www.figma.com/file/Bai2w7x0vUakL5WRjPGRg7?node-id=94:94385#336437434)!

PR #33, opened 2022-12-29
Replace Ant Design with NativeBase

floatplane:
ptal @stevekuznetsov 

_(note: the diff in App.tsx looks huge, but it's not - view it with whitespace changes ignored)_

I did not spend a lot of time rebuilding the Ant Design Card here, I'm sure I could get even closer to perfectly matching the Ant Design card with a bit more effort. That said, there are other things that are actually better now, like having the HTML content left-aligned with everything else!

I think the NativeBase styling props and layout classes are actually pretty nice to use and result in fairly readable code. And things like colors (`light.400`) and text sizes (`sm`) are referenced to the theme so you can change it in a single spot.

https://github.com/stevekuznetsov/avalanche-forecast/blob/98bb891bfb00b6c4ff2e62a5c559b6dfcfa291bb/components/Observations.tsx#L93-L110

before | after
--- | ---
image | image

### DIFF
--- a/components/Observations.tsx
+++ b/components/Observations.tsx
@@ -4,7 +4,7 @@ import {compareDesc, format, parseISO, sub} from 'date-fns';
 import {ActivityIndicator, View, Text, FlatList, TouchableOpacity, useWindowDimensions} from 'react-native';
 import {HomeStackNavigationProps} from 'routes';
 import {useNavigation} from '@react-navigation/native';
-import {Card, WhiteSpace, WingBlank} from '@ant-design/react-native';
+import {Box, Divider, VStack, HStack, Text as NBText, Heading} from 'native-base';
 import {useMapLayer} from 'hooks/useMapLayer';
 import {geoContains} from 'd3-geo';
 import RenderHTML from 'react-native-render-html';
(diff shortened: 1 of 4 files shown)


floatplane:
cc @yuliadub

components/Observations.tsx
```diff
@@ -83,26 +83,32 @@ export const ObservationSummaryCard: React.FunctionComponent<{
...
+    <Box px="2" pt="2">
+      <TouchableOpacity
+        onPress={() => {
+          navigation.navigate('observation', {
+            id: observation.id,
+          });
+        }}>
+        <Box bg="white" borderWidth="2" borderRadius="8" borderColor="light.200" p="4">
+          <VStack space="2">
+            <HStack justifyContent="space-between" alignItems="center">
```
stevekuznetsov, line 95: Is `HStack` basically just a flexbox presenting a column? Time will tell but I hope doing things this way doesn't make it hard to get the full flexbox set of possibilities.

stevekuznetsov: FWIW I really like how the flex-box things are front-and-center in the TSX here ... it was really weird to have them in the style for vanilla React Native, stuff that was super critical to the view ending up very low down on the page.

stevekuznetsov (commented):
This looks really good! Do you know why our bottom navbar seems to have moved up a bunch, looks like we have extra white-space down there now?

floatplane:
> This looks really good! Do you know why our bottom navbar seems to have moved up a bunch, looks like we have extra white-space down there now?

🦅 👀  nice catch 😄 

I do know! I just figured it out today - it's because of the way we have SafeAreaView set up. I had the fix locally when I screenshotted the before state, which is why that image looks good. Then I decided I was going to put the fix in a different PR, so didn't have it when I took the after screenshot. But this PR doesn't change anything with regards to that extra space.

components/Observations.tsx
floatplane: > Is HStack basically just a flexbox presenting a column? Time will tell but I hope doing things this way doesn't make it hard to get the full flexbox set of possibilities.

It's basically that, but then adds a `space` prop. So you can say "I want to stack these things in a column with 2px between them", which turns out to be a very useful way to build a lot of UI as you nest things together. We used this pattern at Stripe for lots of our UI.

floatplane: Totally agree with your assessment of the styles being separated from the components - very hard to make changes.

PR #34, opened 2022-12-30
Fix safe area views; remove debug tab

floatplane:
- Remove the top-level SafeAreaView so that the map goes all the way to the top of the screen (matches Figma mocks). Other screens in the app use SafeAreaViews as needed
- Making this change fixes the extra whitespace at the bottom of the tab bar as well
- As previously discussed, moved debug settings to the "Menu" tab and removed "Debug"

iOS screenshots

home | menu
--- | ---
image | IMG_7A1E8F447244-1

### DIFF
--- a/clientContext.ts
+++ b/clientContext.ts
@@ -1,25 +1,33 @@
 import React, {Context} from 'react';
 
 export interface ClientProps {
+  avalancheCenter: string;
+  setAvalancheCenter: (center: string) => void;
   nationalAvalancheCenterHost: string;
   snowboundHost: string;
   staging: boolean;
   toggleStaging: () => void;
 }
 
-export const productionClientProps = {
+export const productionHosts = {
   nationalAvalancheCenterHost: 'https://api.avalanche.org',
   snowboundHost: 'https://api.snowobs.com',
 };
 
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Eek, I am goofing around with the menu screen in my branch right now but I'll rebase

clientContext.ts
```diff
@@ -1,25 +1,33 @@
 import React, {Context} from 'react';
 
 export interface ClientProps {
+  avalancheCenter: string;
```
stevekuznetsov, line 4: Oops, I also didn't notice this - IMHO we should be careful to put things in context as it would be really easy to accidentally not make a screen routable / deep-link-able if we build something that is pulling from context, not route state

floatplane: mmm interesting point! I guess we could pull the `avalancheCenter` prop out of context so that it can't be read from. the only thing context is allowing us to do is call `setAvalancheCenter` from a nested component.

I'll do that real quick right now.

floatplane:
> Eek, I am goofing around with the menu screen in my branch right now but I'll rebase

Sorry, I hope it isn't too painful! I didn't know you had started on that.

PR #35, opened 2022-12-30
App.tsx: add a Menu screen that hosts debug info conditionally

stevekuznetsov:


components/AvalancheCenterSelector.tsx
```diff
@@ -44,55 +45,84 @@ const center_idsByType: SectionListData<string>[] = [
...
-      renderItem={({item}) => <AvalancheCenterCard center_id={item} set_avalanche_center={setAvalancheCenter} />}
+      renderItem={({item}) => (
+        <AvalancheCenterCard
+          center_id={item}
+          selected={item === currentCenterId}
+          onPress={(center_id: string) => {
+            setAvalancheCenter(center_id);
+            // We need to clear navigation state to force all screens from the
+            // previous avalanche center selection to unmount
+            navigation.reset({
```
stevekuznetsov, line 119: We can probably nix this navigation now - and just have it update the state, let the user pop out of the menu as they need to

.github/workflows/preview.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: yarn install
 
       - name: 🚀 Publish preview
-        run: expo publish --release-channel=pr-${{ github.event.number }} --non-interactive
+        run: expo publish --profile preview --release-channel=pr-${{ github.event.number }} --non-interactive
```
stevekuznetsov, line 38: Hopefully this + the channels will mean that the preview builds show off the debug menu

components/screens/MenuScreen.tsx
```diff
@@ -1,32 +1,90 @@
...
+import {createNativeStackNavigator, NativeStackScreenProps} from '@react-navigation/native-stack';
 
-export const MenuScreen = () => {
-  const {staging, toggleStaging, setAvalancheCenter} = useContext<ClientProps>(ClientContext);
+import {MenuStackParamList, MenuStackNavigationProps} from 'routes';
+import {useNavigation} from '@react-navigation/native';
+
+const MenuStack = createNativeStackNavigator<MenuStackParamList>();
+export const MenuStackScreen = (
+  center_id: string,
```
floatplane, line 19: nitpick, but we should probably default to either camelCase or snake_case

components/screens/MenuScreen.tsx
```diff
@@ -1,32 +1,90 @@
...
+    console.log(`Switching to ${staging ? 'production' : 'staging'} environment`);
+  }, [staging, setStaging]);
+  const navigation = useNavigation<MenuStackNavigationProps>();
+  return function ({route}: NativeStackScreenProps<MenuStackParamList, 'menu'>) {
+    const {current_center_id} = route.params;
+    return (
+      <SafeAreaView style={styles.fullscreen}>
+        <VStack pt="16" px="4" space="4" style={styles.fullscreen}>
+          <Heading>Settings</Heading>
+          {Updates.channel !== 'production' && (
```
floatplane, line 49: nice 👊

components/screens/MenuScreen.tsx
```diff
@@ -1,32 +1,90 @@
...
-    </SafeAreaView>
+    <MenuStack.Navigator initialRouteName="menu">
+      <MenuStack.Screen name="menu" component={MenuScreen(staging, setStaging)} initialParams={{current_center_id: center_id}} options={{title: `Settings`}} />
+      <MenuStack.Screen
+        name="avalancheCenterSelector"
+        component={AvalancheCenterSelectorScreen(setAvalancheCenter)}
+        initialParams={{current_center_id: center_id}}
+        options={{title: `Choose An Avalanche Center`}}
+      />
+    </MenuStack.Navigator>
```
floatplane, line 33: this feels like a lot of code to essentially allow someone to pick from a list. do you really need a separate screen and the navigation mechanism?

to be clear, I don't feel that strongly about it - it's not something we're shipping! but it does make this source file a bit more busy than it was before.

floatplane (commented):
sorry, I saw your emails on my PR before I saw the Slack message about this PR, so I think I responded to things out of order 😄

components/AvalancheCenterSelector.tsx
floatplane: that seems reasonable to me

components/screens/MenuScreen.tsx
stevekuznetsov: Agreed - do you have a preference?

components/screens/MenuScreen.tsx
stevekuznetsov: Hard to say what is a hard need, but I was poking around the other apps on my phone and in pretty much every case, the top-level menu screen had generic-sounding options, each of which would navigate you to a new screen where you made decisions - so I figured aside from learning the navigation bits for myself, if/when we actually have things to be setting in the menu, we would likely end up with a similar pattern?

stevekuznetsov:
@floatplane "fixed" the rendering - would love to pair with you at some point to see your workflow for layout etc - I have no idea what I'm doing.

Some other changes added:

 - reverted to how you had the tabs taking navigation props, instead of my closures, so that we can deep-link to tabs
 - continued breaking out all of the tab screens from the App.tsx into `components/screens`

components/screens/MenuScreen.tsx
stevekuznetsov: I think things are self-consistent in this PR but I can do a pass after this to collapse to a single case in the rest of the codebase

components/screens/MenuScreen.tsx
floatplane: True, the UX pattern is common! 

I did spot https://github.com/jsoendermann/react-native-settings-screen while looking around last week and feel like something like that might be a decent option to explore if we end up with more settings to add.

defaults.ts
```diff
@@ -0,0 +1,3 @@
+export const defaults = {
+  avalancheCenter: 'NWAC',
```
floatplane, line 2: WDYT about putting this in app.json: https://github.com/stevekuznetsov/avalanche-forecast/blob/8bbd2804510d9e5d50fc151dac576443c9735080/app.json#L51-L56 (don't feel strongly, easy enough to move it later)

components/AvalancheCenterSelector.tsx
```diff
@@ -43,56 +44,88 @@ const center_idsByType: SectionListData<string>[] = [
...
+    <Box>
+      <TouchableOpacity
+        style={styles.item}
+        onPress={() => {
+          onPress(avalancheCenterId);
+        }}>
+        <Box bg={selected ? 'blue.100' : 'light.100'}>
+          <VStack>
+            <HStack justifyContent="flex-start" alignItems="center" px="2" width={width}>
+              <AvalancheCenterLogo style={{height: height / 20}} center_id={avalancheCenterId} />
```
floatplane, line 95: can't you just do `height: '5%'`? (I'm also kind of skeptical that we want the height of any icons to change based on screen size, they should be coupled to the size of the text that they're next to)

eas.json
```diff
@@ -1,16 +1,17 @@
...
   "cli": {
-    "version": ">= 0.60.0"
+    "version": ">= 0.60.0",
+    "appVersionSource": "remote"
   },
   "build": {
-    "development": {
-      "developmentClient": true,
-      "distribution": "internal"
-    },
```
floatplane, line 9: do we really want to get rid of this build config? at some point it seems to me that `preview` should be the same as a production build, i.e. `developmentClient` is going to be `false` so that we know that preview users are getting as close to a production experience as possible.

routes.ts
```diff
@@ -6,20 +6,10 @@ export type TabNavigatorParamList = {
   'Weather Data': {center_id: string; date: string};
   Observations: {center_id: string; date: string};
   Menu: {center_id: string};
-  Debug: {center_id: string};
```
floatplane, line 9: thank you 🙈

floatplane (commented):
ptal @stevekuznetsov 

I had a couple of questions, nothing majorly blocking!

floatplane (commented):
OOC what was wrong with the logo? I can't spot a difference in the PR behavior (the splash screen is only up very briefly though, so it's hard to tell...)

components/screens/MenuScreen.tsx
stevekuznetsov: Cool idea - that particular library hasn't seen a commit in 4 years, though - do you think that it's safe to use?

components/AvalancheCenterSelector.tsx
stevekuznetsov: Love the idea - can you show me how to make it relative to the font?

eas.json
stevekuznetsov: Yeah, I guess now that we're back to using Expo Go we don't really need the development client at all - this was a holdover change from the before times, I didn't want us to have to do two builds per commit (if we would have needed the dev client to do anything). 

Let me reinstate this and remove the dev client from `preview`

stevekuznetsov:
> OOC what was wrong with the logo?

Long story of me being an idiot - for the NWAC logo in the picker we had the NWAC + USFS logo in the past and I wanted something more minimal, and my editor made it look like you had added a good minimal one to the splash screen, so I nabbed that without realizing it had a lot of transparent space around the logo to fill out the screen!

defaults.ts
stevekuznetsov: Done

stevekuznetsov:
@floatplane no worries, rebased on `main`

components/AvalancheCenterSelector.tsx
```diff
@@ -43,56 +44,88 @@ const center_idsByType: SectionListData<string>[] = [
...
   if (isLoading) {
     return <ActivityIndicator style={styles.item} />;
   }
   if (isError) {
     return (
-      <View style={styles.item}>
-        <Text>{`Could not fetch data for ${center_id}: ${error?.message}.`}</Text>
-      </View>
+      <Box>
+        <Box bg="light.100">
```
yuliadub, line 61: for styling things such as this, would it make sense to maybe centralize them somewhere? applies to Box (selected or not), NBText, HStack, etc - maybe through a common style sheet or something like that?

stevekuznetsov: Yeah, I think Brian's working on something like that - the styling/visual bits in this PR were copy-pasted from the proto-work we already have, and I'm going to come by and clean things up once we have a better approach figured out.

components/screens/MenuScreen.tsx
```diff
@@ -1,30 +1,81 @@
...
-        </HStack>
-        <AvalancheCenterSelector setAvalancheCenter={setAvalancheCenter} />
-      </VStack>
-    </SafeAreaView>
+    <MenuStack.Navigator initialRouteName="menu">
+      <MenuStack.Screen name="menu" component={MenuScreen(avalancheCenterId, staging, setStaging)} options={{title: `Settings`}} />
+      <MenuStack.Screen
+        name="avalancheCenterSelector"
+        component={AvalancheCenterSelectorScreen(avalancheCenterId, setAvalancheCenter)}
+        options={{title: `Choose An Avalanche Center`}}
```
yuliadub, line 30: not sure if I am just not understanding the flow, but when I load the app, and go to menu which shows settings (assuming that currently when the app loads, center is already selected for me, which is NWAC, the setting screen shows "choose an avalanche center" and the only option below it is nwac. If I click on nwac, it takes me to the list of all centers. if I select nwac again from this list, it takes me to the main map but it appears to be more zoomed in then before (not sure why)

i assume the experience would be something like: "current avalanche center" and then nwac below, and then below that would be a button that would basically say something like "change center"? but let me know if I am misunderstanding. 

I do like this flow (having the menu/settings determine the center that is), easy to take add new centers as they onboard or take them away too and less advanced users wont be too confused hopefully :D

components/screens/TelemetryScreen.tsx
```diff
@@ -0,0 +1,50 @@
...
+const TelemetryStack = createNativeStackNavigator<TelemetryStackParamList>();
+export const TelemetryTabScreen = ({route}: NativeStackScreenProps<TabNavigatorParamList, 'Weather Data'>) => {
+  const {center_id, date} = route.params;
+  return (
+    <TelemetryStack.Navigator initialRouteName="telemetryStations">
+      <TelemetryStack.Screen
+        name="telemetryStations"
+        component={TelemetryScreen}
+        initialParams={{center_id: center_id, date: date}}
+        options={() => ({title: `${center_id} Telemetry Stations`})}
```
yuliadub, line 17: if i switch between centers and the center has no weather stations we can find, it defaults back to wa seattle centric map view, or sometimes will show wa stations randomly. not sure what going on with that - but maybe we should default to the zone of the center for this page no matter what? just random thoughts

components/screens/TelemetryScreen.tsx
```diff
@@ -0,0 +1,50 @@
...
+      />
+      <TelemetryStack.Screen
+        name="telemetryStation"
```
yuliadub, line 20: this shows up as a transparent card right now, only for the first random stations, is the idea to have the stations be clickable kind of like avalanche forecast zone cards? honestly might just be easier to have the clickable part be the name that appears when the pin is clicked?

components/screens/MenuScreen.tsx
stevekuznetsov: I was modelling this off of the Slack menu where you can choose the app icon - it shows the currently-selected icon in the menu and makes it obvious through a combination of title and chevron that you can go into the sub-menu to choose a different icon. Likely need to look at more menus to figure out a cleaner way to get these concepts across?

components/screens/TelemetryScreen.tsx
stevekuznetsov: Yeah, I noticed that bad behavior - unrelated to this PR but we need to figure out how to gracefully handle that.

components/screens/TelemetryScreen.tsx
stevekuznetsov: Yeah, I was expecting to make these two UX bits similar once we have them figured out, just slapped something trivial together in the past when I created the telemetry screen to prove out the SnowBound API - I think having the pin clickable would be good, or having it auto-scroll to the correct card (and show some metadata about what sensors exist) would also be good.

PR #37, opened 2022-12-30
Remove current avalanche center from context

floatplane:
as discussed in https://github.com/stevekuznetsov/avalanche-forecast/pull/34#discussion_r1059199857


floatplane:
@stevekuznetsov this is definitely going to conflict with #35 so feel free to close this if you'd like. It's not obvious that it's worth using context to plumb `setAvalancheCenter` down since the hierarchy isn't _that_ deep.

stevekuznetsov:
Yeah if you're OK with something like the changes in #35 - maybe we can close this one out?

PR #40, opened 2023-01-03
Make reusable Card component, rework top of avy forecast to use it

floatplane:
pretty close to the figma mocks now (or at least one version of them 😄). not going to spend more time sweating pixels at this point when the designs are still in flux. 

figma | app
--- | ---
image | image

### DIFF
--- a/components/Card.tsx
+++ b/components/Card.tsx
@@ -1,30 +1,23 @@
-import React, { PropsWithChildren, ReactNode, useCallback } from 'react';
-import { compareDesc, format, parseISO, sub } from 'date-fns';
-import { ActivityIndicator, View, Text, FlatList, TouchableOpacity, useWindowDimensions } from 'react-native';
-import { HomeStackNavigationProps } from 'routes';
-import { useNavigation } from '@react-navigation/native';
-import { Box, Divider, VStack, HStack, Text as NBText, Heading } from 'native-base';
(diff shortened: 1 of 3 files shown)


components/Card.tsx
```diff
@@ -0,0 +1,28 @@
...
+  borderColor?: string;
+}
+
+export const Card: React.FunctionComponent<PropsWithChildren<CardProps>> = ({header, onPress, borderColor, borderRadius, children, ...boxProps}) => {
+  const pressHandler = useCallback(() => onPress?.(), [onPress]);
+
+  return (
+    <Box {...boxProps}>
+      <TouchableOpacity onPress={pressHandler}>
+        <Box bg="white" borderWidth="2" borderRadius={borderRadius ?? 8} borderColor={borderColor ?? 'light.200'} p="4">
```
stevekuznetsov, line 18: Are we sure that we're not going to need to overwrite anything else on this nested `Box` (and, if not, is there a pattern here that doesn't require us to expose each additional prop as an optional parameter for the `Card`?)

floatplane: Not sure if we'll want more control than this - I wouldn't want to spend too much time implementing solutions until there's a need. (I'm actually not certain the outer box is even needed, we might want to just pass through Box properties onto the inner box and remove the outer one)

stevekuznetsov: Sounds good!

PR #41, opened 2023-01-03
Updates

floatplane:
- changes to config files recommended by `eas update:configure`
- remove a `console.dir` that was causing our updates to crash


floatplane:
@stevekuznetsov I'm hoping these changes will allow Expo Go users to pull the latest bundle from `main` in the `preview` channel

Issue #42, opened 2023-01-04
Expired forecasts causing problems?


floatplane:
Noticed that many regions are showing no data after 6 PM PST, possibly because of a lag between the latest NWAC forecast being published and the NAC receiving it?

Missing data:
<img width="330" alt="image" src="https://user-images.githubusercontent.com/101196/210474220-1ccf6e0c-13a2-4510-b867-f6b26663770b.png">

Also realizing that we should probably be able to display that a forecast is expired *and* that's expected, like in the Olympics:

<img width="1196" alt="image" src="https://user-images.githubusercontent.com/101196/210474308-ba4cb632-1546-4af2-810c-b732984a6f05.png">

floatplane:
I don't think this is necessarily expired forecasts - it seems like the data is there. But I wonder if our date filtering in `useAvalancheForecastFragment` is using dates generated in the local timezone when it should be using UTC!

floatplane:
This does seem to be it. 6 PM PST is ~2 AM UTC so we were in different days 😬

stevekuznetsov:
`date-fns` should have an easy fix for the timezone issue, would be good to add documentation to the hooks that makes it clear they expect UTC.

re: expired - yes! I don't know if the design has anything for that yet, I think I left a comment on Figma about that as well.

PR #43, opened 2023-01-04
A pile of accumulated changes

floatplane:
in order:
- continue breaking up the avalanche tab into cards
- start making cards collapsible
- fix the missing outline around map regions
- change to the newest React Native JS engine so that remote debugging just works
- hack in a solution to #42, which appears to be caused by UTC/local time mismatches


components/forecast/AvalancheTab.tsx
```diff
@@ -53,88 +52,67 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
 
   const elevationBandNames: ElevationBandNames = zone.config.elevation_band_names;
 
+  const [textColor] = useToken('colors', ['darkText']);
```
stevekuznetsov, line 55: What's this dark/light switch for? Is it using the built-in dark-mode support in Native Base?

components/forecast/AvalancheTab.tsx
```diff
@@ -53,88 +52,67 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
...
-      <View style={styles.discussion}>
-        <RenderHTML source={{html: forecast.hazard_discussion}} contentWidth={windowWidth} />
-      </View>
-      <Text style={styles.heading}>Media</Text>
+      <Card
+        borderRadius={0}
+        borderColor="white"
+        header={
+          <HStack space={2} alignItems="center">
+            {/* TODO wrong icon here */}
```
stevekuznetsov, line 95: What's wrong about this icon?

components/forecast/AvalancheTab.tsx
```diff
@@ -53,88 +52,67 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
...
+            <AvalancheDangerIcon style={{height: 24}} level={highestDangerToday} />
```
stevekuznetsov, line 96: FWIW 24 is very tiny :)

patches/react-native-collapsible+1.6.0.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/node_modules/react-native-collapsible/index.d.ts b/node_modules/react-native-collapsible/index.d.ts
```
stevekuznetsov, line 1: Is there an upstream issue tracking this (or a PR open to their repo to fix it)?

components/forecast/AvalancheTab.tsx
```diff
@@ -102,14 +102,14 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
       <Card borderRadius={0} borderColor="white" header={<Heading>Avalanche Danger</Heading>}>
         <AvalancheDangerTable date={parseISO(forecast.published_time)} current={currentDanger} outlook={outlookDanger} elevation_band_names={elevationBandNames} />
       </Card>
-      <Card borderRadius={0} borderColor="white" header={<Heading>Avalanche Problems</Heading>}>
+      <CollapsibleCard startsCollapsed borderRadius={0} borderColor="white" header={<Heading>Avalanche Problems</Heading>}>
```
stevekuznetsov, line 105: nit: do we want to collapse all of the problems at once, or each individual problem?

package.json
```diff
@@ -53,7 +53,7 @@
     "react-native-collapsible": "^1.6.0",
     "react-native-gesture-handler": "~2.8.0",
     "react-native-maps": "1.3.2",
-    "react-native-reanimated": "^2.13.0",
+    "react-native-reanimated": "~2.12.0",
```
stevekuznetsov, line 56: Should we add `expo doctor` to our CI?

app.json
```diff
@@ -15,6 +15,7 @@
     "runtimeVersion": {
       "policy": "sdkVersion"
     },
+    "jsEngine": "hermes",
```
stevekuznetsov, line 18: This is iOS only, right?

hooks/useAvalancheForecastFragment.ts
```diff
@@ -17,8 +17,10 @@ export const useAvalancheForecastFragment = (center_id: string, forecast_zone_id
     const {data} = await axios.get(url, {
       params: {
         avalanche_center_id: center_id,
-        date_start: format(sub(date, {days: 1}), 'y-MM-dd'),
-        date_end: format(add(date, {days: 1}), 'y-MM-dd'),
+        forecast_zone_id,
+        // TODO(brian): remove this hack of adding/subtracting two days, which works around issues converting between local day and UTC day
```
stevekuznetsov, line 21: `date-fns` can convert between time-zones, should we just do that instead of this?

hooks/useAvalancheForecastFragment.ts
```diff
@@ -36,6 +38,7 @@ export const useAvalancheForecastFragment = (center_id: string, forecast_zone_id
       });
       throw parseResult.error;
     } else {
+      // TODO(brian): This is assuming that a forecast always exists for the given zone/date range. That's not a good assumption!
```
stevekuznetsov, line 41: Doesn't the `find` return an option, encoding the possibility of no response in the output type?

patches/react-native-collapsible+1.6.0.patch
floatplane: there are both :)

components/forecast/AvalancheTab.tsx
floatplane: this is for grabbing the default text color from the NativeBase theme

components/forecast/AvalancheTab.tsx
floatplane: actually probably nothing! I was comparing it to the Figma mock which uses the "no information" icon, but we _should_ use the current conditions icon - I'll remove the comment.

components/forecast/AvalancheTab.tsx
floatplane: yeah, probably should be 32!

components/forecast/AvalancheTab.tsx
floatplane: 🤷 they're separate in [Figma](https://www.figma.com/file/Bai2w7x0vUakL5WRjPGRg7/App---V3?node-id=97%3A89887&t=VvefbQ7LI0dqsU2z-0), I assume this is one of those things that might change with feedback.

package.json
floatplane: it's an interesting idea, I like it!

app.json
floatplane: as far as I know this is universal, but only became available for iOS in Expo 47. 
- https://hermesengine.dev/
- https://docs.expo.dev/guides/using-hermes/

hooks/useAvalancheForecastFragment.ts
floatplane: yeah, I just want to write a test to make sure the fix is right, so I was hoping to unblock this PR and then take a careful pass. 

I _think_ the issue is the format string implicitly says "give me the date in my current TZ" but want to be sure - and also want to check for any other place we might be doing this.

hooks/useAvalancheForecastFragment.ts
floatplane: I think my concern is that we probably want to show the most recent forecast, expired or not. Right now we don't show anything for the Olympics, but there is useful info there, even if it's just to know when the next forecast is coming: https://nwac.us/avalanche-forecast/#/olympics

floatplane:
> @floatplane the preview does not work - perhaps the changes to the preview you made in the previous PR were inconsistent? Unclear if the preview build is broken or the comment is bad - but scanning the image I get
> 
> ```
> There was a problem running the requested app.
> 
> No project found at exp://exp.host/@nwac/avalanche-forecast?release-channel=pr-43.
> ```

It's the comment that's bad - you can load the preview by going to Expo Go, selecting the project and finding pr-43 under "Branches". I'll see if I can fix the comment.

components/forecast/AvalancheTab.tsx
stevekuznetsov: Should we have it per-problem for now (I imagine designers will call it out on review?)

components/forecast/AvalancheTab.tsx
stevekuznetsov: Yeah, there was a comment somewhere in the file about that

hooks/useAvalancheForecastFragment.ts
stevekuznetsov: Yeah it is not clear to me what the dates mean on this API - maybe we can reach out to Chris and ask? Are we providing the date range for the publish date for the thing, or the date range for which it's valid. or ... ?

hooks/useAvalancheForecastFragment.ts
stevekuznetsov: /shrug

I don't know that we have per-day deadlines that you need to hit, or obversely the cost of this "block" is small :)

FWIW I think we could also ask for clarification for what, specifically, the API is applying these date ranges to - the date at which the forecast was published? Last updated? The dates on which the forecast is valid?

stevekuznetsov:
Some comments from reviewing on iOS:

- when a card is collapsed, should we still show the horizontal divider below the header?
- we're missing the per-avalanche-problem header
- issued & expires are breaking differently- might be nice to have it consistently render with the line-break
  - the current format has the time of day after the year, which seems weird?

components/forecast/AvalancheTab.tsx
```diff
@@ -53,88 +52,66 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
...
+        color: textColor,
+      },
+    },
+  };
+
   return (
-    <VStack space="4" bgColor={'#f0f2f5'}>
-      <Card paddingTop={2} borderRadius={0} borderColor="white" header={<Heading>Avalanche Forecast</Heading>}>
+    <VStack space="2" bgColor={'#f0f2f5'}>
+      <Card marginTop={2} borderRadius={0} borderColor="white" header={<Heading>Avalanche Forecast</Heading>}>
```
stevekuznetsov, line 68: Probably a better question for Charlotte & the designers, but what is this header doing for us? We're inside of an avalanche forecast tab, which someone clicked on specifically, and we're already showing the zone at the top of the screen - do we really need to spend 10% of our vertical real estate above the fold on a big header that delivers little info?

components/forecast/AvalancheTab.tsx
```diff
@@ -53,88 +52,66 @@ export const AvalancheTab: React.FunctionComponent<AvalancheTabProps> = React.me
...
+        borderColor="white"
+        header={
+          <HStack space={2} alignItems="center">
+            <AvalancheDangerIcon style={{height: 32}} level={highestDangerToday} />
+            <Heading>The Bottom Line</Heading>
+          </HStack>
+        }>
+        <RenderHTML source={{html: forecast.bottom_line}} {...renderHTMLProps} />
+      </Card>
+      <Card borderRadius={0} borderColor="white" header={<Heading>Avalanche Danger</Heading>}>
```
stevekuznetsov, line 101: This `<Heading>Avalanche Danger</Heading>` is duplicated inside of the `<AvalancheDangerTable/>`

components/AvalancheForecastZonePolygon.tsx
```diff
@@ -128,7 +128,8 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
     <Polygon
       coordinates={coordinates}
       fillColor={colorFor(dangerLevel).alpha(feature.properties.fillOpacity).string()}
-      strokeColor={colorFor(dangerLevel).string()}
+      strokeColor={'#484848'}
```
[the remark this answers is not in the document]
stevekuznetsov: we will need to carry forward a solution for that to the danger rose and the danger pyramid as well

components/forecast/AvalancheTab.tsx
floatplane: AvalancheDangerTable needs major surgery

components/AvalancheForecastZonePolygon.tsx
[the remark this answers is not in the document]
floatplane: no idea? I mean, it doesn't render that well now - I can't see boundaries between zones very well

components/forecast/AvalancheTab.tsx
floatplane: I think this is a discussion to have in Figma, or while user-testing - I don't have the answer you seek. 

But also, the designers have explicitly said that the Avalanche Forecast page is a WIP, so I wouldn't get too worried either way. We can iterate.

floatplane:
> Some comments from reviewing on iOS:
> 
> * when a card is collapsed, should we still show the horizontal divider below the header?

I plan to fix

>   
>   * currently you need to tap the chevron to open and close the card, and the touch target is pretty small - might be nice to be able to touch the entire header instead

I plan to fix

>   * did we get some idea about whether or not the sections (and which) should be collapsed by default?

not yet

> * we're missing the per-avalanche-problem header

I plan to fix

>   
>   * might not be something you tried to tackle at all in this PR, if so disregard, but we seem to lose a lot of horizontal width by the time we're in the per-avalanche-problem card
> * issued & expires are breaking differently- might be nice to have it consistently render with the line-break

it'd probably be nice to change the design so we don't need a line break at all, but I think that's a design question

>   
>   * the current format has the time of day after the year, which seems weird?

[is it?](https://nwac.us/avalanche-forecast/#/olympics) it's the style currently in use

> * might it be possible to justify the text?

can you be more specific?

> * the underline for the Avalanche/Weather/Observations tab is maybe 80% of the width of the card (or 100% of the text?) in our Figma design - looks like this PR has it as 100% of the width of the tab picker

I plan to fix

> * might pre-date any work you're doing, but tapping on an image inside of an avalanche problem makes it go away forever

Haven't touched that code, I'm guessing it might be a hyperlink that's not handled

components/AvalancheForecastZonePolygon.tsx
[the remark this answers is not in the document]
stevekuznetsov: Yep, I'll file us an issue to remember to figure this out at some point

stevekuznetsov:
> > the current format has the time of day after the year, which seems weird?
> 
> [is it?](https://nwac.us/avalanche-forecast/#/olympics) it's the style currently in use

Gotcha - definitely matches what the website does, but it's just a weird way to format a date, in that computers love ISO but humans are more likely to say "seven PM on the third of january" than "the third of january seven PM" :)

> > might it be possible to justify the text?
>
> can you be more specific?

I wonder if the `<RenderHTML>` components would let us justify the text to the width of the screen, rather than having a jagged right margin - might just be my personal preference but it might look nicer that way.

components/AvalancheForecastZonePolygon.tsx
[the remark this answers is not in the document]
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

components/forecast/AvalancheTab.tsx
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

components/forecast/AvalancheTab.tsx
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

hooks/useAvalancheForecastFragment.ts
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

hooks/useAvalancheForecastFragment.ts
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

package.json
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

components/forecast/AvalancheTab.tsx
stevekuznetsov: Tracked in https://github.com/stevekuznetsov/avalanche-forecast/issues/45

PR #47, opened 2023-01-05
Forecast page changes

floatplane:
- bugfix: disable touch on avalanche problem images: the component isn't fully baked yet
- bugfix: hide divider on collapsible cards, allow touch anywhere in header to toggle
- cleanup: get configuration for RenderHTML into a single component called `HTMLRenderer`, start using preferred fonts

### DIFF
--- a/components/AvalancheProblemImage.tsx
+++ b/components/AvalancheProblemImage.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {Text, Image, StyleSheet, /*TouchableHighlight,*/ useWindowDimensions, View, ActivityIndicator} from 'react-native';
+import {Text, Image, StyleSheet, useWindowDimensions, View, ActivityIndicator} from 'react-native';
 import RenderHTML from 'react-native-render-html';
 
 import {MediaItem} from 'types/nationalAvalancheCenter';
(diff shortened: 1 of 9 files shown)


components/AvalancheProblemImage.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {Text, Image, StyleSheet, TouchableHighlight, useWindowDimensions, View, ActivityIndicator} from 'react-native';
+import {Text, Image, StyleSheet, /*TouchableHighlight,*/ useWindowDimensions, View, ActivityIndicator} from 'react-native';
```
stevekuznetsov, line 3: Feel free to just delete the thing - `git revert` is easy and won't leave comments in here.

stevekuznetsov:
When the card is collapsed, the divider is not shown but there's a bunch of white-space left at the bottom of it - it looks like more space than we have on top of the header, I think? Otherwise this looks great!

floatplane:
> When the card is collapsed, the divider is not shown but there's a bunch of white-space left at the bottom of it - it looks like more space than we have on top of the header, I think? Otherwise this looks great!

fixed!

floatplane:
Long press to show HTML source (only when `__DEV__` is true):

Simulator Screen Recording - iPhone 14 Pro Max - 2023-01-05 at 20 04 34

PR #50, opened 2023-01-06
Map screen - forecast card styling

floatplane:
what's in this PR:
- card has been reskinned with font styles and spacing as defined in figma

what's not in this PR
- ~centering cards correctly - they get off center as you scroll through. I broke something here, just haven't looked hard at it yet.~ now fixed
- various selection behaviors
    - starting off with the zone list being partially visible
    - scrolling the correct zone into view when map region is tapped
    - moving the list out of the way when no region is selected

simulator_screenshot_7BEBF40E-49A9-4422-8750-52F98C31C97A

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -6,15 +6,16 @@ import MapView, {Region} from 'react-native-maps';
 import {useNavigation} from '@react-navigation/native';
 
 import {DangerScale} from 'components/DangerScale';
-import {AvalancheCenterID, Feature, MapLayer} from 'types/nationalAvalancheCenter';
+import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
 import {AvalancheCenterForecastZonePolygons} from './AvalancheCenterForecastZonePolygons';
 import {AvalancheDangerIcon} from './AvalancheDangerIcon';
-import {useMapLayer} from 'hooks/useMapLayer';
+import {MapViewZone, useMapViewZones} from 'hooks/useMapViewZones';
 import {HomeStackNavigationProps} from 'routes';
 import {SafeAreaView} from 'react-native-safe-area-context';
 import {Body, BodySmSemibold, Caption1, Caption1Black, Title3Black} from 'components/text';
 import {colorFor} from './AvalancheDangerPyramid';
 import {utcDateToLocalTimeString} from 'utils/date';
+import {parseISO} from 'date-fns';
 
 export const defaultRegion: Region = {
   // TODO(skuznets): add a sane default for the US?
@@ -48,7 +49,7 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
     latitudeDelta: 1.05 * region.latitudeDelta,
     longitudeDelta: 1.05 * region.longitudeDelta,
   };
-  const {isLoading, isError, data: mapLayer} = useMapLayer(center);
+  const {isLoading, isError, data: zones} = useMapViewZones(center, parseISO(date));
 
   return (
     <>
@@ -81,7 +82,7 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
           </Alert>
         </Center>
       )}
-      {!isLoading && !isError && <AvalancheForecastZoneCards key={center} center={center} date={date} mapLayer={mapLayer} />}
+      {!isLoading && !isError && <AvalancheForecastZoneCards key={center} date={date} zones={zones} />}
     </>
   );
 };
@@ -92,16 +93,15 @@ export const CARD_SPACING = CARD_WHITESPACE / 2; // proportion of overall width 
 export const CARD_MARGIN = CARD_SPACING / 2; // proportion of overall width that each card needs as a margin
 
 const AvalancheForecastZoneCards: React.FunctionComponent<{
-  center: AvalancheCenterID;
   date: string;
-  mapLayer: MapLayer;
-}> = ({center, date, mapLayer}) => {
+  zones: MapViewZone[];
+}> = ({date, zones}) => {
   const {width} = useWindowDimensions();
 
   const props = {
     snapToAlignment: 'start',
     decelerationRate: 'fast',
-    snapToOffsets: mapLayer?.features.filter(feature => feature.type === 'Feature').map((feature, index) => index * CARD_WIDTH * width + (index - 1) * CARD_SPACING * width),
+    snapToOffsets: zones?.map((_itemData, index) => index * CARD_WIDTH * width + (index - 1) * CARD_SPACING * width),
     contentInset: {
       left: CARD_MARGIN * width,
       right: CARD_MARGIN * width,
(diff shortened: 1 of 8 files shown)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -53,154 +48,82 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
...
+        zoomEnabled={true}
+        scrollEnabled={true}
+        provider={'google'}>
+        {isReady && <AvalancheCenterForecastZonePolygons key={center} center_id={center} setRegion={setRegion} date={date} />}
+      </MapView>
+      <SafeAreaView>
+        <DangerScale px="4" width="100%" position="absolute" top="12" />
+      </SafeAreaView>
+
+      {isLoading && <FeatureTitleBlack>Loading...</FeatureTitleBlack>}
```
stevekuznetsov: Why do none of these conditionally-rendered components require a `<SafeAreaView>`?

stevekuznetsov (commented):
How can we test the loading and error states here? Not even in an automated way - is there a simple way to mock out the hook to manually test it?

utils/date.tsx
```diff
@@ -0,0 +1,9 @@
+import {format, parseISO} from 'date-fns';
+
+export const dateToString = (dateString: string | undefined): string => {
```
stevekuznetsov, line 3: nit: let's call this `canonicalDateFormat` or something - lots of ways to `dateToString` ;)

components/AvalancheForecastZoneMap.tsx
```diff
@@ -53,154 +48,94 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
...
+const AvalancheForecastZoneCard: React.FunctionComponent<{
   feature: Feature;
   date: string;
-  style: ViewStyle;
-}> = ({feature, date, style}) => {
-  const forecastDate: Date = parseISO(date);
+}> = ({feature, date}) => {
   const {width} = useWindowDimensions();
   const navigation = useNavigation<HomeStackNavigationProps>();
-  const {isLoading, isError, data: forecast, error} = useAvalancheForecastFragment(feature.properties.center_id, feature.id, forecastDate);
```
stevekuznetsov, line 181: Also - the reason I had done it this way is:

- other than this transient data, everything else in the center metadata is effectively static, meaning we can safely cache it for very long periods of time
- if we derive this part of the map from the data in the map layer, in order to keep the entire app in sync we will need to methodically refresh the map layer every time we refresh the avalanche forecast, whether that's a background refresh or driven by push notifications

It seems more efficient and likely easier to ensure we don't make mistakes if we paint this based on the avalanche forecast API itself, but the dichotomy between `v2/public/products` (`useAvalancheForecastFragment`) and `v2/public/product` (`useAvalancheForecast`) has similar source-of-truth issues. 

We should also think about what happens if a user's in a forecast view, refreshes to get an updated forecast, and goes back to the map - ideally, they get the updated forecast colored in to the zone. 

Probably good to file this one away for thinking about more later, the APIs here do not give us what we need, really.

components/AvalancheForecastZoneMap.tsx
```diff
@@ -25,23 +25,18 @@ export const defaultRegion: Region = {
 };
 
 export interface MapProps {
-  centers: string[];
+  center: string;
```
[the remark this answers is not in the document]
floatplane: what are we losing exactly? this behavior was unused, therefore untested, and so it just muddied the logic of how the component worked. how much of the app is set up to handle multiple avalanche centers at once?

components/AvalancheForecastZoneMap.tsx
floatplane: because they're absolutely positioned, in the center or at the bottom of the screen

components/AvalancheForecastZoneMap.tsx
floatplane: But this was pulling N requests just to get the transient data for each zone, rather than 1 request for the whole center. It feels like a no-brainer to remove these requests from this view. What am I missing? Even if we pre-cache the map regions and just pull the danger level from the response, it's still way less resource intensive.

Pre-loading the detailed forecasts optimistically seems like a worthy thing, but not at the expense of slowing this component down.

utils/date.tsx
floatplane: agreed that it should have a less generic name, but I wanted to wait until we add a second date format so that we could come up with names that are less generic. something tells me there won't be a single canonical format (there might be date vs date-and-time, for example).

components/AvalancheForecastZoneMap.tsx
floatplane: And even if we decide to go back to N requests, I think it's probably saner to group the N requests with `Promise.all` and treat the whole group as succeeding or failing. Having individual cards in success/loading/error states feels like the wrong level of specificity for end-users.

floatplane:
> How can we test the loading and error states here? Not even in an automated way - is there a simple way to mock out the hook to manually test it?

I just did it manually by changing the component to display what I wanted. I suppose another way is to take your device offline. And a 3rd way would be to use a debugging proxy and make that URL return a 500.

components/AvalancheForecastZoneMap.tsx
floatplane: > We should also think about what happens if a user's in a forecast view, refreshes to get an updated forecast, and goes back to the map - ideally, they get the updated forecast colored in to the zone.

Agreed! Things like this have got me thinking about integrating Redux or some other state management library. Redux would also be a clean way to boot up the app with data that we know doesn't change much, like map region data, and still allow for filling in changes dynamically based on network requests.

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: Gotcha - if someone turned their phone upside down, wouldn't the "bottom" have the iOS notch?

components/AvalancheForecastZoneMap.tsx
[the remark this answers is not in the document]
stevekuznetsov: Everything downstream of the data aggregation layer was 100% ambivalent to how many centers were supplying the set of polygons. _Nothing_ in the app is "tested" - and I'm happy to add some end-to-end tests for this view once we figure out a testing flow we are going to use.

Showing a map with multiple centers is how the [avalanche.org](https://avalanche.org/#/current) homepage works and I could see us using that behavior in the future. Can you explain why you need this change?

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: I might be wrong here, but the beauty of `react-query` is that you get the state management for free and there's no need whatsoever for more management on top. We can set the caching behavior for the map layer to have a TTL of a month or something similar.

The issue with pulling forecast data from the map layer or the many-products API is just a source-of-truth thing, as I noted - we will need to be very careful to keep our caches coherent.

utils/date.tsx
stevekuznetsov: OK, let's start with `canonicalizeDateAndTime()` or `formatDateAndTime()` and iterate from there? We know even today it's doing more than `string` :)

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: Request de-duplication bug fixed in https://github.com/stevekuznetsov/avalanche-forecast/pull/52

stevekuznetsov: I was thinking more about your comments about slowing the component down and leading to a bad UX when only one zone was failing to load, and I think there might also be a fairly straightforward way to achieve that - the `useAvalancheForecastFragment()` hook is simply filtering the data from `useAvalancheForecastFragments()` - which applies to all zones in the center. If we pulled that state up one level, we'd have to handle the entire set of fragments being missing or not, but never any individual one. In practice, we'd of course never see an individual one missing since there's only one network request underneath it all, but you are 100% right that all the extra error handling code for each zone is completely unnecessary. 

WDYT? Would such an approach provide the better UX you are looking for?

floatplane: Is your goal ultimately to never call `/v2/public/products/map-layer`? I guess I'm not understanding what's driving you here, maybe we can dig into that in our chat.

> Notably, the two layers of requests remain required to display the forecast details- take a look at the useAvalancheForecast flow - there is no way to get the ID of the latest forecast for a zone without asking for all the forecasts for that zone during some period of time as the map layer leaves out this crucial bit of information.

I could also use an explainer here because it seems like you can always just fetch the latest forecast for a given `center_id` and `zone_id`. I think I'm missing something!

```
❯ curl 'https://api.avalanche.org/v2/public/product?type=forecast&center_id=NWAC&zone_id=1131' | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9835    0  9835    0     0  65702      0 --:--:-- --:--:-- --:--:-- 71788
{
  "id": 116746,
  "published_time": "2023-01-09T02:00:00+00:00",
  "expires_time": "2023-01-10T02:00:00+00:00",
  "created_at": "2023-01-09T01:36:29+00:00",
  "updated_at": "2023-01-09T01:40:45+00:00",
  "author": "Dallas Glass",
  "product_type": "forecast",
  // etc...
```

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: Oh - gotcha, had no idea that was an opt-in thing!

floatplane:
ptal @stevekuznetsov 

0f17766 has the big change we discussed earlier - melding data from `useMapView` and `useAvalancheForecastFragments`. while doing that, I did uncover a problem that we may have to solve later: forecast fragments don't include travel advice, so if we're going to aggressively cache the results of `useMapView`, then we'll need to find a way to generate updated travel advice from the current danger level. I can add a note to that effect to #54 if we decide to go ahead with this.

stevekuznetsov:
@floatplane this all looks good! gotta love the consistency between these APIs, sigh.

PR #51, opened 2023-01-06
components/Observation: flesh out the observation view

stevekuznetsov:
This tries to exhaustively present data about observations. Navigate to the BTAC page to see any.


hooks/useMapLayer.ts
```diff
@@ -11,22 +11,28 @@ import {ZodError} from 'zod';
 
 export const useMapLayer = (center_id: string) => {
   const clientProps = React.useContext<ClientProps>(ClientContext);
-  return useQuery<MapLayer, AxiosError | ZodError>(['map-layer', center_id], async () => {
```
stevekuznetsov, line 14: I have no idea why `prettier` decided to entirely reformat this file - @floatplane any thoughts?

floatplane: I'm guessing that once you added a 3rd argument to `useQuery`, prettier decided it had to break the arguments and align them?

hooks/useMapLayer.ts
```diff
@@ -11,22 +11,28 @@ import {ZodError} from 'zod';
...
+            center_id,
+          },
+        });
+        throw parseResult.error;
+      } else {
+        return parseResult.data;
+      }
+    },
+    {
+      enabled: !!center_id,
```
floatplane, line 35: seems like we should have a predefined data type for `Center`, as opposed to potentially letting it be any string

types/nationalAvalancheCenter/observations.ts
```diff
@@ -0,0 +1,244 @@
+const Unknown = 'Unknown';
+
+function reverseLookup<T extends PropertyKey, U extends PropertyKey>(mapping: Record<T, U>, lookup: U): string {
+  return Object.entries(mapping)
+    .filter(([_, value]) => value === lookup)
+    .map(([key, _]) => key)
+    .reduce((accumulator, currentValue) => currentValue, Unknown);
```
floatplane, line 7: is the `reduce` just to get the first matching value?

types/nationalAvalancheCenter/observations.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export const Activity = {
+  'Skiing/Snowboarding': 'skiing_snowboarding',
+  'Snowmobiling/Snowbiking': 'snowmobiling_snowbiking',
+  'XC Skiing/Snowshoeing': 'xcskiing_snowshoeing',
+  Climbing: 'climbing',
+  'Walking/Hiking': 'walking',
+  Driving: 'driving',
+  Other: 'other',
+} as const;
```
floatplane, line 18: curious why you'd use the human-readable strings as keys and the api strings as values, instead of vice-versa. the reverse lookup seems like what you'd ~always want. 

or am I misunderstanding and the API is actually returning the human-readable strings? in which case WTF 😄

components/Observations.tsx
```diff
@@ -53,62 +58,130 @@ export const Observations: React.FunctionComponent<{
...
-        <NBText color="light.400">{observation.observerType}</NBText>
-      </HStack>
+      <Row flexWrap="wrap" space="2">
+        <IdentifiedInformation header={'Submitted'} body={dateToString(observation.createdAt)} />
+        <IdentifiedInformation header={'Observer'} body={FormatPartnerType(observation.observerType as PartnerType)} />
+        <IdentifiedInformation header={'Author(s)'} body={observation.name || 'Unknown'} />
+      </Row>
+      {anySignsOfInstability && (
+        <Column space="2" style={{flex: 1}}>
+          <Text bold style={{textTransform: 'uppercase'}}>
```
floatplane, line 122: maybe use `Body` instead of `Text` (fine if you want to handle in a follow-up PR)

App.tsx
```diff
@@ -150,6 +150,7 @@ const App = () => {
       Lato_700Bold_Italic,
       Lato_900Black,
       Lato_900Black_Italic,
+      NAC_Icons: require('./assets/fonts/nac-icons.ttf'),
```
floatplane, line 153: with the font, I think you don't need `avalanche-icon.png` in this PR anymore

hooks/useMapLayer.ts
floatplane: like, I'm not sure how we would ever let an empty string in here, but we could enforce that with TS

types/nationalAvalancheCenter/observations.ts
stevekuznetsov: Yep!

App.tsx
stevekuznetsov: Oops, good catch!

components/Observations.tsx
stevekuznetsov: This is what I wanted your help on :)

types/nationalAvalancheCenter/observations.ts
stevekuznetsov: Yeah, would love your input here - so for the mapping/display, I 100% agree that what you always want is the reverse mapping .. but when we're consuming data from the API, we want the value of the type to be e.g. `other` or `driving` (from the code you commented on) .... so it's a bit diametrically opposed.

stevekuznetsov:
I'll come back and address the comments today!

hooks/useMapLayer.ts
stevekuznetsov: Done in [bc598a9](https://github.com/stevekuznetsov/avalanche-forecast/pull/51/commits/bc598a95ffee9590d3d7dc2e83a849fcba9998a4)

PR #52, opened 2023-01-07
hooks: correctly de-duplicate requests for products

stevekuznetsov:
My previous implementation here was incorrectly adding the zone ID to the key-set for the fragments, whivh are zone-agnostic.

Only two requests now @floatplane !
IMG_0017


stevekuznetsov:
CI looks good!

floatplane:
nice!

PR #55, opened 2023-01-09
Fix avalanche danger size display

floatplane:
This change does the following:
- stop clamping avalanche danger size to an integer/enumerated value
- fix the logic in `AvalancheProblemSizeLine` so that it renders the danger correctly

I didn't remove `numberToProblemSize` from `enums.ts`, but it's currently unused - so maybe I should? Right now its main usefulness is an example of how to write unit tests, since we don't have any other tests 😬


stevekuznetsov (approved):
In e.g. the observations API, the size is also given as a number and I could totally see us trying to map that to the long-form descriptions of each size ("large enough to bury a car"), at which point that utility method will be useful, so let's keep it in.

Issue #54, opened 2023-01-09
Preload Avalanche Center, Map, Forecasts, Media on Startup


stevekuznetsov:
Need to do some more thinking, but at startup and perhaps driven by state (push notifications, etc) we should do pre-fetching of the data as much as we can (in parallel with loading fonts, etc).

We can then also abstract a hook to give us information about a forecast that can try to give the forecast API response, falling back to the fragment API, falling back to the map layer.

cc @floatplane

floatplane:
as discussed in #50, we will want to be able to generate travel advice from current danger level - but Charlotte confirmed that the advice is hard-coded and @stevekuznetsov found it hardcoded in the JS widget as well:

```
        let Xa = [{
            level: 0,
            rating: "No Rating",
            advice: "Insufficient data for issuing of danger ratings, but a summary of avalanche conditions exists. Read the summary for more information.",
            icon: Ya + "0.png"
        }, {
            level: 1,
            rating: "Low",
            advice: "<strong>Generally safe avalanche conditions.</strong> Watch for unstable snow on isolated terrain features.",
            icon: Ya + "1.png"
        }, {
            level: 2,
            rating: "Moderate",
            advice: "<strong>Heightened avalanche conditions on specific terrain features.</strong> Evaluate snow and terrain carefully; identify features of concern.",
            icon: Ya + "2.png"
        }, {
            level: 3,
            rating: "Considerable",
            advice: "<strong>Dangerous avalanche conditions.</strong> Careful snowpack evaluation, cautious route-finding and conservative decision-making essential.",
            icon: Ya + "3.png"
        }, {
            level: 4,
            rating: "High",
            advice: "<strong>Very dangerous avalanche conditions.</strong> Travel in avalanche terrain not recommended.",
            icon: Ya + "4.png"
        }, {
            level: 5,
            rating: "Extreme",
            advice: "<strong>Extraordinarily dangerous avalanche conditions.</strong> Avoid all avalanche terrain.",
            icon: Ya + "5.png"
        }];
 ```

stevekuznetsov:
We have robust top-level pre-fetching now!

PR #56, opened 2023-01-10
Improve API behavior, add top-level prefetching

stevekuznetsov:
api: use a client-side mapping for travel advice

We only get travel advice from the map layer and full forecast API, not
the fragment API, which makes relying on that field problematic. It
looks like the extant web application is simply using a client-side
mapping, anyway, instead of the API fields, so we can do that as well.

---

hooks: cache ~static data forever

The avalanche center metadata and map layer contain relatively static
data (like the name of the avalanche center and where it is in the
country), so we can cache these values for the lifetime of our app and
re-fetch them infrequently.

---

*: pre-fetch all data for the forecasts

On app load, we will now be fetching the center metadata, map layer, all
forecasts that pertain to our zones, and all media in all of the
forecasts. This will allow us to make sure that all data is present for
the user regardless of whether they've looked at all the forecasts
before going out of service.

---

AvalancheTab: add a key to the collapsible cards

---

Partially addresses #54

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -37,6 +37,7 @@ import {MenuStackScreen} from 'components/screens/MenuScreen';
 import {ObservationsTabScreen} from 'components/screens/ObservationsScreen';
 import {TelemetryTabScreen} from 'components/screens/TelemetryScreen';
 import {AvalancheCenterID} from './types/nationalAvalancheCenter';
+import {usePrefetchAllActiveForecasts} from './hooks/usePrefetchAllActiveForecasts';
 
 // The SplashScreen stays up until we've loaded all of our fonts and other assets
 SplashScreen.preventAutoHideAsync();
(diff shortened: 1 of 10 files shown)


App.tsx
```diff
@@ -130,90 +131,106 @@ const App = () => {
...
-                  </TabNavigator.Navigator>
-                </View>
-              </NavigationContainer>
-            </SafeAreaProvider>
-          </NativeBaseProvider>
-        </QueryClientProvider>
-      </ClientContext.Provider>
+      <QueryClientProvider client={queryClient}>
+        <AppUsingQueryClient />
+      </QueryClientProvider>
```
floatplane, line 137: this refactoring doesn't make the code immediately clearer to me - the old structure was deeply nested to be sure, but at least it was clear.

I could see an argument for splitting it up if it was to get a particular chunk of the component under test, but this feels kind of arbitrary.

App.tsx
```diff
@@ -130,90 +131,106 @@ const App = () => {
...
     );
   } catch (error) {
     Sentry.Native.captureException(error);
     throw error;
   }
 };
 
+const AppUsingQueryClient = () => {
```
floatplane, line 145: 1. this name is kind of odd, since this component is not referring to a query client at all. Maybe this refactoring would make more sense to me if the ordering was inverted, i.e. BaseApp, AppWithClientContext (which wraps BaseApp and adds ClientContext), App (which wraps AppWithClientContext and adds query client)
1. it seems inconsistent to have `staging` be in this component's state (and have to be plumbed downward), while values derived from `staging` are in context 🤔

components/helpers/travelAdvice.ts
```diff
@@ -0,0 +1,21 @@
...
+  switch (dangerLevel) {
+    case DangerLevel.GeneralInformation:
+    case DangerLevel.None:
+      return 'Insufficient data for issuing of danger ratings, but a summary of avalanche conditions exists. Read the summary for more information.';
+    case DangerLevel.Low:
+      return '<strong>Generally safe avalanche conditions.</strong> Watch for unstable snow on isolated terrain features.';
+    case DangerLevel.Moderate:
+      return '<strong>Heightened avalanche conditions on specific terrain features.</strong> Evaluate snow and terrain carefully; identify features of concern.';
+    case DangerLevel.Considerable:
+      return '<strong>Dangerous avalanche conditions.</strong> Careful snowpack evaluation, cautious route-finding and conservative decision-making essential.';
```
floatplane, line 13: does `<strong>` work in React Native? I would have thought we'd need to return something like 
```suggestion
      return '<Caption1Semibold>Dangerous avalanche conditions.</Caption1Semibold> Careful snowpack evaluation, cautious route-finding and conservative decision-making essential.';
```

(but if it just works, then great)

hooks/useMapLayer.ts
```diff
@@ -33,6 +33,8 @@ export const useMapLayer = (center_id: AvalancheCenterID) => {
     },
     {
       enabled: !!center_id,
+      staleTime: 24 * 60 * 60 * 1000, // don't bother re-fetching for one day (in milliseconds)
+      cacheTime: Infinity, // hold on to this cached data forever
```
floatplane, line 37: if we're going to cache this request as aggressively as we are here, then we probably should remove the code from `useMapViewZones` that sets danger levels based on the results of this query (line 39). otherwise there's a strong chance that we display invalid data on the map. https://github.com/stevekuznetsov/avalanche-forecast/blob/05f5a4a8343e9cb70f12cf8ccee5e3b649e9049b/hooks/useMapViewZones.ts#L33-L42

yarn.lock
```diff
@@ -1142,14 +1142,14 @@
     pirates "^4.0.5"
     source-map-support "^0.5.16"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.12.5", "@babel/runtime@^7.14.0", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.2", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.4":
+"@babel/runtime@^7.0.0", "@babel/runtime@^7.14.0", "@babel/runtime@^7.8.4":
```
floatplane, line 1145: OOC what happened here?

floatplane (changes requested):
sending back to you for a look!

App.tsx
stevekuznetsov: It was just because I needed to get the hook to still be as top-level as possible, but within a component that had the query context - what would you suggest instead?

components/helpers/travelAdvice.ts
stevekuznetsov: It does seem to work, but I am actually surprised - potentially we're wrapping this in an HTML renderer - should we just do full-fat text components here?

yarn.lock
stevekuznetsov: Ah, I did a `yarn upgrade react-query` since I thought I was going to need to move to v4, I can revert these changes.

App.tsx
stevekuznetsov: Your suggestions in 1 sound good to me!

Re: 2 - while the changes in this PR lay that fact bare, they don't actually introduce that inversion of control. I agree with you that it's odd that a sub-item (`Menu`) owns showing the state and mutating it, and the entire app consumes the outcome of that via the context, but I'm not sure if there's a logical inconsistency there. Do you have a suggestion for how else we might approach this? Throwing a larger context or redux at the problem IMHO just puts some syntactic sugar over the thing, without really changing the data ownership or data use flows.

stevekuznetsov:
@floatplane attempted to address all of your comments!

hooks/useMapLayer.ts
stevekuznetsov: Does that basically just collapse `useMapViewZones` into `useAvalancheForecastFragments` then? Are there values of the caching tunables that you'd think better serve us? Do we need to make the decision today?

floatplane (approved):
lgtm! thanks for making those changes

hooks/useMapLayer.ts
stevekuznetsov: > useQuery has the ability to both return cached data immediately and fetch fresh data

From my read of the docs, I think that's what we do with this PR - the stale time determines how often that background fetch happens, and the other param determines how long we keep the thing in the cache in general.

PR #57, opened 2023-01-10
Animating zone card list in map view

floatplane:
I'll add some notes inline, but the easiest way to see what's going on is to look at the preview!

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -1,7 +1,7 @@
-import React, {useState} from 'react';
+import React, {useRef, useState} from 'react';
 
-import {ActivityIndicator, StyleSheet, Text, TouchableOpacity, useWindowDimensions} from 'react-native';
-import {Alert, Center, FlatList, HStack, View, VStack} from 'native-base';
+import {Animated, ActivityIndicator, StyleSheet, Text, useWindowDimensions, PanResponder, TouchableWithoutFeedback} from 'react-native';
+import {Alert, Center, HStack, View, VStack} from 'native-base';
 import MapView, {Region} from 'react-native-maps';
 import {useNavigation} from '@react-navigation/native';
 
@@ -92,13 +92,29 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
 export const CARD_SPACING = CARD_WHITESPACE / 2; // proportion of overall width that the spacing between two cards takes up
 export const CARD_MARGIN = CARD_SPACING / 2; // proportion of overall width that each card needs as a margin
 
+enum CardState {
+  Hidden = 'Hidden',
+  Docked = 'Docked',
+  Visible = 'Visible',
+}
+
+// These offsets are applied through translateY on the FlatList
+const OFFSETS = {
+  [CardState.Hidden]: 160,
+  [CardState.Docked]: 120,
+  [CardState.Visible]: 0,
+};
+
+// When a pan gesture goes beyond this distance, we animate to the final state
+const SNAP_THRESHOLD = 16;
+
 const AvalancheForecastZoneCards: React.FunctionComponent<{
   date: string;
   zones: MapViewZone[];
 }> = ({date, zones}) => {
   const {width} = useWindowDimensions();
 
-  const props = {
+  const flatListProps = {
     snapToAlignment: 'start',
     decelerationRate: 'fast',
     snapToOffsets: zones?.map((_itemData, index) => index * CARD_WIDTH * width + (index - 1) * CARD_SPACING * width),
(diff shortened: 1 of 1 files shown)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -92,13 +102,112 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
 export const CARD_SPACING = CARD_WHITESPACE / 2; // proportion of overall width that the spacing between two cards takes up
 export const CARD_MARGIN = CARD_SPACING / 2; // proportion of overall width that each card needs as a margin
 
+enum AnimatedDrawerState {
+  Hidden = 'Hidden',
+  Docked = 'Docked',
+  Visible = 'Visible',
+}
+
+class AnimatedDrawerController {
```
floatplane, line 111: using a traditional class here, since the animation state can't be used to drive React state

components/AvalancheForecastZoneMap.tsx
```diff
@@ -92,13 +102,112 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
...
+}
+
 const AvalancheForecastZoneCards: React.FunctionComponent<{
   date: string;
   zones: MapViewZone[];
 }> = ({date, zones}) => {
   const {width} = useWindowDimensions();
 
-  const props = {
+  const flatListProps = {
```
floatplane, line 210: just renamed to be more specific

components/AvalancheForecastZoneMap.tsx
```diff
@@ -131,7 +262,7 @@ const AvalancheForecastZoneCard: React.FunctionComponent<{
   const dangerColor = colorFor(zone.danger_level);
 
   return (
-    <TouchableOpacity
+    <TouchableWithoutFeedback
```
floatplane, line 265: it's very distracting to have the opacity shift while dragging the drawer up/down. we'll have to see if we need an alternate form of feedback.

components/AvalancheForecastZoneMap.tsx
```diff
@@ -92,13 +102,112 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
 export const CARD_SPACING = CARD_WHITESPACE / 2; // proportion of overall width that the spacing between two cards takes up
 export const CARD_MARGIN = CARD_SPACING / 2; // proportion of overall width that each card needs as a margin
 
+enum AnimatedDrawerState {
+  Hidden = 'Hidden',
```
floatplane, line 106: Hidden state is currently unused, but will eventually be the default - then tapping a region in the map will make the zone cards visible. See Figma mocks for more details.

This PR just focuses on the mechanics of making the zone cards (the drawer) able to be pulled up & down.

components/AvalancheForecastZoneMap.tsx
```diff
@@ -92,13 +102,112 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
 export const CARD_SPACING = CARD_WHITESPACE / 2; // proportion of overall width that the spacing between two cards takes up
 export const CARD_MARGIN = CARD_SPACING / 2; // proportion of overall width that each card needs as a margin
 
+enum AnimatedDrawerState {
```
stevekuznetsov, line 105: super-nit: 

```js
const AnimatedDrawerState = {
  Hidden = 'Hidden',
  Docked = 'Docked',
  Visible = 'Visible',
} as const;
type AnimatedDrawerState = typeof AnimatedDrawerState[keyof typeof AnimatedDrawerState];
```

?

components/AvalancheForecastZoneMap.tsx
```diff
@@ -92,13 +102,112 @@ export const CARD_WHITESPACE = (1 - CARD_WIDTH) / 2; // proportion of overall wi
...
+      return;
+    }
+
+    // Are we moving too far in the X direction? If so, treat as a scroll and stop panning the drawer
+    if (Math.abs(gestureState.dx) > AnimatedDrawerController.SNAP_THRESHOLD) {
+      this.onPanResponderRelease();
+      return;
+    }
+
+    // Detect overscroll in the invalid direction - we allow a little bit of give,
```
stevekuznetsov, line 159: Might be overkill but I feel like I've seen apps where the animation "decelerates" when being pushed out too far from the bounds, which makes this feel less bouncy.

stevekuznetsov (approved):
One tiny nit, can be done in a follow-up or ignored - feel free to merge if you're happy with this state! Code looks good and the behavior seems like what the designers expected - super awesome!

components/AvalancheForecastZoneMap.tsx
floatplane: can you point me to the reading on why enums are bad? I don't think I've seen this alternative (but I'm not a super TS expert, either).

components/AvalancheForecastZoneMap.tsx
floatplane: I'd like to get this in the designers' hands and see if it's what they were expecting - they may want something more like easing in/out, but it's pretty easy to tweak later. I didn't have a motion example to work off of.

floatplane:
> One tiny nit, can be done in a follow-up or ignored - feel free to merge if you're happy with this state! Code looks good and the behavior seems like what the designers expected - super awesome!

I'm going to go ahead and merge, and then address followups later (map screen is definitely not code complete yet)

PR #59, opened 2023-01-11
Update AvalancheDangerTable styling

floatplane:
This gets the danger table close to what's in Figma. I expect it could shift somewhat, since the Figma design is actually just a screenshot 😬 - not sure I've picked the right text formats, but that's a quick fix.

Known issue: the Outlook tab needs to be added, that'll be in a future PR.

image


stevekuznetsov (commented):
All the ugly flex stuff is gone! So nice!

PR #60, opened 2023-01-11
AvalancheForecastZoneCard: render 'general information' better

stevekuznetsov:
Similarly to 'no rating', we just want the words here, not the numerical value.

Follows up on #58


stevekuznetsov:
@floatplane second commit in here did some bash surgery to fix goofy quotation breakages

PR #63, opened 2023-01-11
Make preloading imperative, not hook-driven

floatplane:
~still need to remove logging / make it optional~

Logging is now opt-in - this should be ready for review


floatplane:
looks like the failure was a transient Expo failure. I'm going to merge this so I can rebase other work on it - will address followups in a future PR!

PR #64, opened 2023-01-11
Fix Android crash - invalid value passed to justifyContent

floatplane:
Need to get a preview build out of this PR before I feel totally confident, but this seems to be the very surprising source


stevekuznetsov:
With this build, the app loads for me, but the cards render weird and I crash on opening a forecast.
Screenshot_20230111-163834_Expo Go

stevekuznetsov:
The QR code doesn't seem to work, so I am running locally from this branch. I'm getting a `justifyContent` crash, but the logs don't show up in my terminal and the app hard-crashes Expo Go ...

floatplane:
looks like I used `stretch` somewhere else, hold please

floatplane:
@stevekuznetsov things are working for me on my Android device, although loading the forecast screen is incredibly slow (but quite fast on iOS)

stevekuznetsov:
@floatplane yep, seeing the same for me - it all works now but boy is that splash screen kicking around for a long time

PR #71, opened 2023-01-13
Break nativeBase dependencies: Text, Switch, ScrollView, useToken

floatplane:
Decided it's probably better to merge this work a bit at a time, rather than one mega PR! Here's what's in this PR:

- I cloned the NativeBase theme object into `theme/index.ts` and `theme/colors.ts`. I added a color lookup function so that we can keep resolving things like `darkText` and `light.300`, and applied those where I saw them. this also meant we could stop calling `useTheme`.
- I removed all references to the NativeBase Text component, in favor of our predefined styled versions (which now uses its own wrapper around RN's Text component)
- I added support for NB-style props on text, e.g. setting `color` directly instead of via `style`
- I removed a couple of references to NB components that did the same thing as their React Native counterparts (ScrollView, Switch)

Things that I think could go wrong with this PR (none of them major IMO, but calling them out):
- I might have picked the wrong wrapper when replacing things like `Heading` or `<Text bold` - things will look visually different in spots
- I might have missed wrapping color references like `"light.300"` in `colorLookup`, in which case the color would be ignored


stevekuznetsov:
Things look bigger and bolder than before, but I think that's `Heading` being translated to our chosen font. TBH looks pretty jarringly-large in many places, but orthogonal to this PR I guess. On some collapsible cards the title crowds out the chevron for collapsing before it wraps the line:

File (2)

stevekuznetsov:
Up to you how you want to draw the finish line for this particular PR! The approach & changes here LGTM.

floatplane:
Changed the obs page to use Title1Black for card sections instead of FeatureTitleBlack, looks better

image

Issue #66, opened 2023-01-12
Android: Avalanche Forecast Cards Rest Too Low


stevekuznetsov:
Resting state of the cards on the map screen is too low - it doesn't show the zone name. I can still pull them up correctly, but when pushing them down they go back to resting too low. 

I'm on a Samsung Galaxy S10e

![Screenshot_20230111-163834_Expo Go](https://user-images.githubusercontent.com/7328370/211944431-5ec51814-084e-4bc4-8df8-3f044b47ee40.jpg)

floatplane:
@stevekuznetsov are you still seeing this in recent builds?

stevekuznetsov:
Nope!

PR #72, opened 2023-01-16
Use our own wrapper around View, remove NativeBase wrapper

floatplane:
This mimics the NativeBase View wrapper by adding style props as top-level props to View, and by adding some handy shortcuts to those props, e.g. `mx` for `marginHorizontal`.

I'll add some notes inline!

### DIFF
--- a/components/AvalancheDangerTable.tsx
+++ b/components/AvalancheDangerTable.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 
-import {HStack, View, VStack} from 'native-base';
+import {HStack, VStack} from 'native-base';
 
 import {AvalancheDangerForecast, ElevationBandNames} from 'types/nationalAvalancheCenter';
+import {View} from 'components/core/View';
 import {dangerText} from 'components/helpers/dangerText';
 import {utcDateToLocalDateString} from 'utils/date';
 import {Body, Caption1, Caption1Semibold} from 'components/text';
@@ -22,12 +23,12 @@ export interface AvalancheDangerTableProps {
 export const AvalancheDangerTable: React.FunctionComponent<AvalancheDangerTableProps> = ({date, forecast, elevation_band_names, size}: AvalancheDangerTableProps) => {
   const {height, marginLeft, paddingTop} = {
     main: {
-      height: '200',
+      height: 200,
       paddingTop: '13px',
       marginLeft: -6,
     },
     outlook: {
-      height: '150',
+      height: 150,
       paddingTop: '10px',
       marginLeft: 16,
     },
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -11,10 +11,11 @@ import {
   PanResponderGestureState,
   GestureResponderEvent,
 } from 'react-native';
-import {Alert, Center, HStack, View, VStack} from 'native-base';
+import {Alert, Center, HStack, VStack} from 'native-base';
 import MapView, {Region} from 'react-native-maps';
 import {useNavigation} from '@react-navigation/native';
 
+import {View} from 'components/core/View';
 import {DangerScale} from 'components/DangerScale';
 import {AvalancheCenterID, DangerLevel} from 'types/nationalAvalancheCenter';
 import {AvalancheCenterForecastZonePolygons} from './AvalancheCenterForecastZonePolygons';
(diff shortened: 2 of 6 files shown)


components/AvalancheDangerTable.tsx
```diff
@@ -22,12 +23,12 @@ export interface AvalancheDangerTableProps {
 export const AvalancheDangerTable: React.FunctionComponent<AvalancheDangerTableProps> = ({date, forecast, elevation_band_names, size}: AvalancheDangerTableProps) => {
   const {height, marginLeft, paddingTop} = {
     main: {
-      height: '200',
+      height: 200,
```
floatplane, line 26: NativeBase would silently convert these values to numbers - that seems like a bad practice for me. Now we validate and throw. `'200pt'` and `200` are valid values for React Native, `'200'` is not.

components/AvalancheForecastZoneMap.tsx
```diff
@@ -75,7 +76,9 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
         {isReady && <AvalancheCenterForecastZonePolygons key={center} center_id={center} setRegion={setRegion} date={date} />}
       </MapView>
       <SafeAreaView>
-        <DangerScale px="4" width="100%" position="absolute" top="12" />
+        <View flex={1}>
+          <DangerScale px={4} width="100%" position="absolute" top={12} />
+        </View>
```
floatplane, line 81: not sure why this worked before, but in this PR the absolute positioning stopped respecting the safe area - it needs to be absolute positioning relative to a View that's sized to the safe area.

it's possible that #66 is caused by a similar issue, not sure yet!

components/AvalancheForecastZoneMap.tsx
```diff
@@ -273,7 +276,7 @@ const AvalancheForecastZoneCard: React.FunctionComponent<{
         });
       }}>
       <VStack borderRadius={8} bg="white" width={width * CARD_WIDTH} marginX={CARD_MARGIN * width}>
-        <View height="8px" width="100%" bg={dangerColor.string()} borderTopRadius={8} pb={0} />
+        <View height={8} width="100%" bg={dangerColor.string()} borderTopLeftRadius={8} borderTopRightRadius={8} pb={0} />
```
floatplane, line 279: I didn't bother adding a facility for having one pseudo prop (`borderTopRadius`) set two props in the final style. If it becomes common I'll consider it, but I don't see the need yet.

components/core/View.tsx
```diff
@@ -0,0 +1,173 @@
+import * as React from 'react';
+
+import {View as RNView, ViewProps as RNViewProps, ViewStyle as RNViewStyle} from 'react-native';
+
+import {colorLookup} from 'theme';
+
+class ViewStyleProps {
```
floatplane, line 7: This is declared in a class so that we can instantiate it and get the list of keys into an array. You can't do that with an interface. Lord knows I tried 😄

components/core/View.tsx
```diff
@@ -0,0 +1,173 @@
...
+      if (['backgroundColor'].includes(prop) && typeof value === 'string') {
+        value = colorLookup(value);
+      }
+      validateProp(prop, value);
+      style[prop] = value;
+    } else {
+      resolvedProps[prop] = value;
+    }
+  });
+  // console.log(JSON.stringify(resolvedProps, null, 2));
```
floatplane, line 171: will remove this

components/core/View.tsx
```diff
@@ -0,0 +1,173 @@
...
+      const error = `Invalid string value "${value}" for property ${prop}: string dimensions must specify pt or %`;
+      console.error(error);
+      throw new Error(error);
+    }
+  }
+};
+
+export interface ViewProps extends RNViewProps, ViewStyleProps, ViewAliasProps {}
+export const View: React.FC<ViewProps> = ({children, style = {}, ...props}) => {
+  const resolvedProps: RNViewProps = {style};
```
floatplane, line 158: the logic in this method is slightly different from the way I implemented the TextWrapper - in this component, all top-level style properties are assumed to overwrite any `style` entries that are set.

generally I would be suspicious of code that's setting the same property through `style` and through top-level props - we should avoid that unless there's an amazing reason to do so.

I'll do a revision on Text to make its behavior the same as this method.

stevekuznetsov: Either here or as a follow-up, this method sounds like a good candidate for a couple unit tests to document & lock down the expected behavior.

stevekuznetsov (commented):
Looks good!

components/core/View.tsx
```diff
@@ -0,0 +1,172 @@
...
+  'paddingLeft',
+  'paddingRight',
+  'paddingTop',
+  'paddingVertical',
+  'right',
+  'top',
+  'width',
+];
+
+const validateProp = (prop: ViewStyleProp, value): void => {
```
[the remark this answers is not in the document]
floatplane: I am not sure how you'd do this via types, but if you have ideas LMK. The actual underlying type is `number | string | undefined`, but if you try something that typechecks (like `width="whatever"`), you'll get a very unhelpful runtime backtrace in native code:


This validation turns it into something more helpful, and you can set an unhandled exception breakpoint in the JS debugger to track it down:

components/core/View.tsx
floatplane: fixed in #75

PR #73, opened 2023-01-16
Finish removing NativeBase from app

floatplane:
Builds on #72 and does the following to get Android happier:

- add implementations of useful NativeBase components:
  - VStack/HStack
  - Center
  - Divider
- replace use of Alert
- memoize core components
- fix for forecast cards not receiving touches on Android

Android build is still not zippy yet, but it is much better.

### DIFF
--- a/components/AvalancheCenterSelector.tsx
+++ b/components/AvalancheCenterSelector.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {Box, VStack, HStack} from 'native-base';
+import {Box, HStack} from 'native-base';
 import {Text, SectionList, SectionListData, StyleSheet, TouchableOpacity, ActivityIndicator, useWindowDimensions} from 'react-native';
 import {useNavigation} from '@react-navigation/native';
 
@@ -9,6 +9,7 @@ import {useAvalancheCenterMetadata} from 'hooks/useAvalancheCenterMetadata';
 import {TabNavigationProps} from 'routes';
 import {AvalancheCenterID} from '../types/nationalAvalancheCenter';
 import {Body} from './text';
+import {VStack} from 'components/core';
 
 const avalancheCenterIDsByType: SectionListData<AvalancheCenterID>[] = [
   {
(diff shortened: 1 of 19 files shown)


components/AvalancheCenterSelector.tsx
```diff
@@ -59,50 +59,50 @@ export const AvalancheCenterCard: React.FunctionComponent<AvalancheCenterCardPro
   }
   if (isError) {
     return (
-      <Box>
-        <Box bg="light.100">
+      <View>
```
floatplane, line 62: `Box` had to be one of the more useless NativeBase wrappers. It's just a `View`!

components/AvalancheForecastZoneMap.tsx
```diff
@@ -82,18 +84,20 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
...
-          </Alert>
+        <Center width="100%" position="absolute" bottom={6}>
+          <VStack space={8}>
+            <Center bg={COLORS['warning.200']} px={24} py={16} borderRadius={4}>
+              <HStack space={8} flexShrink={1}>
+                <FontAwesome5 name="exclamation-triangle" size={16} color={COLORS['warning.700']} />
+                <Body>Unable to load forecast data</Body>
+              </HStack>
+            </Center>
+          </VStack>
```
floatplane, line 100: `Alert` on top, hand-crafted version below. Now the icon and text are actually centered correctly, and it's an actual warning icon. I'll refactor this once we need something similar elsewhere.

image

components/AvalancheForecastZoneMap.tsx
```diff
@@ -230,7 +234,7 @@ const AvalancheForecastZoneCards: React.FunctionComponent<{
 
   const panResponder = useRef(
     PanResponder.create({
-      onMoveShouldSetPanResponder: () => true,
+      onMoveShouldSetPanResponder: (_event, {dx, dy}) => dx > 0 || dy > 0,
```
floatplane, line 237: this fix courtesy of https://stackoverflow.com/questions/68377590/react-native-panresponder-onrelease-event-stealing-touchable-opacity-onpress-eve

PR #78, opened 2023-01-16
Map view + selection behavior

floatplane:
This implements selection behaviors as outlined in Figma. See the video in #mobile-design for details.


.github/workflows/publish.yaml
```diff
@@ -68,6 +68,8 @@ jobs:
         run: eas build --non-interactive --platform android --profile preview
 
   build-ios:
+    if: ${{ false }} # disable iOS builds until we migrate to a different Expo tier
+
```
floatplane, line 72: iOS builds are hanging and reducing our ability to run concurrent builds

components/AvalancheForecastZonePolygon.tsx
```diff
@@ -85,59 +65,27 @@ export const updateRegionToContain = (previous: Region, coordinates: LatLng[]):
...
-  }
-
-  let dangerLevel: DangerLevel = DangerLevel.None;
-  if (forecast) {
-    const currentDanger: AvalancheDangerForecast | undefined = forecast.danger.find(item => item.valid_day === ForecastPeriod.Current);
-    if (currentDanger) {
-      dangerLevel = Math.max(currentDanger.lower, currentDanger.middle, currentDanger.upper);
-    }
-  }
-
```
floatplane, line 126: Polygons don't need to fetch their own data, display loading errors, etc. The parent component (AFZoneMap) has already done the hard work.

hooks/useMapViewZones.ts
```diff
@@ -15,6 +15,7 @@ export type MapViewZone = {
   start_date: string | null;
   end_date: string | null;
   geometry?: FeatureComponent;
+  fillOpacity: number;
```
floatplane, line 18: just need one more thing to pass down to the polygons

components/AvalancheForecastZoneMap.tsx
```diff
@@ -63,8 +68,9 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
         onLayout={setReady}
         zoomEnabled={true}
         scrollEnabled={true}
-        provider={'google'}>
-        {isReady && <AvalancheCenterForecastZonePolygons key={center} center_id={center} setRegion={setRegion} date={date} />}
```
floatplane, line 67: The Polygons (plural) wrapper really wasn't adding value - just extra plumbing.

App.tsx
```diff
@@ -134,10 +134,10 @@ const BaseApp: React.FunctionComponent<{
...
 
   const onLayoutRootView = useCallback(async () => {
-    if (fontsLoaded) {
-      await SplashScreen.hideAsync();
-    }
-  }, [fontsLoaded]);
+    // This callback won't execute until fontsLoaded is true, because
+    // otherwise we won't render the view that triggers this callback
+    await SplashScreen.hideAsync();
+  }, []);
```
floatplane, line 140: didn't need to re-evaluate `useCallback` because it would never be executed in the `fontsLoaded === false` case

.github/workflows/publish.yaml
stevekuznetsov: @floatplane let's make sure this gets surfaced this week in the mtg

PR #81, opened 2023-01-17
Create a media item carousel

floatplane:
The final design of this is subject to tweaks, but this is the first version of Carousel component, which can:

- render one or more `MediaItem`s in a horizontal list
- optionally render the caption below the image

Next up for this component:
- render a modal on tap
- allow zoom/pan on image in the modal
- allow swiping to next/previous in modal
- video playback (?!)

# screenshots

rendering a single full-width image in the problem discussion section:
image

rendering a set of thumbnails at the end of the forecast:

image


floatplane:
# Other screenshots

1) loading
image

2) load error
image

PR #84, opened 2023-01-19
Improve date handling

### LINKED ISSUE #42: Expired forecasts causing problems?

### LINKED ISSUE #53: We are playing fast and loose with dates (and maybe the API is too)
Note the discrepancy here: 

forecast summary from center endpoint | forecast 
--- | ---
image | image

The LHS is rendering UTC dates without conversion - the RHS is rendering PST. Yet AFAIK they are both going through the same dateToString conversion. _Something_ is screwed up here.

Suggestions for improvement:
- pass around Dates as much as possible, avoid using date strings everywhere. Only render to string for presentation to the user
- parse date strings in API fields to Dates, and rename fields to reflect whether they're UTC or not (i.e. `expires_date_utc: Date`)

floatplane:
This was a fun one!

- Generally, make us use Date objects instead of date strings everywhere (fixes #53)
  - **except** in React Navigation, which wants all route params to be JSON serializable 😢 
- Add a new `apiDateString` method to return date strings that are suitable to be query params and route params (fixes #42, allows us to remove hack of extra days in the date windows)
- Update Zod schema to fix API date strings that are missing timezone information, and to parse time values into Date objects automatically
- Added tests for all the date-related utility methods in `date.tsx`

What did I learn while doing this?

- Don't use `format*` to create strings that you'll pass to an API. They format the time into the local time zone, which is almost never what you want.  You can just use `apiDateString`, which internally uses `Date.toISOString()` to get an ISO formatted string in UTC (zero offset).
- You don't need `parseISO`, `new Date(str)` does the same thing

I'll add more notes inline.


components/Observations.tsx
```diff
@@ -51,6 +50,7 @@ export const Observations: React.FunctionComponent<{
     );
   }
   if (!observations.getObservationList || observations.getObservationList.length === 0) {
+    // TODO: when cleaning this up, fix it so that it renders the date in the user's locale, not UTC date
```
floatplane, line 53: I ran this at around 5 PM on 1/18 and the error message referred to the next day, since in UTC it was already the next day. This might be confusing to a user: 

image

package.json
```diff
@@ -64,7 +64,7 @@
     "react-native-svg": "13.4.0",
     "react-query": "^3.39.2",
     "sentry-expo": "~6.0.0",
-    "zod": "^3.19.1"
+    "zod": "^3.20.0"
```
floatplane, line 67: I think this version bump turned out to be unnecessary - I thought I was going to use `z.coerce.date()` but ended up going a different way. I can revert if you want, but it seems like some nice new features were added.

routes.ts
```diff
@@ -3,46 +3,46 @@ import {NativeStackNavigationProp} from '@react-navigation/native-stack';
 import {AvalancheCenterID} from './types/nationalAvalancheCenter';
 
 export type TabNavigatorParamList = {
-  Home: {center_id: AvalancheCenterID; date: string};
-  'Weather Data': {center_id: AvalancheCenterID; date: string};
-  Observations: {center_id: AvalancheCenterID; date: string};
+  Home: {center_id: AvalancheCenterID; dateString: string};
```
floatplane, line 6: tried to be explicit about names whenever a date was in string format

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -16,6 +17,11 @@ import {
   Units,
 } from './enums';
 
+const datePipeline = z
+  .string()
+  .transform(s => fixMalformedISO8601DateString(s))
+  .transform(s => new Date(s));
```
floatplane, line 23: 1. not married to the name `datePipeline`, LMK if you have suggestions
2. probably marginally faster to pass a single function to `.transform`, I was trying to be super-explicit here but I can be faster instead

utils/date.tsx
```diff
@@ -1,13 +1,24 @@
-import {format, parseISO} from 'date-fns';
+import {format} from 'date-fns';
 
-export const utcDateToLocalTimeString = (dateString: string | undefined): string => {
-  if (!dateString) {
+const MISSING_TIMEZONE = /^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?$/;
+
+// Some avalanche.org endpoints return dates without timezone information! This method is used to detect and fix them, by assuming they're UTC.
+export const fixMalformedISO8601DateString = (date: string) => (MISSING_TIMEZONE.test(date) ? `${date}Z` : date);
```
floatplane, line 6: I tried to be very conservative here: the input has to match a specific date and time format and not have any timezone information at the end. Otherwise, we just return the input string. There's a bunch of tests but LMK if there are others you think should be in there.

App.tsx
```diff
@@ -165,13 +166,13 @@ const BaseApp: React.FunctionComponent<{
                   }
                 },
               })}>
-              <TabNavigator.Screen name="Home" initialParams={{center_id: avalancheCenterId, date: date}}>
+              <TabNavigator.Screen name="Home" initialParams={{center_id: avalancheCenterId, dateString}}>
```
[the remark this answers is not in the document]
floatplane: If you are setting an object key with a variable of the same name, you’re allowed to just type the name once, instead of `dateString: dateString`. 

I guess this is ES6 syntax maybe? It’s not just typescript…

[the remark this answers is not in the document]
floatplane: https://eslint.org/docs/latest/rules/object-shorthand

PR #86, opened 2023-01-20
Add InfoTooltip component and dialog content from NAC

floatplane:
In this PR we have
- a new `Button` component that renders like the AntDesign button, and looks correct cross-platform
- a new `InfoTooltip` component that can be placed next to text, which displays a dialog when tapped
image

- the HTML content from NAC for displaying in the dialogs:
image

---

Here it is in action:

https://user-images.githubusercontent.com/101196/213781216-c6601463-83e8-468d-a54d-0f037ce2148e.mp4


components/content/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+import {colorLookup} from 'theme';
+
+export interface ButtonProps extends ViewProps {
+  //  title: string;
+  color?: ColorValue;
+  onPress?: (event: GestureResponderEvent) => void;
+}
+
+export const Button: React.FC<ButtonProps> = ({color = 'color-primary', children, onPress, ...props}) => {
+  const [pressedBgColor] = useState<ColorValue>(tinycolor(colorLookup(color)).setAlpha(0.2).toRgbString());
```
floatplane, line 14: `tinycolor` library is used to parse a color string like "#1677FF", apply an opacity and output it as `rgba(r, g, b, opacity)`

components/text/HTML.tsx
```diff
@@ -25,7 +25,7 @@ export interface HTMLRendererConfigProps {
   baseStyle?: MixedStyleDeclaration;
 }
 
-const baseStyleDefaults = {
+const baseStyleDefaults: MixedStyleDeclaration = {
```
floatplane, line 28: didn't really _need_ to do this but it doesn't hurt

components/text/index.tsx
```diff
@@ -48,7 +48,10 @@ export const Title3: React.FunctionComponent<TextWrapperProps> = props => (
 export const Title3Semibold: React.FunctionComponent<TextWrapperProps> = props => <Title3 fontFamily="Lato_700Bold" {...props} />;
 export const Title3Black: React.FunctionComponent<TextWrapperProps> = props => <Title3 fontFamily="Lato_900Black" {...props} />;
 
-export const Body: React.FunctionComponent<TextWrapperProps> = props => <TextWrapper fontSize={16} lineHeight={22} fontFamily="Lato_400Regular" letterSpacing={-0.43} {...props} />;
+export const bodySize = 16;
```
floatplane, line 51: needed to make icon the same size as text

components/content/Button.tsx
```diff
@@ -0,0 +1,26 @@
+import {Center, View, ViewProps} from 'components/core';
+import React, {useState} from 'react';
+import {ColorValue, GestureResponderEvent, Pressable, Text} from 'react-native';
+import tinycolor from 'tinycolor2';
+import {colorLookup} from 'theme';
+
+export interface ButtonProps extends ViewProps {
+  //  title: string;
```
stevekuznetsov, line 8: holdover from a previous iteration?

content/helpStrings.ts
```diff
@@ -0,0 +1,18 @@
+export default {
```
stevekuznetsov, line 1: I remember seeing the source for all of their tooltips - can't seem to find it now but let's file away a task to just bulk-import all of them.

floatplane: there's a task in Asana for adding them

components/content/Button.tsx
floatplane: yup, will fix

PR #95, opened 2023-01-23
hooks: include host in all query keys

stevekuznetsov:
Now that we support asking the staging or production API for this data, we need to include the host in our query keys so that our caches are separate.


floatplane (changes requested):
we need this in `useMapLayer` as well, I think?

stevekuznetsov:
@floatplane updated!

Issue #91, opened 2023-01-23
Something isn't working right after midnight UTC


floatplane:
Right after 4 pm PST (midnight UTC) we sometimes fail to load a forecast for a region.

Today (Sunday 1/22/23 4:30 PM PST | Monday 1/23/23 12:30 AM UTC) I'm not getting a forecast for East Slopes South.

floatplane:
Ok, now I understand why this is happening!

1. Load the app at Monday 2023-01-23 4:01 PM PST, which is Tuesday 2023-01-24 00:01 UTC
2. App startup looks at the clock and says our `defaultDate` is Tuesday 2023-01-24, based on UTC time
3.  `useAvalancheForecastFragments` tried to get forecasts in a 1-day window around the current date, so it looks at `https://api.avalanche.org/v2/public/products?avalanche_center_id=NWAC&date_start=2023-01-23&date_end=2023-01-25`
4. In that query, no forecasts were found for the East Slopes South zone! It turns out that the most recent forecast was _created_ just before UTC midnight on the 22nd (`"created_at": "2023-01-22T23:34:23+00:00"`), even though it wasn't _published_ until a few hours later (`"published_time": "2023-01-23T02:00:00+00:00"`). `date_start` was excluding this forecast, seemingly because it's filtering on `created_at`. But if we could have gotten that forecast in our query, we would have accepted it, because it didn't expire until later on the 24th.
5. The workaround, then, is to expand our lookback window by 1 day in `useAvalancheForecastFragments`

PR #96, opened 2023-01-24
Fix for issue #91, ~with~ without test

floatplane:
@stevekuznetsov here's a slightly more targeted take on #94, because it only widens the window for `start_date`. also, there's a test for the case that was failing in #91, to make sure that it works with the changed query.

the test is hopefully useful in the future if we want to get other tests on our hooks. it took me a while to get all the different issues sorted out (mocking axios, avoiding some of react-query's retry behavior, etc).


hooks/useAvalancheForecastFragment.ts
```diff
@@ -15,13 +19,14 @@ export const useAvalancheForecastFragment = (center_id: AvalancheCenterID, forec
     },
     {
       enabled: !!fragments,
-      staleTime: 60 * 60 * 1000, // re-fetch in the background once an hour (in milliseconds)
-      cacheTime: 24 * 60 * 60 * 1000, // hold on to this cached data for a day (in milliseconds)
+      staleTime: options?.skipCacheForTests ? 0 : 60 * 60 * 1000, // re-fetch in the background once an hour (in milliseconds)
```
stevekuznetsov, line 22: Are we certain there is no approach that does not require this? IMHO production code with conditionals for testing is a huge anti-pattern.

stevekuznetsov (commented):
I like the approach, but I wonder - what are we testing here? In my opinion, we should be able to test:

1. that we hit the right API endpoint with the right parameters
2. that we post-process the data from the API correctly

An "end-to-end" test here requires us to entirely mock the server, so I posit that it does not add huge amounts of value, necessarily. For instance, the tests in this PR actually do not require your changes in the hook to pass, for that reason:

```
$ git log -n 1
commit 95e843e672f280c5048e490516b39ae157119e27 (HEAD, upstream/date-hack)
Author: Brian Sharon <brian@nwac.us>
Date:   Mon Jan 23 17:04:35 2023 -0800

    Fix for issue #91, with test
$ git diff
diff --git a/hooks/useAvalancheForecastFragments.ts b/hooks/useAvalancheForecastFragments.ts
index df33557..62f7007 100644
--- a/hooks/useAvalancheForecastFragments.ts
+++ b/hooks/useAvalancheForecastFragments.ts
@@ -43,7 +43,7 @@ const fetchAvalancheForecastFragments = async (nationalAvalancheCenterHost: stri
   const {data} = await axios.get(url, {
     params: {
       avalanche_center_id: center_id,
-      date_start: apiDateString(sub(date, {days: 2})),
+      date_start: apiDateString(sub(date, {days: 1})),
       date_end: apiDateString(add(date, {days: 1})),
     },
   });
$ yarn test
yarn run v1.22.19
$ jest
 PASS  tests/types/nationalAvalancheCenter/enums.test.ts
 PASS  components/core/View.test.tsx
 PASS  utils/date.test.tsx
 PASS  hooks/useAvalancheForecastFragment.test.tsx

Test Suites: 4 passed, 4 total
Tests:       25 passed, 25 total
Snapshots:   0 total
Time:        1.575 s, estimated 3 s
Ran all test suites.
Done in 2.39s.
```

hooks/useAvalancheForecastFragment.ts
floatplane: I think the alternative is to make callers always pass an explicit cache time. This feels like a reasonable tradeoff to me but open to alternatives 🤷 

It turns out the way cache expiration is implemented in React Query is via `setTimeout`. Jest will wait for all pending async tasks to complete if you have a lingering `setTimeout`, and then the test itself will eventually timeout after 5 seconds (Jest's default test timeout) and fail.

stevekuznetsov:
Ultimately the bug was in our use of (and understanding of) the API provided, which is why I guess a test on our end is not really that instructive. I'm not sure that the test as written gives us more useful coverage than we had before, and at the cost of having test code in production, so my inclination would be to drop that specific part of the PR - WDYT?

floatplane:
sure, that's fine, although changing cache behavior in tests is going to come up again. our hooks are both crucial to the functioning of the product **and** very undertested.

stevekuznetsov:
Sure, but I guess my point is more broad - we can assume `react-query` works and test up to that boundary, but not through it. I'm 100% open to testing in there, but I am not sure in principle what we'd gain. If we are to test that we pass the right parameters to the right endpoints, do we still need the behavioral bits?

PR #98, opened 2023-01-24
*: fetch forecasts using the latest product API

stevekuznetsov:
hooks: use a mapping custom for options

This is common in all of the react-query documentation for capturing
options passed to a query. No reason we should buck the trend.

---

*: fetch forecasts using the latest product API

Ultimately, we will need two types of views:

1. "show me the currently-valid products for a zone"
2. "show me the valid products for a zone at some point in the past"

Due to quirks of how the date filtering works on the products API from
the NAC (filtering on the publication date, not the date for which the
product is valid), we need to split these cases up. This PR does the
first pass to allow us to fetch the currently-valid forecast.
Follow-up PRs will:

1. update our routes to not contain a date, and resolve "now" when we're
   in the component
2. add a new route for a particular time in history
3. add warnings & synopses to the data we fetch for a forecast date

---

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -36,7 +36,7 @@ import {TelemetryTabScreen} from 'components/screens/TelemetryScreen';
 import {AvalancheCenterID} from './types/nationalAvalancheCenter';
 import {prefetchAllActiveForecasts} from './network/prefetchAllActiveForecasts';
 import {HTMLRendererConfig} from 'components/text/HTML';
-import {apiDateString} from 'utils/date';
+import {formatInTimeZone} from 'date-fns-tz';
 
 // The SplashScreen stays up until we've loaded all of our fonts and other assets
 SplashScreen.preventAutoHideAsync();
@@ -110,7 +110,7 @@ const BaseApp: React.FunctionComponent<{
 }> = ({staging, setStaging}) => {
   const [avalancheCenterId, setAvalancheCenterId] = React.useState(Constants.expoConfig.extra.avalanche_center as AvalancheCenterID);
   const [date] = React.useState<Date>(defaultDate);
-  const dateString = apiDateString(date);
+  const dateString = formatInTimeZone(date, 'UTC', 'yyyy-MM-dd HH:mm:ssXXX');
 
   const {nationalAvalancheCenterHost} = React.useContext<ClientProps>(ClientContext);
   const queryClient = useQueryClient();
--- a/App.tsx
+++ b/App.tsx
@@ -1,7 +1,7 @@
 import React, {useCallback, useEffect} from 'react';
 
 import {AppStateStatus, Platform, StyleSheet, View} from 'react-native';
-import {NavigationContainer} from '@react-navigation/native';
+import {NavigationContainer, ParamListBase, RouteProp} from '@react-navigation/native';
 import {createBottomTabNavigator} from '@react-navigation/bottom-tabs';
 import {SafeAreaProvider} from 'react-native-safe-area-context';
 import {AntDesign} from '@expo/vector-icons';
@@ -186,7 +186,11 @@ const BaseApp: React.FunctionComponent<{
   );
 };
 
-const withParams = (state, params) => {
+/* eslint-disable @typescript-eslint/no-explicit-any */
+function withParams<U extends ParamListBase, V extends keyof U>(
+  state: {route: RouteProp<U, V>; navigation: any},
+  params: Readonly<U[V]>,
+): {route: RouteProp<U, V>; navigation: any} {
   return {
     ...state,
     route: {
@@ -197,6 +201,6 @@ const withParams = (state, params) => {
       },
     },
   };
-};
+}
 
 export default App;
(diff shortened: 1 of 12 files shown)


utils/date.tsx
```diff
@@ -9,6 +9,17 @@ export const fixMalformedISO8601DateString = (date: string) => (MISSING_TIMEZONE
 // The National Avalanche Center API expects 'YYYY-MM-DD' date-strings in query parameters, and it operates in UTC.
 export const apiDateString = (date: Date) => formatInTimeZone(date, 'UTC', 'yyyy-MM-dd');
 
+// Forecasts expire in the middle of a calendar day, so the forecast that we expect to be valid at any given time during
+// the day changes based on the relationship of that time to the expected expiry time, as recorded for the avalanche
+// center.
```
floatplane, line 14: I think this comment could be improved by just adding a concrete example or two, something like 

requestedTime | expiryTimeHours | returns
--- | --- | ---
datetime before expiry hours | 18 | result
datetime after expiry hours | 18 | result

utils/date.tsx
```diff
@@ -9,6 +9,17 @@ export const fixMalformedISO8601DateString = (date: string) => (MISSING_TIMEZONE
...
+export const nominalForecastDate = (requestedTime: Date, expiryTimeHours: number): string => {
+  if (isAfter(add(startOfDay(requestedTime), {hours: expiryTimeHours}), requestedTime)) {
+    return apiDateString(startOfDay(requestedTime));
+  } else {
+    return apiDateString(add(startOfDay(requestedTime), {days: 1}));
+  }
```
floatplane, line 20: I would have expected the two return values here to be reversed - what am I missing? (this might be where the comment would have helped me)

utils/date.tsx
```diff
@@ -9,6 +9,17 @@ export const fixMalformedISO8601DateString = (date: string) => (MISSING_TIMEZONE
 // The National Avalanche Center API expects 'YYYY-MM-DD' date-strings in query parameters, and it operates in UTC.
 export const apiDateString = (date: Date) => formatInTimeZone(date, 'UTC', 'yyyy-MM-dd');
 
+// Forecasts expire in the middle of a calendar day, so the forecast that we expect to be valid at any given time during
+// the day changes based on the relationship of that time to the expected expiry time, as recorded for the avalanche
+// center.
+export const nominalForecastDate = (requestedTime: Date, expiryTimeHours: number): string => {
+  if (isAfter(add(startOfDay(requestedTime), {hours: expiryTimeHours}), requestedTime)) {
```
floatplane, line 16: optional: maybe pull `startOfDay(requestedTime)` into a variable? would reduce some of the nesting in these expressions

utils/date.tsx
floatplane: also, a couple of tests in `date.test.tsx` would probably help clarify this

App.tsx
```diff
@@ -110,7 +109,7 @@ const BaseApp: React.FunctionComponent<{
 }> = ({staging, setStaging}) => {
   const [avalancheCenterId, setAvalancheCenterId] = React.useState(Constants.expoConfig.extra.avalanche_center as AvalancheCenterID);
   const [date] = React.useState<Date>(defaultDate);
-  const dateString = apiDateString(date);
+  const dateString = date.toISOString();
```
floatplane, line 112: this is a different string format than what all the other routes are using, is that problematic? or are things generally working and the follow up PRs will resolve this?

floatplane (changes requested):
Generally seems good! I had some questions about `nominalForecastDate`

stevekuznetsov:
@floatplane updated the function to be clearer, added tests

stevekuznetsov:
@floatplane blargh, `startOfDay()` is locale-specific too. Fixed.

App.tsx
floatplane: I mean the react navigation `dateString` field used internally

floatplane:
if you fix up commits and force push then I have to look at the whole diff again 😢

App.tsx
```diff
@@ -110,7 +110,7 @@ const BaseApp: React.FunctionComponent<{
 }> = ({staging, setStaging}) => {
   const [avalancheCenterId, setAvalancheCenterId] = React.useState(Constants.expoConfig.extra.avalanche_center as AvalancheCenterID);
   const [date] = React.useState<Date>(defaultDate);
-  const dateString = apiDateString(date);
+  const dateString = formatInTimeZone(date, 'UTC', 'yyyy-MM-dd HH:mm:ssXXX');
```
floatplane, line 113: why would you not extract this into a common helper? this seems like a maintenance headache for future you

floatplane (approved):

App.tsx
stevekuznetsov: As far as I was able to tell, this one line is the source of truth for every one of those `dateString` values, no?

App.tsx
stevekuznetsov: done :)

App.tsx
floatplane: I would say no, not exactly! The `dateString` params get parsed into Date objects when a screen is entered, so that most downstream/child components deal with Date objects internally. 

When a child component decides we have to navigate somewhere else (in response to a user action or whatever), the current date is converted back to a string for react-navigation purposes. 

This was all done in #84

PR #100, opened 2023-01-24
map: ensure zones are not obstructed by other elements

stevekuznetsov:
This needs to be integrated into the animation controller, and for *some* reason the offset callback broke swipe gestures on there, ...

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -106,7 +106,7 @@ enum AnimatedDrawerState {
   Visible = 'Visible',
 }
 
-class AnimatedDrawerController {
+class AnimatedMapWithDrawerController {
   // These offsets are applied through translateY on the FlatList
   static readonly OFFSETS = {
     [AnimatedDrawerState.Hidden]: 220,
@@ -124,7 +124,7 @@ class AnimatedDrawerController {
 
   constructor(state = AnimatedDrawerState.Docked) {
     this.state = state;
-    this.baseOffset = AnimatedDrawerController.OFFSETS[state];
+    this.baseOffset = AnimatedMapWithDrawerController.OFFSETS[state];
     this.panning = false;
     this.yOffset = new Animated.Value(this.baseOffset);
   }
@@ -132,7 +132,7 @@ class AnimatedDrawerController {
   setState(state: AnimatedDrawerState) {
     this.state = state;
     this.panning = false;
-    this.baseOffset = AnimatedDrawerController.OFFSETS[state];
+    this.baseOffset = AnimatedMapWithDrawerController.OFFSETS[state];
     this.yOffset.flattenOffset();
     Animated.spring(this.yOffset, {toValue: this.baseOffset, useNativeDriver: true}).start();
   }
@@ -149,7 +149,7 @@ class AnimatedDrawerController {
     }
 
     // Are we moving too far in the X direction? If so, treat as a scroll and stop panning the drawer
-    if (Math.abs(gestureState.dx) > AnimatedDrawerController.SNAP_THRESHOLD) {
+    if (Math.abs(gestureState.dx) > AnimatedMapWithDrawerController.SNAP_THRESHOLD) {
       this.onPanResponderRelease();
       return;
     }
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
@floatplane just a draft but would appreciate some code review so far

stevekuznetsov:
@floatplane this works and it's ready for review. I hear you about separate controllers, but:

1. we really do have a tight coupling between the map and the drawer, so I'm not sure that from an ideological perspective it's that clear of a distinction
2. as you know, getting these animation bits and classes to work is really not a straightforward process. I could not find any way to get the class to call back into functional hook-land without a lag, so the `mapView.animateToRegion()` call would always stutter pretty terribly. If you can find a way to make that work in the future, seems like we could quickly go back to having the controller only at the lower level

components/AvalancheForecastZoneMap.tsx
```diff
@@ -197,13 +242,105 @@ class AnimatedDrawerController {
...
+      (this.windowHeight - this.topElementsHeight - cardDrawerHeight - this.tabBarHeight - regionHeightPixels) / 2; // half the leftover space in the screen
+    // finally, we calculate the offset, in pixels, from the center of the region to the center of the screen
+    const regionCenterXOffsetPixels = regionCenterXPixel - this.windowWidth / 2;
+    const regionCenterYOffsetPixels = regionCenterYPixel - this.windowHeight / 2;
+
+    // now, we can position our region relative to our overlays by asking the displayed region to:
+    // - center around the virtual point that results in the correct layout
+    // - contain enough longitude and latitude deltas to match the whole screen
+    const displayedRegion: Region =
+      degreesPerPixelVertically > 0
```
stevekuznetsov, line 324: We need this funky check as this method fires in the first couple of render passes when we still don't know the window dimensions, the size of the tab bar, etc. Open to better approaches here!

components/AvalancheForecastZonePolygon.tsx
```diff
@@ -20,25 +19,19 @@ const toLatLng = (item: number[]): LatLng => {
...
-  const coordinates: LatLng[] = coordinateList(zone.geometry).map(toLatLng);
-  React.useEffect(() => {
-    setRegionBounds((previous: RegionBounds) => updateBoundsToContain(previous, coordinates));
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
-
+export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheForecastZonePolygonProps> = ({zone, setSelectedZone}: AvalancheForecastZonePolygonProps) => {
   return (
     <Polygon
-      coordinates={coordinates}
+      coordinates={toLatLngList(zone.geometry)}
```
stevekuznetsov, line 34: We could pass in the coordinates we calculated at the owner component, but I figured it would be valuable for this component to always be self-consistent in terms of the state it's using - e.g. no real benefit to allowing a mistake to propagate new coordinates but not a new map zone.

components/AvalancheForecastZoneMap.tsx
```diff
@@ -35,39 +45,58 @@ export interface MapProps {
...
   const [selectedZone, setSelectedZone] = useState<MapViewZone | null>(null);
   const onPress = useCallback(() => {
     setSelectedZone(null);
   }, []);
 
+  const regionBounds: RegionBounds = zones
+    ? zones.reduce((accumulator, currentValue) => updateBoundsToContain(accumulator, toLatLngList(currentValue.geometry)), defaultRegionBounds)
+    : defaultRegionBounds;
+  zones?.forEach(zone => updateBoundsToContain(regionBounds, toLatLngList(zone.geometry)));
+  const region: Region = regionFromBounds(regionBounds);
```
floatplane, line 64: `region` and `regionBounds` are kind of generic, is there a more specific name that isn't horrible to type? something like `avalancheCenterRegion` but maybe shorter?

components/AvalancheForecastZoneMap.tsx
```diff
@@ -35,39 +45,58 @@ export interface MapProps {
...
+
+  // useRef has to be used here. Animation and gesture handlers can't use props and state,
+  // and aren't re-evaluated on render. Fun!
+  const mapView = useRef<AnimatedMapView>(null);
+  const controller = useRef<AnimatedMapWithDrawerController>(new AnimatedMapWithDrawerController(AnimatedDrawerState.Hidden, region, mapView)).current;
+  React.useEffect(() => {
+    controller.setRegion(region);
+  }, [region, controller]);
+
+  const {width: windowWidth, height: windowHeight} = useWindowDimensions();
```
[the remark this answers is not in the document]
floatplane: This seems right to me, as opposed to blindly calling `setWindowDimensions` every time through render.

I do wonder if the name `setWindowDimensions` is a little misleading, as the function is doing more than just setting internal state. Maybe `animateToWindowDimensions` would make it clearer that there's more going on here, and why it should be handled in `useEffect`.

floatplane (approved):
looks great! just a couple optional suggestions

PR #102, opened 2023-01-25
First pass at the weather forecast

floatplane:
there's still a few TODOs in here, but I think this is far enough along to check in - works in all zones and doesn't throw runtime errors.

image

### DIFF
--- a/components/forecast/WeatherTab.tsx
+++ b/components/forecast/WeatherTab.tsx
@@ -1,27 +1,35 @@
 import {Card} from 'components/content/Card';
-import {Center, HStack, VStack} from 'components/core';
+import {Center, HStack, View, VStack} from 'components/core';
 import {useNWACWeatherForecast} from 'components/forecast/useNWACWeatherForecast';
-import {AllCapsSm, AllCapsSmBlack, Body, Title3Black} from 'components/text';
-import {HTML} from 'components/text/HTML';
-import {ActivityIndicator} from 'react-native';
+import {AllCapsSm, AllCapsSmBlack, Body, BodyBlack, BodySemibold, Title3Black} from 'components/text';
+import {ActivityIndicator, StyleSheet} from 'react-native';
+import {colorLookup} from 'theme';
 import {AvalancheForecastZone} from 'types/nationalAvalancheCenter';
 
 interface WeatherTabProps {
   zone: AvalancheForecastZone;
 }
 
+const timeOfDayString = (period: 'day' | 'night', subperiod: 'early' | 'late') => {
+  if (period === 'day') {
+    return subperiod === 'early' ? 'morning' : 'afternoon';
+  } else {
+    return subperiod === 'early' ? 'evening' : 'overnight';
+  }
+};
+
 export const WeatherTab: React.FC<WeatherTabProps> = ({zone}) => {
   const {isLoading, isError, data: forecast} = useNWACWeatherForecast();
 
   if (isLoading) {
     return (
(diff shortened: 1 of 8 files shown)


components/forecast/WeatherTab.tsx
```diff
@@ -0,0 +1,114 @@
...
+const timeOfDayString = (period: 'day' | 'night', subperiod: 'early' | 'late') => {
+  if (period === 'day') {
+    return subperiod === 'early' ? 'morning' : 'afternoon';
+  } else {
+    return subperiod === 'early' ? 'evening' : 'overnight';
+  }
+};
+
+export const WeatherTab: React.FC<WeatherTabProps> = ({zone}) => {
+  const {isLoading, isError, data: forecast} = useNWACWeatherForecast();
```
stevekuznetsov, line 22: What does this do for a different center? Did we make an attempt to structure the output of this to be similar to what the NAC provides for weather forecasts?

stevekuznetsov: Also - do we want to model this as one hook without params to get all forecasts for all zones? Why?

components/forecast/WeatherTab.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <VStack space={8} style={{flex: 1}}>
+            <AllCapsSmBlack>Author</AllCapsSmBlack>
+            <AllCapsSm style={{textTransform: 'none'}} color="lightText">
+              {forecast.author || 'Unknown'}
+              {'\n'}
+            </AllCapsSm>
+          </VStack>
+        </HStack>
+        <VStack alignItems="stretch" pt={4}>
+          {forecast.zones[zone.name].slice(0, 2).map((forecast, index) => {
```
stevekuznetsov, line 63: Are we guaranteed that `zone.name` is a key in `forecast.zones`?

components/forecast/WeatherTab.tsx
floatplane: 1) Obviously, it does nothing for a different center (hence the NWAC in the name). I have clearly understood that making things work for other avalanche centers is not supposed to be my top priority for v1 (though I recognize that it is a top priority for you). Also, this is just a placeholder hook. There's a real API coming, at which time we'll bend the UI to match a different shape. 

2) Sure, I could split this a la `useAvalancheForecastFragment(s)`.

stevekuznetsov: And, similarly, since we get to choose the format here green-field, it seems like choosing a format that mimics the NAC would be a clear win and no extra cost as opposed to any other format - WDYT?

components/forecast/WeatherTab.tsx
floatplane: 🤷 as guaranteed as we can be given that this is scraping a web page! but I'll add a null check and display the error

components/forecast/WeatherTab.tsx
floatplane: I think that if you're doing that work for the JSON API, it doesn't make sense to invest here!

Here's some sample NAC forecasts if it's useful:
https://bridgertetonavalanchecenter.org/forecasts/#/weather/117982
https://api.avalanche.org/v2/public/product/117982

components/forecast/WeatherTab.tsx
floatplane: actually I'll pull this in as part of the splitting into two hooks and adding params

PR #107, opened 2023-01-27
More weather cleanups

floatplane:
- `useWeatherForecasts` -> `useLatestWeatherForecasts`
- add a `useLatestWeatherForecast` hook to get weather for a particular zone
- make parser a little more robust when handling table entries like `Saturday Afternoon*`
- make a little more room for precipitation amounts so they don't wrap as easily, but it's still quite a tight squeeze in that space: 
image

### DIFF
--- a/hooks/useLatestWeatherForecasts.ts
+++ b/hooks/useLatestWeatherForecasts.ts
@@ -9,6 +9,8 @@ import {getTimezoneOffset} from 'date-fns-tz';
 import {add, parse} from 'date-fns';
 import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
 
+const LOG_LEVEL = 0;
+
 const precipZoneToForecastZone = {
   'Hurricane Ridge': 'Olympics',
   'Mt Baker Ski Area': 'West Slopes North',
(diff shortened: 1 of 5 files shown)


hooks/useLatestWeatherForecasts.ts
```diff
@@ -40,28 +43,58 @@ interface WindSpeed {
...
+  | 'Sunday Night'
+  | 'Monday Night'
+  | 'Tuesday Night'
+  | 'Wednesday Night'
+  | 'Thursday Night'
+  | 'Friday Night'
+  | 'Saturday Night';
+
+export interface ZoneForecastPeriod {
+  // TODO: does this object need an actual time? Right now, all it has is a relative field like "Sunday Night",
```
stevekuznetsov, line 63: We should draw a line in the sand for the amount of effort put into this particular take on this approach - e.g. instead of attempting to parse "Sunday Night" into a UTC date-time, we can just assert that the JSON API give us a UTC date-time.

floatplane:
Oh 100%, my musing was more for the final API (though I can see that wasn’t
at all in the comment).

PR #119, opened 2023-01-30
map: remove ineffectual code

stevekuznetsov:

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -76,7 +76,6 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
   const avalancheCenterMapRegionBounds: RegionBounds = zones
     ? zones.reduce((accumulator, currentValue) => updateBoundsToContain(accumulator, toLatLngList(currentValue.geometry)), defaultAvalancheCenterMapRegionBounds)
     : defaultAvalancheCenterMapRegionBounds;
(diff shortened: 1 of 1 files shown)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -76,7 +76,6 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
   const avalancheCenterMapRegionBounds: RegionBounds = zones
     ? zones.reduce((accumulator, currentValue) => updateBoundsToContain(accumulator, toLatLngList(currentValue.geometry)), defaultAvalancheCenterMapRegionBounds)
```
floatplane, line 77: using `defaultAvalancheCenterMapRegionBounds` as the accumulator will modify that constant in place, is that the right thing to do?

maybe `defaultAvalancheCenterMapRegionBounds` should instead be a function that returns an empty bounds object every time it's called.

floatplane (changes requested):
ptal @stevekuznetsov

components/AvalancheForecastZoneMap.tsx
floatplane: or do I have this wrong, and `updateBoundsToContain` doesn't modify its argument in place? if that's the case then ignore my feedback!

stevekuznetsov: Yeah it is written specifically not to mutate the input

stevekuznetsov: (and there's a test for that) :)

PR #121, opened 2023-01-31
Autogenerated types for Snowbound API

floatplane:
This PR does the following:

1. Adds scripts to take the Snowbound OpenAPI definition and turn it into type definitions and wrapper classes. The scripts manually insert `token` params into each API path because the package that we're using to generate code (`openapi-typescript-codegen`) doesn't fully understand the OpenAPI spec*. In other words, I don't think Snowbound was actually doing anything wrong.
1. Fixes various bits of infra (eslint, prettier) to not touch the generated code

Everything inside `types/generated` is in fact generated, so if you have comments on those files I will probably not act upon them :) 

--- 

*the part of the spec that the codegen package doesn't seem to pick up on is this:
 ```
        "securitySchemes": {
            "OAuth2PasswordBearerWithCookie": {
                "type": "oauth2",
                "flows": {
                    "password": {
                        "scopes": {},
                        "tokenUrl": "/token"
                    }
                }
            },
            "APIKeyQuery": {
                "type": "apiKey",
                "in": "query",
                "name": "token"
            }
        }
 ```
which I believe should be interpreted as "add a `token` query param as optional to every request method"


scripts/gen-openapi.sh
```diff
@@ -0,0 +1,11 @@
+# Download Snowbounds published API and fix it up locally
```
stevekuznetsov, line 1: Could we please:


```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail
```

scripts/getSnowboundOpenAPISpec
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+npx ts-node 'scripts/getSnowboundOpenAPISpec.ts'
```
stevekuznetsov, line 2: Why is this a separate shell script? Could we just inline this?

scripts/gen-openapi.sh
```diff
@@ -0,0 +1,11 @@
...
+tmpfile=$(mktemp /tmp/openapi.XXXXXX)
```
stevekuznetsov, line 2:
```sh
tmpfile="$(mktemp /tmp/openapi.XXXXXX)"

function cleanup() {
    rm -rf "${tmpfile}"
}

trap EXIT cleanup
```

scripts/gen-openapi.sh
```diff
@@ -0,0 +1,11 @@
...
+./scripts/getSnowboundOpenAPISpec > $tmpfile
+
+# Generate TS type definitions from Snowbound's published API
+yarn openapi \
+    --input $tmpfile \
```
stevekuznetsov, line 7:
```suggestion
    --input "${tmpfile}" \
```

scripts/gen-openapi.sh
stevekuznetsov, line 2:
```suggestion
tmpfile="$(mktemp /tmp/openapi.XXXXXX)"
```

scripts/gen-openapi.sh
```diff
@@ -0,0 +1,11 @@
+# Download Snowbounds published API and fix it up locally
+tmpfile=$(mktemp /tmp/openapi.XXXXXX)
+./scripts/getSnowboundOpenAPISpec > $tmpfile
```
stevekuznetsov, line 3:
```suggestion
npx ts-node 'scripts/getSnowboundOpenAPISpec.ts' > "${tmpfile}"
```

types/generated/snowbound/core/ApiError.ts
```diff
@@ -0,0 +1,24 @@
+/* istanbul ignore file */
```
stevekuznetsov, line 1: super-nit: not sure we _need_ another level of directory nesting for `generated` - are we going to co-locate generated and non-generated types for one API?

scripts/getSnowboundOpenAPISpec
floatplane: 🤷  sure

scripts/gen-openapi.sh
floatplane: are you really worried that `mktemp` will produce a filename with a space in it?

types/generated/snowbound/core/ApiError.ts
floatplane: I don't know, maybe? I think it's useful to have a clear delineation between "things you should edit" and "things you shouldn't".

scripts/gen-openapi.sh
stevekuznetsov: Rule number one of bash - never trust anything :) In this context, you never have to wonder - why is it valid for _this_ variable to be unquoted? Simply not a useful distinction

types/generated/snowbound/core/ApiError.ts
stevekuznetsov: I see. Our GraphQL generated stuff is in the wrong place then - I am used to CI that will re-gen and ensure that things are up-to-date, as well. In order for CI to not trigger diffs, I'd check in the OpenAPI schema (like we did for GraphQL) and validate that the schema -> code generation is correct at merge-time.

PR #122, opened 2023-01-31
Weather station list in weather forecast tab

floatplane:
This PR adds the list of weather stations to the end of the weather forecast. Like the overall weather forecast, this has only been tested with NWAC data and may need more work to function for other avalanche centers.

The flow in `useWeatherStations` is as follows:
- take the input avalanche center
- get the overall bounding box for all zones in the center
- ask Snowbound for all weather stations in the center's bounding box
- take the list of weather stations and break it down by zone, by testing station location against the zone polygon boundary
- apply any additional grouping to the weather stations, if defined (only defined for NWAC)
- return the map of zone => weather station groups

## screenshots

loading state | error state | loaded state
--- | --- | ---
image | image | i

### DIFF
--- a/hooks/useWeatherStations.ts
+++ b/hooks/useWeatherStations.ts
@@ -41,7 +41,7 @@ const stationGroupMapping = {
   'Mount Rainier - Paradise': ['35', '36'],
   'Mount Rainier - Camp Muir': ['34'],
   'Chinook Pass': ['32', '33'],
-  'White Pass': ['37', '39'],
+  'White Pass': ['37', '39', '49'],
   'Mt St Helens': ['40'],
 
   // West Central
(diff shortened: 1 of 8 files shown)


hooks/useWeatherStations.ts
```diff
@@ -0,0 +1,129 @@
...
+  'Timberline Ski Area - Magic Mile Chair': ['45'],
+  'Mt Hood Meadows Ski Area': ['42', '43'],
+  'Mt. Hood Meadows Cascade Express': ['41'],
+};
+
+const decommissionedStations = [
+  '15', // Stevens Pass - Brooks Wind (Retired 2019)
+];
+
+export const useWeatherStations = ({center, sources}: Props) => {
```
stevekuznetsov, line 70: Could you speak to why this is per-center, not per-zone?

floatplane: I'll ultimately need it to be per-center for the weather station list view, I think. This at least gives me that flexibility.

PR #124, opened 2023-02-02
map: update card behavior

stevekuznetsov:
map: consider the card someone scrolled to as selected

---

map: ensure all cards are the same height

---

https://user-images.githubusercontent.com/7328370/216345306-e462430b-3998-4427-9229-e98d2cbfbb24.MP4

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -485,7 +485,7 @@ const AvalancheForecastZoneCard: React.FunctionComponent<{
           dateString: apiDateString(date),
         });
       }}>
-      <VStack borderRadius={8} bg="white" width={width * CARD_WIDTH} mx={CARD_MARGIN * width}>
+      <VStack borderRadius={8} bg="white" width={width * CARD_WIDTH} mx={CARD_MARGIN * width} height={200}>
         <View height={8} width="100%" bg={dangerColor.string()} borderTopLeftRadius={8} borderTopRightRadius={8} pb={0} />
         <VStack px={24} pt={4} pb={12} space={8}>
           <HStack space={8} alignItems="center">
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -11,6 +11,8 @@ import {
   GestureResponderEvent,
   TouchableOpacity,
   LayoutChangeEvent,
+  NativeSyntheticEvent,
+  NativeScrollEvent,
 } from 'react-native';
 import MapView, {Region} from 'react-native-maps';
 import AnimatedMapView from 'react-native-maps';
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
This does not play well with our calls to animate the drawer to a particular index ...

components/AvalancheForecastZoneMap.tsx
```diff
@@ -394,6 +400,26 @@ const AvalancheForecastZoneCards: React.FunctionComponent<{
...
+    // that far in the iteration that the offset prior to that is on the other side of the center
+    // of the screen)
+    let index = 0;
+    const offset = event.nativeEvent.contentOffset.x + width / 2;
+    for (let i = offsets.length - 1; i >= 0; i--) {
+      if (offsets[i] <= offset) {
+        index = i;
+        break;
+      }
+    }
```
floatplane, line 419: Does it work to just look for the first offset that's greater, and then go back one? (

```suggestion
    const cardCenter = event.nativeEvent.contentOffset.x + width / 2;
    const index = Math.max(0, offsets.findIndex(o => o > cardCenter) - 1);
```

components/AvalancheForecastZoneMap.tsx
```diff
@@ -415,6 +441,8 @@ const AvalancheForecastZoneCards: React.FunctionComponent<{
           transform: [controller.getTransform()],
         },
       ]}
+      onScroll={handleScroll}
+      scrollEventThrottle={200}
```
floatplane, line 445: might be worth a quick comment why you need this - is it to prevent the highlight from zooming around the map when you scroll? 

also you might want to look at https://lodash.com/docs/4.17.15#debounce for changing the selection only when panning has settled

floatplane (approved):
seems good! I am curious whether we'll need to tweak this to avoid rapidly changing selection while flipping through cards, and added a suggestion to handle that case, but we can do that as a followup if needed.

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: It just didn't seem necessary to fire this 1k times a second, is all. I actually realized that we have a bigger problem - `onScroll` doesn't distinguish between a user panning the thing and us programatically calling `scrollToIndex()`, so sometimes but not always when you push a zone and we scroll through many others on the way there, this `onScroll` would interrupt that and everything would get weird. Using `onMomentumScrollEnd` makes it look really laggy - it waits until scrolling is entirely finished before firing. Not sure how to proceed.

floatplane:
> This does not play well with our calls to animate the drawer to a particular index ...

I wondered if there would be some kind of circular tussle there around who controlled the selection

stevekuznetsov:
We could bifurcate the state - one `selected` as a downward indication of user state, one upward `visible` as an indication of what is on screen. `selected` is driven by user input and causes the drawer to pan; user input can cause drawer to pan; drawer panning updates `visible` that drives UI.

stevekuznetsov:
Hm that won't really work since we likely don't want to cycle through selecting all the zones on the map while we shuffle through them in the carousel on a `scrollToIndex()` call. Sounds like we need some "future intent" distinction here.

stevekuznetsov:
OK, everyone online just said to track when you're programmatically scrolling with another boolean state variable. I did that - it seems to work just fine. @floatplane ready for a final pass.

PR #125, opened 2023-02-03
Weather data table

floatplane:
This is *not* the finished table! But it's a useful checkpoint along the way. I've added a helpful disclaimer to the top of the UI ;)

What's working:
- rendering data from Snowbound API (last 24 hours only)

What's tbd:
- bunch of TODOs in the code, but off the top of my head
  - top header, navigation, other things that aren't designed yet
  - 24 hour/7 day selector
  - plumb through date from top level
  - cumulative snowfall column - not returned by Snowbound API, looks like it can just be synthesized on the client
  - other issues as they're uncovered

app | website
--- | ---
 | image

### DIFF
--- a/hooks/useWeatherStationTimeseries.ts
+++ b/hooks/useWeatherStationTimeseries.ts
@@ -6,6 +6,7 @@ import {ClientContext, ClientProps} from 'clientContext';
 import {ApiError, OpenAPI, StationMetadata, TimeseriesDataService} from 'types/generated/snowbound';
 import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
 import AvalancheCenterMetadata from 'hooks/useAvalancheCenterMetadata';
+import {toSnowboundStringUTC} from 'utils/date';
 
 type Source = 'nwac' | 'snotel' | 'mesowest';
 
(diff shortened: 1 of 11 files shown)


hooks/useWeatherStationTimeseries.ts
```diff
@@ -0,0 +1,98 @@
...
+
+interface Props {
+  center: AvalancheCenterID;
+  stids: string[];
+  sources: Source[];
+  startDate: Date;
+  endDate: Date;
+}
+
+type Variable =
```
stevekuznetsov, line 21: How is this different from `types/snowbound.ts`?

floatplane: well, it's different because you made that one and I made this one?

floatplane: just kidding, I didn't see the other one

stevekuznetsov: OK, just hoping to save you time here ... that was why I had already implemented the entire API thing and had pages showing data ...

floatplane: Ah. I was trying to lean on the OpenAPI definitions and only fill in gaps around those, which is why I ended up duplicating work for these types 😬 

Are the API shapes identical, or subtly different? I haven't dug into them enough to see.

Issue #67, opened 2023-01-12
We should clean up old PR channels and branches in Expo


floatplane:
I don't think it hurts anything to let these pile up, but it'd be good to keep the dashboard tidy and it's probably more manageable to view in Expo Go as well.

We just need a PR closed action that executes the following commands for each branch (substitute the PR number, of course):

```sh
$ eas channel:delete --non-interactive pr-56
✔ ️Deleted channel "pr-56".

$ eas branch:delete --non-interactive pr-56 
✔ ️Deleted branch "pr-56" and all of its updates on project @nwac/avalanche-forecast.
```

floatplane:
workaround: run this
```sh
eas branch:list --json --non-interactive \
| jq '.[] | del(.updates)| select( .name | startswith("pr-")) | .name' -r \
| xargs -n 1 -I {} sh -c "eas channel:delete --non-interactive {}; eas branch:delete --non-interactive {}"
```

yuliadub:
would it make sense to set the setting for branches to be auto deleted when a PR is merged to main?

floatplane:
are you talking about Git branches? in which case I think we should 😄. but this issue was about EAS branches. continues to be so annoying that they used the same word.

stevekuznetsov:
@yuliadub I updated the GitHub settings to do that this morning actually - and mentioned to @floatplane that we can write a GitHub action to trigger on (`git`) branch deletion to remove the EAS branches

yuliadub:
@floatplane ah yes, I was thinking github - good points on both, thanks for getting this figured out!

PR #129, opened 2023-02-04
New workflow: clean up expo channel when PR closed

floatplane:
Going to close this one after opening to make sure it works correctly :)


floatplane:
yay, fixes #67

floatplane:
Also reverting the change to publish to the `latest` channel from #128, keeping the channel name as `preview` is simpler. Changing PR channel names instead.

PR #134, opened 2023-02-07
forecast: update headers

### LINKED ISSUE #133: Typo on avalanche forecast page
`Avalanche Problems` shouldn't be plural - each header is only talking about one problem. Probably should also include "#1", "#2", etc. the way the website does.

Design is still thinking about whether we're going to change the text to include the problem type in the header, but this issue is just for a quick fix to what's there now.

image

stevekuznetsov:
File (4)

Fixes #133


floatplane (approved):
thanks!

PR #145, opened 2023-02-15
Organize imports

floatplane:
This is a pair of changes that I have wanted for a while now, but not sure if anyone else feels as passionately as I do about it:

- apply consistent sorting to imports
- enforce non-relative imports

Prettier does the sorting, so it should just work. ESLint enforces import style, and can autofix them too. Also, VSCode will automatically add imports with absolute paths if you change this setting:

image

I've had that setting enabled for weeks now - it just works.

No big rush on this - we can discuss when you're back @stevekuznetsov


stevekuznetsov:
@floatplane I am 100% for this change as long as it's easy to get that outside of VSCode - here's my JetBrains config:
image (2)

floatplane:
@stevekuznetsov I would guess you'd need to toggle on that second checkbox (`Use paths relative to the project...`). And maybe turn off `Sort imported members` since prettier would be taking over that job. LMK if that works! I'll rebase this branch on something newer in the meantime.

PR #147, opened 2023-02-16
Fix some Android issues

floatplane:
1. Crash on the weather tab - short-term fix is to not call `getTimezoneOffset`. Longer-term fix is to switch to a weather forecast API and remove the web scraper. Kind of worrying that other functions in `date-fns-tz` might have problems though (https://github.com/marnusw/date-fns-tz#usage-with-android). We should be careful about taking new dependencies on that library.

2. For submit obs, handle the Android back button. React navigation normally handles this, but we're not using react nav for that screen at the moment.


floatplane:
and one more commit - TextInput + padding don't seem to work correctly on Android, so add the border and padding to a View that wraps around the TextInput instead

PR #154, opened 2023-02-19
persist data to async local storage

### LINKED ISSUE #1: Persist Data in Async Storage
Recording here for the future, we need to be using the [`PersistQueryClientProvider`](https://tanstack.com/query/v4/docs/plugins/persistQueryClient#persistqueryclientprovider) and the [`AsyncStoragePersister`](https://tanstack.com/query/v4/docs/plugins/createAsyncStoragePersister) and we should transparently get caching in async storage. Need to check the footprint limits here, since there are two (?) mechanisms for storage in the App world and one of them has a small enough cap that we wouldn't fit.

stevekuznetsov:
*: upgrde to newest react-query

---

App.tsx: use an asyc storage persister for queries

---

network: prefetch problem icons

---

network: dedupe center icon fetches

---

network: prefetch thumbnails and originals for images

---

weather: render the loading and error states within the tab

---

Fixes #1 
/assign @floatplane


stevekuznetsov:
Two remaining issues:

- for whatever reason our images served from local storage don't load when in airplane mode / offline
- even if you wait for all the image prefetching to finish, it's not 100% that the media carousels work ... why?

stevekuznetsov:
OK! This is ready :)

App.tsx
```diff
@@ -70,7 +74,20 @@ if (Sentry?.init) {
...
 }
 
-const queryClient: QueryClient = new QueryClient();
+const queryCache: QueryCache = new QueryCache();
+
+const queryClient: QueryClient = new QueryClient({
+  queryCache: queryCache,
+  defaultOptions: {
+    queries: {
+      cacheTime: 1000 * 60 * 60 * 24, // 24 hours
```
floatplane, line 83: Why cache in memory for 24 hours when we are *also* caching to storage?

How much ram do we use keeping all these query results loaded?

components/AvalancheCenterLogo.tsx
```diff
@@ -34,57 +31,62 @@ export const AvalancheCenterLogo: React.FunctionComponent<AvalancheCenterLogoPro
...
 
+  const {data: uri} = useCachedImageURI(source[avalancheCenterId].uri);
+  if (!uri) {
+    return <ActivityIndicator style={style} />;
+  }
+
   const images: Record<string, {(s: ImageStyle): ReactElement}> = {
     ['BTAC']: (s: ImageStyle) => {
-      return <Image style={s} source={require('../assets/logos/BTAC.png')} />;
+      return <Image style={s} source={{uri: uri, cache: 'force-cache'}} />;
```
floatplane, line 41: I am still super confused about this. If we're requiring something from the app bundle, why do we need to cache it? It's already in storage!

components/AvalancheCenterLogo.tsx
```diff
@@ -94,48 +96,48 @@ export const AvalancheCenterLogo: React.FunctionComponent<AvalancheCenterLogoPro
...
   return images[avalancheCenterId](actualStyle);
 };
 
-export const preloadAvalancheCenterLogo = async (avalancheCenter: AvalancheCenterID) => {
+export const preloadAvalancheCenterLogo = async (queryClient: QueryClient, avalancheCenter: AvalancheCenterID) => {
   /* eslint-disable @typescript-eslint/no-var-requires */
   switch (avalancheCenter) {
     case 'BTAC':
-      return Image.prefetch(Image.resolveAssetSource(require('../assets/logos/BTAC.png')).uri);
+      return ImageCache.prefetch(queryClient, Image.resolveAssetSource(require('../assets/logos/BTAC.png')).uri);
```
floatplane, line 107: Similar question here - what's the point of pre-fetching a thing that's already in the app bundle? (I feel like I've asked this in Slack and not really understood the answer, sorry that I keep repeating myself)

components/AvalancheCenterLogo.tsx
```diff
@@ -94,48 +96,48 @@ export const AvalancheCenterLogo: React.FunctionComponent<AvalancheCenterLogoPro
...
     case 'ESAC':
-      return Image.prefetch(Image.resolveAssetSource(require('../assets/logos/ESAC.png')).uri);
+      return ImageCache.prefetch(queryClient, Image.resolveAssetSource(require('../assets/logos/ESAC.png')).uri);
     case 'WAC':
-      return Image.prefetch(Image.resolveAssetSource(require('../assets/logos/WAC.png')).uri);
+      return ImageCache.prefetch(queryClient, Image.resolveAssetSource(require('../assets/logos/WAC.png')).uri);
   }
   const invalid: never = avalancheCenter;
-  throw new Error(`Unknown avalanche center: ${invalid}`);
+  throw new Error(`Unknown avalanche center: ${JSON.stringify(invalid)}`);
```
floatplane, line 142: isn't `invalid` just a string?

hooks/useCachedImageURI.ts
```diff
@@ -0,0 +1,69 @@
...
+function queryKey(uri: string) {
+  return ['image', {uri: uri}];
+}
+
+const fetchCachedImageURI = async (uri: string) => {
+  if (!uri.startsWith('http')) {
+    // this is already a local file, no need to download
+    return uri;
+  }
+  const destination = FileSystem.cacheDirectory + '/' + md5(uri);
```
floatplane, line 24: suggestion: add a date component to the path (`2022-02-21`) and then at startup you can prune stale directories efficiently, in case things get missed by the cleanup routine

hooks/useCachedImageURI.ts
```diff
@@ -0,0 +1,69 @@
...
+      Log.prefetch(`prefetching image ${uri}`);
+      const result = await fetchCachedImageURI(uri);
+      Log.prefetch(`finished prefetching image ${uri} to ${result}`);
+      return result;
+    },
+  });
+};
+
+// we clean up data in our filesystem cache when react-query removes our reference
+const cleanupCachedImage = async event => {
```
floatplane, line 45: this seems necessary but maybe not sufficient. what happens if the user kills the app, does this routine get a chance to run?

hooks/useCachedImageURI.ts
floatplane: also, what happens to `destination` here? how does a later request for `uri` get turned into a filesystem read of `destination`? I think there's some magic I'm missing

components/forecast/WeatherTab.tsx
```diff
@@ -52,14 +55,19 @@ export const WeatherTab: React.FC<WeatherTabProps> = ({zone, center_id, date}) =
...
+        </VStack>
       </Center>
     );
   } else if (isError) {
     return (
-      <Center style={StyleSheet.absoluteFillObject}>
-        <Body>Could not fetch weather forecast for zone {zone.name}</Body>
+      <Center style={{flex: 1}}>
+        <Body>
+          Could not fetch weather forecast for the {zone.name} zone: {forecastError?.message}
```
floatplane, line 69: Is the error here human-readable, or is it programmer-readable? (My suspicion would be it's the latter)

More information is not always helpful, especially when the user can't meaningfully act on it. It's often more reassuring to say something like "it's our fault, please try again".

We should probably get a coherent view from design on these topics.

components/forecast/WeatherTab.tsx
```diff
@@ -52,14 +55,19 @@ export const WeatherTab: React.FC<WeatherTabProps> = ({zone, center_id, date}) =
 
   if (isLoading) {
     return (
-      <Center style={StyleSheet.absoluteFillObject}>
-        <ActivityIndicator />
+      <Center style={{flex: 1}}>
+        <VStack space={8} style={{flex: 1}} alignItems={'center'}>
+          <Body>Loading current weather forecast for the {zone.name} zone...</Body>
```
floatplane, line 60: does this text really add anything that the activity indicator doesn't tell you? good question for design

components/content/carousel/NetworkImage.tsx
```diff
@@ -43,24 +49,16 @@ export const NetworkImage: React.FC<NetworkImageProps> = ({uri, width, height, o
...
-            onLoad={() => {
-              setState('ready');
-              onStateChange('ready');
-            }}
-            onError={_e => setState('error')}
-            {...croppedThumbnailProps}
-          />
+      {(status === 'success' || status === 'loading') && (
+        <TouchableOpacity activeOpacity={0.8} onPress={() => onPress(index)} disabled={status !== 'success'}>
+          <Image source={{uri: cachedUri, cache: 'force-cache'}} {...croppedThumbnailProps} />
```
floatplane, line 61: `force-cache` is iOS only. it's also reading from the HTTP cache, not from your local filesystem cache that you wrote to in useCachedImageURI. I think this is probably working when you're testing, but maybe not as you intend?

App.tsx
floatplane: also, I like the idea of setting a global policy but then it'd be good to remove all the explicit settings on each query.

App.tsx
```diff
@@ -88,14 +105,19 @@ const defaultDate = new Date();
 
 const App = () => {
   try {
+    // we need to subscribe to the react-query cache in order to remove
+    // images from the local filesystem when their TTL expires
+    React.useEffect(() => {
+      return queryCache.subscribe(event => ImageCache.cleanup(event));
+    }, []);
```
floatplane, line 112: whoa this is a spooky level of indirection. I only realized this connection existed on my second pass through the PR. 

suggestion: make a method in the ImageCache called `initialize` that contains this code, and pass the `queryCache` into it. 
```suggestion
    React.useEffect(() => {
      return ImageCache.initialize(queryCache);
    }, []);
```

components/AvalancheCenterLogo.tsx
floatplane: also this is referring to the http cache which I assume would not be in play for bundle assets?

floatplane (commented):
I have lots of questions! I want to make sure I understand what's going on here

App.tsx
stevekuznetsov: React Query will hydrate and de-hydrate from the async persister for it's storage, so the footprint will be roughly double, but it's *only* the textual content that comes back from API calls. The asset caching is *not* in this cache. Async storage itself is limited to something like 5MB, and it's stored in some weird plethora of places based on the [platform](https://react-native-async-storage.github.io/async-storage/docs/advanced/where_data_stored),

components/AvalancheCenterLogo.tsx
stevekuznetsov: I addressed this more in the commit message, but the long-and-short of it is that in developer mode, Expo will serve assets via the local network even if those assets would otherwise be bundled into the binary in production. So, this is "just" to support local development, but you'll see the hook no-ops for `file://` URIs in the first place and I think it's very nice to have this working locally to make sure we actually work :)

components/AvalancheCenterLogo.tsx
stevekuznetsov: Same answer :)

components/AvalancheCenterLogo.tsx
stevekuznetsov: I don't know! I saw `[object Object]` :)

hooks/useCachedImageURI.ts
stevekuznetsov: So in general, I do not want to be in the business of cache management. What we do is that we store the _pointer_ in the `react-query` store, e.g.:

```
['image', {uri: 'https://...'}] => 'file://'
```

This is stored in memory and dehydrated into async storage. The normal cache TTL rules apply for the lifecycle of this _pointer_ -- we operate via side-effects in the hook and clean-up function: the actual query function side-effect places the file on disk before returning `destination` (which is the same as `result.uri`), and the cleanup function side-effect removes the file when the pointer is removed. Does that make sense?

hooks/useCachedImageURI.ts
stevekuznetsov: See https://github.com/stevekuznetsov/avalanche-forecast/pull/154#discussion_r1113303686

components/forecast/WeatherTab.tsx
stevekuznetsov: Yeah, I would love to have some more clarity here - this was pretty necessary for me to iterate locally, but I totally agree with you that the error is useless for users. Suiting both cases should be possible.

stevekuznetsov: (hope it's ok if we leave it as-is for now).

(also, I really, really dislike watching e.g. my girlfriend struggle with Apple's very opaque error modals for printer setup, when my Linux box showed me the exact error and I was able to debug it in seconds)

components/content/carousel/NetworkImage.tsx
stevekuznetsov: Oops, this is a hold-over from a previous attempt, we can remove these.

App.tsx
stevekuznetsov: Oops, this is before I realized that variables in the global scope outside of the app can't change, so we don't need `useEffect` here - let me hoist this up to the top of the file

components/AvalancheCenterLogo.tsx
floatplane: that...should be impossible? well let me know what you see with this change!

hooks/useCachedImageURI.ts
stevekuznetsov: Oh, one other super critical bit - we are storing filesystem things in the cache directory, which is a location that the host OS can clean up whenever it wants, if the phone is low on storage, so we can't negatively impact the system here.

To that end, removing a file from disk (even based on dates) if `react-query` does not think it is removed, would lead to us returning the pointer to an empty file.

However, now that I think about it, we could have the same issue with the system removing parts of our cache under filesystem pressure, so I think a complete solution here needs to reconcile the filesystem state with the `react-query` state. Cache coherence is hard.

stevekuznetsov: It doesn't look like it's possible to watch the filesystem, so the best we can do is a periodic poll to reconcile the two caches.

components/AvalancheCenterLogo.tsx
floatplane: hmm ok. I would add a comment to that effect in the code because it's non-obvious and the code looks strange to someone coming in cold.

I would also try to skip calling `ImageCache.prefetch` unless we're in `__DEV__` mode since it just wastes cache space.

stevekuznetsov: > I would also try to skip calling ImageCache.prefetch unless we're in __DEV__ mode since it just wastes cache space.

If you call these with a `file://` URL, they no-op.

hooks/useCachedImageURI.ts
stevekuznetsov: No, it's not guaranteed, but because we can't harm the system, I simply did not care -  more logic for no real benefit there :)

I've written a simple reconciler that we can run in the background anyway for the case where the system removed something, so we should be good here.

> If you're only deleting old files at startup (which is what I suggested) then I don't see how it could impact react-query.

`react-query` stores the cache state in async (persistent) storage, so when the app starts after a) being killed or b) crashing, it re-hydrates the cache, and any file links that were persisted but had their files removed would be invalid.

components/AvalancheCenterLogo.tsx
stevekuznetsov: Added a comment to the hook.

floatplane (approved):
:shipit:

PR #157, opened 2023-02-21
Pulse map region color when an avalanche warning is in effect

floatplane:
Finally got a chance to test this for real today and make sure it works as intended 😄 

https://user-images.githubusercontent.com/101196/220224246-9f8dd7e4-e2ba-46d2-90ba-26da8497ff74.mp4


floatplane:
I don't know what's up with these videos...maybe check out the preview instead

stevekuznetsov:
Looks good - I have a WIP branch where we actually fetch warning objects for rendering them in the forecast, will wire that into your `hasWarning` boolean.

PR #158, opened 2023-02-21
map: add a search bar

stevekuznetsov:
@floatplane currently stuck on how to get this to work in Expo Go for us - our API keys are restricted for use only by our "real" app - and this library (the only one ...) chooses not to use the normal SDKs, so whatever magic Expo Go does to make the map work, is not applied here by default. Will need to figure that out, or get a dev token ...

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -25,6 +25,7 @@ import {defaultMapRegionForZones, ZoneMap} from 'components/content/ZoneMap';
 import {Center, HStack, View, VStack} from 'components/core';
 import {DangerScale} from 'components/DangerScale';
 import {TravelAdvice} from 'components/helpers/travelAdvice';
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
We need to figure out what the UX of this thing is. As is, the library expects to render everything in one view, so doing an overlay of the results outside of the header thing will be impossible, I think.

floatplane:
> @floatplane currently stuck on how to get this to work in Expo Go for us - our API keys are restricted for use only by our "real" app - and this library (the only one ...) chooses not to use the normal SDKs, so whatever magic Expo Go does to make the map work, is not applied here by default. Will need to figure that out, or get a dev token ...

that's annoying! I wonder if we could create a separate set keys under our own account that work correctly with Expo Go, by following the instructions in https://docs.expo.dev/versions/latest/sdk/map-view/ and using the Expo Go bundle id instead of our own. I presume that mismatch is what keeps our existing keys from working correctly with this component.

we could also ask for help in Expo support channels?

PR #166, opened 2023-02-23
workflows: Enable ability to submit ios apps to app store connect via github actions

yuliadub:
added the following: 
- needed configuration in eas.json
- added info.plist to app.json to ensure apple will accept submission (when we request new app capabilities like location, more will be added to this)
- edited build command to submit the build if successful

until we lock in the details on when we auto submit builds to ios/android, both of them will be disabled, since currently we do not do anything with the Android build we do create (keep me honest though) and we have a limit on how many builds we get in expo free tier per month. Once we nail down the details for how often we want to push to TestFlight for both, as well as what branch we should do it from, we can re-enable.

we will need to add secrets to github actions (doesnt look like I have permissions, @stevekuznetsov  will need your help). Will need to add these 4 values, I will provide the team and app id in slack, for account to use to submit, we should chat (I assume our developer one?) 

- IOS_USER_ID
- IOS_TEAM_ID
- IOS_APP_ID
- EXPO_APPLE_PASSWORD


.github/workflows/publish.yaml
```diff
@@ -59,5 +61,14 @@ jobs:
           install_expo: true
           expo_token: ${{ secrets.EXPO_TOKEN }}
 
+      - name: Replace IOS USER value in eas.json
+        run: sed -i 's/$IOS_USER_ID/${{ secrets.IOS_USER_ID }}/g' eas.json
```
yuliadub, line 65: there is likely a better way to do this? let me know and I can change, I just went with the path of least resistance :D

stevekuznetsov: What we've done in the past is to add a placeholder into the `app.json`:

https://github.com/stevekuznetsov/avalanche-forecast/blob/a7ead5c25b18b2017371657fd7ece8a83744cda2/app.json#L28

Then overwrite it at load-time from the environment using TypeScript:
https://github.com/stevekuznetsov/avalanche-forecast/blob/a7ead5c25b18b2017371657fd7ece8a83744cda2/app.config.ts#L11

Let's check if we can also do this with `eas.json` - I would assume so!

yuliadub: ok cool will test this today! I like this much better :)

yuliadub: @stevekuznetsov something I realized, the eas.json is simply used for the eas submission, has nothing to do with the app, so loading values from env during app build doesnt help us here. Eas need to be populated in the github action so it can then use that info when it goes to submit the app. 

I am torn with just hardcoding those values in eas, what are your thoughts on that? are those values actually secret? I cant seem to find any info one way or another.

.github/workflows/publish.yaml
```diff
@@ -13,6 +14,7 @@ jobs:
     uses: ./.github/workflows/ci.yaml # use the callable ci job to run CI checks
 
   publish-update-bundle:
+    if: github.head_ref == 'preview'
```
yuliadub, line 17: thoughts?: only create expo view for preview, since anything that would be released would go through preview/test flight first anyways, no need to have redundant expo views.

.github/workflows/publish.yaml
```diff
@@ -26,7 +28,7 @@ jobs:
           message: '${{ github.event.head_commit.message }}'
           expo_token: ${{ secrets.EXPO_TOKEN }}
 
-  build-android:
+  build-and-submit-android:
```
yuliadub, line 31: there is a way to trigger builds and submits for both platforms, is that something we'd like to do? that sure would make it easy to keep the build numbers in sync, but comes back to the conversation of do we want to have the two apps be in sync when it comes to store submissions? 

https://docs.expo.dev/build/building-on-ci/

.github/workflows/publish.yaml
yuliadub: ok hilarious, not sure if this is more expo magic but those values do not even have to be populated at all it seems - I created a build that had placeholders for them and the wrong apple password and the version of the app still submitted. so weird!

.github/workflows/publish.yaml
stevekuznetsov: Isn't the `bundle` here the JavaScript bundle we're shipping OTA for everyone? Seems to me like we should always build it, but based on the branch we're building from, we put it into different branches.

.github/workflows/publish.yaml
stevekuznetsov: As long as we have a trivially easy way to derive the repo state from some combination of runtime version, build version and update UUID, I don't know that having the same literal build number is that important. The more important part would be making sure that Android and iOS are running the same code.

yuliadub: agreed, and we should be able to do that easily. 

do you think it would be nice to have 2 separate github action jobs to submit the apps per platform? or merging it into one would be cleaner?

.github/workflows/publish.yaml
yuliadub: in theory, yes, and what I am saying for now, until we understand how the bundle works better, lets not update it in prod and only push updates via builds/new versions. 

going to read about bundles today.

.github/workflows/publish.yaml
stevekuznetsov: One for each and a third to do both could give us the best of all worlds

.github/workflows/publish.yaml
stevekuznetsov: Ah, gotcha - if we do that, we will have a day- to days-long lag between new builds we publish and users getting them. If we are diligent about pushing bundles to the right branch (so that e.g. folks on TestFlight only get things we intend for them to) I think we can reduce app store review headache by using them.

eas.json
```diff
@@ -14,11 +14,21 @@
...
+      "autoIncrement": true
     }
   },
   "submit": {
-    "production": {},
+    "production": {
+      "ios": {
+        "appleId": "$IOS_USER_ID",
+        "appleTeamId": "$IOS_TEAM_ID",
+        "ascAppId": "$IOS_APP_ID"
```
floatplane, line 26: does this actually work, i.e. EAS interprets this as a request to read these values from the environment?

.github/workflows/publish.yaml
floatplane: wouldn't you want to build bundles for `main` as well? I'm thinking that someone like a product owner (Charlotte) should be able to open Expo Go and see work that's been recently completed on main, and not have to wait until the point of prepping a new preview branch.

.github/workflows/publish.yaml
floatplane: With OTA updates, the installed app version (as seen by the app stores) is probably less significant than the bundle version / git SHA that the user is running. 

I might be running an ancient app binary, but as long as Expo keeps downloading new JS through OTA updates, I'm staying fresh.

.github/workflows/publish.yaml
yuliadub: yes, great point Brian! I think between the channel for main being called preview today I got confused :D

eas.json
yuliadub: no, the environment stuff doesnt work sine this config is specifically for the eas command, not part of the app. So my initial approach is how we will need to do it! As per your suggestion putting that into a script vs commands in the action

eas_build_and_submit.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sed -i 's/$IOS_USER_ID/${{ secrets.IOS_USER_ID }}/g' eas.json
```
stevekuznetsov, line 3:
```suggestion
sed -i 's/$IOS_USER_ID/${IOS_USER_ID}/g' eas.json
```

shell, not GitHub action ;)

.github/workflows/publish.yaml
```diff
@@ -22,11 +24,12 @@ jobs:
       - name: Publish EAS update
         uses: ./.github/actions/expo/update
         with:
-          channel: 'preview'
+          channel: '{{ github.head_ref }}'
           message: '${{ github.event.head_commit.message }}'
           expo_token: ${{ secrets.EXPO_TOKEN }}
 
-  build-android:
+  build-and-submit-android-and-ios:
```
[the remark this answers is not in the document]
yuliadub: I just screwed it up :D

app.json
```diff
@@ -26,11 +26,13 @@
       "bundleIdentifier": "us.nwac.forecast",
       "config": {
         "googleMapsApiKey": "LOADED_FROM_ENVIRONMENT"
+      },
+      "infoPlist": {
+        "NSPhotoLibraryUsageDescription": "The app is requesting access to the photo library to enable submission of avalanche observations that can include photos and/or video to provide context of an avalanche that was observed."
```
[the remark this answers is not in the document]
yuliadub: this is required for the submission to work. we can change the text to be whatever but the blurb needs to be here for apple to like us

eas_build_and_submit.sh
yuliadub: haha good call!

.github/workflows/publish.yaml
```diff
@@ -22,30 +29,12 @@ jobs:
       - name: Publish EAS update
         uses: ./.github/actions/expo/update
         with:
-          channel: 'preview'
+          channel: '${GITHUB_REF##*/}'
```
stevekuznetsov, line 32: Why do we need to do shell string manipulations here? Isn't the ref one of (`main`, `release`, `preview`)?

eas_build_and_submit.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sed -i 's/IOS_USER_ID/${IOS_USER_ID}/g' eas.json
```
stevekuznetsov, line 3: As written this will replace the value with the literal string `${IOS_USER_ID}`

stevekuznetsov, line 3:
```suggestion
sed -i "s/IOS_USER_ID/${IOS_USER_ID}/g" eas.json
```

app.json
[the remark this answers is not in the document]
stevekuznetsov: I was moreso surprised it was required, since the doc on the config plugins says that they will handle this for us.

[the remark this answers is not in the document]
floatplane: AFAIK it is required, each app has to say why they specifically want to browse photo libraries.

I think the part that's handled automagically is getting this value out of this JSON file and into the Info.plist before building the iOS binary.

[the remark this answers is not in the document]
yuliadub: what Brian said. Apple doing Apple things :D

.github/workflows/publish.yaml
yuliadub: so its interesting, the ${{ github.head_ref }} works in the if statement in the yaml but appears blank when used in a command. The GITHUB_REF has the whole path of the branch (i.e. refs/heads/release), hence the manipulation

app.json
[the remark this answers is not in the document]
yuliadub: basically the app submits just fine but then apple rejects it unless this is in here 
image

PR #167, opened 2023-02-23
*: close over query state in hooks

stevekuznetsov:
We have a couple of similar issues coming to a head here:

 - you can't nest hooks inside of hooks, so you can't cleanly close over a hook that needs some ancillary data for its inner workings
 - when a screen needs data from a number of sources, say, when one of the hooks being used has further data requirements, the handling logic for any of those hooks still loading or having an error gets annoying, quick

In this commit I move the handling logic to a central component, which makes it much less onerous for a hook to simply ask for the data it needs as input.

### DIFF
--- a/components/AvalancheCenterSelector.tsx
+++ b/components/AvalancheCenterSelector.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 
 import {useNavigation} from '@react-navigation/native';
-import {ActivityIndicator, SectionList, SectionListData, StyleSheet, Text, TouchableOpacity, useWindowDimensions} from 'react-native';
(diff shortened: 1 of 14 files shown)


stevekuznetsov:
Prerequisite for fixing https://github.com/stevekuznetsov/avalanche-forecast/issues/120

Addresses https://github.com/stevekuznetsov/avalanche-forecast/issues/120#issuecomment-1441931548

stevekuznetsov:
When we can't load something super basic like the map layer at the top level:

IMG_0046

When we can't load the forecast, but can load the center data:
IMG_0047

When we are loading the weather:
IMG_0048

When we can't load the weather:
IMG_0050

stevekuznetsov:
I know the styling is relatively bad, would love to hold off on that until we get something from design. Note that we eat errors and send them off to Sentry.

PR #168, opened 2023-02-23
Place DangerScale wrapper accurately, push it down with margin

floatplane:
Works in Simulator, we should check it on a couple of devices before merging


stevekuznetsov:
@floatplane couldn't get this to work in preview, tried it locally in Expo Go on both iOS and Android, works

PR #170, opened 2023-02-24
observations: load observations from nwac.us

stevekuznetsov:


App.tsx
```diff
@@ -52,20 +52,43 @@ import {toISOStringUTC} from 'utils/date';
...
+        {
+          status: response.status,
+          data: response.data,
+        },
+        null,
+        2,
+      ),
+    );
+    return response;
+  });
```
floatplane, line 91: is this super noisy?

clientContext.ts
```diff
@@ -3,17 +3,20 @@ import React, {Context} from 'react';
 export interface ClientProps {
   nationalAvalancheCenterHost: string;
   snowboundHost: string;
+  nwacHost: string;
```
floatplane, line 6: very minor: I'm using `observationsUrlPrefix` instead of `Host` for the obs path, since these technically aren't hosts.

types/nationalAvalancheCenter/observations.ts
```diff
@@ -242,3 +246,127 @@ export type PartnerType = (typeof PartnerType)[keyof typeof PartnerType];
...
+    avalanches_caught: z.boolean(),
+    cracking: z.boolean(),
+    cracking_description: z.nativeEnum(InstabilityDistribution).or(z.string().length(0)).nullable(),
+    collapsing: z.boolean(),
+    collapsing_description: z.nativeEnum(InstabilityDistribution).or(z.string().length(0)).nullable(),
+  }),
+  instability_summary: z.string(),
+  observation_summary: z.string(),
+  media: z.array(mediaItemSchema).nullable(),
+  // urls: z.array(z.string()).nullable(), // what even is this?
```
floatplane, line 281: I think this corresponds to uploading video URLs from YouTube/Insta/etc

floatplane (approved):
awesome! this will help me so much

App.tsx
stevekuznetsov: Yes, but when something's goofy, you need it - should we have a separate toggle for inbound & outbound requests?

clientContext.ts
stevekuznetsov: `nwac.us` is a host, what do you mean re: prefixing?

types/nationalAvalancheCenter/observations.ts
stevekuznetsov: Ah, crap ... would have thought those were in media but not image types ... sigh

stevekuznetsov:
@floatplane merge whenever - shouldn't break anything, but NWAC obs is janky

components/observations/ObservationDetailView.tsx
```diff
@@ -20,42 +22,74 @@ import {
...
-  const mapResult = useMapLayer(observation?.getSingleObservation.centerId?.toUpperCase() as AvalancheCenterID);
+  const mapResult = useMapLayer(observation?.getSingleObservation.center_id?.toUpperCase() as AvalancheCenterID);
   const mapLayer = mapResult.data;
 
   if (incompleteQueryState(observationResult, mapResult)) {
     return <QueryState results={[observationResult, mapResult]} />;
   }
 
-  return <ObservationCard observation={observation.getSingleObservation} mapLayer={mapLayer} />;
+  const parseResult = observationSchema.deepPartial().safeParse(observation.getSingleObservation);
```
stevekuznetsov, line 76: @floatplane not sure if you saw `.deepPartial()` - causing all the possible `undefined`, but necessary because of how these APIs are ...

clientContext.ts
floatplane: I mean that `https://nwac.us` is not a host, it's a url prefix 😄

App.tsx
floatplane: yeah, feels like `log requests/log responses` should be two different settings

clientContext.ts
stevekuznetsov: A level of hair-splitting I don't really have strong opinions over :upside_down_face:

PR #178, opened 2023-02-25
Warnings, historical views, etc

stevekuznetsov:
config: allow a more fine-grained logging of requests

We now support:

 - $LOG_NETWORK=all,requests,responses,response-bodies,queries
   This will log only the bits that you ask for, any comma-delimited
   list is fine here.
 - $LOG_NETWORK_MATCHING=''
   This will use a simple .includes() check on the fully-formatted URL
   (with parameters) to filter the requests and responses that are
   logged.

---

map: update the warning animation

This thing is pretty jittery still ... we might want to work on this.

---

form/LocationField: fetch the correct center's map

We need to show the correct map view for the correct center for which
we're currntly submitting an observation. Also, we don't need map zones
and all the extra fetching that comes with that, just the map layer.

---

screens: make a semantic distinction between a date and latest

We need to do different behavior in the API layer depending on whether
we're looking for a particular time in the past for our data or the
latest data, regardless of what time it is.

---

logging-squash

---

latest-squash

---

hooks: log the query paramters we're sending

---

map: hoist queries to the top for zones

We have a couple of different places we get data from for map zones -
the map layer query and all the individual queries for the forecasts and
warnings we do for each zone. Not only was the previous query silently
failing to do anything with the updated data (oops, setting fields that
did not exist since we were not very specific with our types), but it
also had the isseus with composed queries that we know about.

Hoisting the values up to the top allows us to not chain queries,
thereby making our error and loading states clearer, and it allows us to
more explicitly tesselate data from all of our sources.

---

android: ask for filesystem permissions

---

squash-latest

---

forecast: fetch and show warnings

---

network: misc updates to prefetch

Don't need to block on the metadata, need to prefetch danger icons.

---

*: fix rebase conflicts

---

weather: parse Pacific Time strings

---


hooks/useAvalancheForecast.ts
```diff
@@ -1,57 +1,130 @@
...
   });
 };
 
-function queryKey(nationalAvalancheCenterHost: string, forecastId: number) {
-  return ['avalanche-forecast', {host: nationalAvalancheCenterHost, forecast: forecastId}];
+// the query we make to the NAC API does not make use of the nominal date for which we're asking, but we want a good
+// user experience around the cut-over moment between nominal dates. Consider a user that has loaded this query before
+// the cut-over, and has a cached version of the previous day's forecasts. When they refresh their view after the
+// cut-over, we want to render only once we've done a fresh load, since we expect to have new data, instead of using
+// the cache. By adding the nominal date to the query key, we get this behavior.
```
floatplane, line 37: > we want to render only once we've done a fresh load

I disagree pretty strongly with this statement. I would much rather see new data pop in on that rare occasion (we could make the transition smoother if it happens), in exchange for a much faster browsing experience. 

Also I think this behaves poorly if I'm offline just after midnight - there might be data in cache that's not very stale but I can't see it because it was cached as "yesterday's news". Whereas if it was cached under the "latest" key then I'd be able to see it, even if it was slightly stale.

hooks/useMapLayerForecasts.ts
```diff
@@ -0,0 +1,43 @@
...
+            queryFn: async () => AvalancheForecastQuery.fetch(queryClient, nationalAvalancheCenterHost, center_id, feature.id, requestedTime, expiryTimeZone, expiryTimeHours),
+            enabled: !!expiryTimeHours,
+            cacheTime: 24 * 60 * 60 * 1000, // hold on to this cached data for a day (in milliseconds)
+          };
+        })
+      : [],
+  });
+};
+
+export const useMapLayerAvalancheWarnings = (center_id: AvalancheCenterID, requestedTime: RequestedTime, mapLayer: MapLayer) => {
```
floatplane, line 29: minor: move this to its own file?

components/observations/SimpleForm.tsx
```diff
@@ -246,11 +246,7 @@ export const SimpleForm: React.FC<{
                           multiline: true,
                         }}
                       />
-<<<<<<< HEAD
-                      <LocationField name="location_point" label="Latitude/Longitude" />
-=======
                       <LocationField name="mapLocation" label="Latitude/Longitude" center={center_id} />
```
floatplane, line 249: this is wrong, should be `location_point`

floatplane (commented):
I have Feelings™ about trying to turn `latest` into a Date object

components/form/LocationField.tsx
```diff
@@ -1,30 +1,32 @@
...
   const {
     field: {onChange, value},
     fieldState: {error},
   } = useController({name});
   const [modalVisible, setModalVisible] = useState(false);
 
-  const zonesResult = useMapViewZones('NWAC', new Date());
-  const zones = zonesResult.data;
+  const mapLayerResult = useMapLayer(center);
+  const mapLayer = mapLayerResult.data;
```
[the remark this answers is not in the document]
stevekuznetsov: Using the `useMapLayer()` hook just declares our dependency on the data - the `react-query` bits underneath will pull this from the cache if we've fetched it before.

hooks/useAvalancheForecast.ts
stevekuznetsov: > I disagree pretty strongly with this statement. I would much rather see new data pop in on that rare occasion 

The implication for this app, though, is that we show the wrong forecast to the user and can't be certain that we ever re-fresh and show the correct one. At first blush that seems like something we can't afford to do, even if, as you point out, the way that an application would handle this in the usual case is to prefer a faster experience.

> Also I think this behaves poorly if I'm offline just after midnight - there might be data in cache that's not very stale but I can't see it because it was cached as "yesterday's news". Whereas if it was cached under the "latest" key then I'd be able to see it, even if it was slightly stale.

We use a nominal forecast date, which cuts over at e.g. 6pm for NWAC. Midnight or not should have absolutely no bearing on this.

utils/date.ts
```diff
@@ -71,3 +75,35 @@ export const utcDateToDayOfWeekString = (date: Date | string | undefined): strin
...
+// we can cache a request for e.g. 01-01-2001 01:00:00 the same as 01-01-2001 02:00:00, since those will net the
+// same result.
+export type RequestedTime = Date | 'latest';
+export type RequestedTimeString = string | 'latest';
+export const requestedTimeToUTCDate = (requestedTime: RequestedTime): Date => {
+  if (requestedTime !== 'latest') {
+    return toDate(new Date(requestedTime), {timeZone: 'UTC'});
+  }
+
+  return toDate(new Date(), {timeZone: 'UTC'});
```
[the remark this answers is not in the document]
stevekuznetsov: > isn't this going to behave strangely around either side of midnight?

No, I don't think so - we use nominal dates where they matter, and we resolve to the millisecond.

components/observations/SimpleForm.tsx
stevekuznetsov: Oops, fixed

hooks/useMapLayerForecasts.ts
stevekuznetsov: done

components/AvalancheForecastZonePolygon.tsx
```diff
@@ -73,12 +73,14 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
 
   if (useAnimation) {
     const fillColor = animationProgress.interpolate({
-      inputRange: [0, 1, 2, 3],
+      inputRange: [0, 1, 2, 3, 4, 5],
```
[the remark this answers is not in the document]
stevekuznetsov: I was moreso trying to get the ratio of opaque to less-opaque right

stevekuznetsov:
@floatplane I hear your questions re: what our behavior looks like during the cut-over between forecast periods and how we render the app when the user's been offline for more than a day, but I think we'll merge this to allow for more iteration since this PR does not introduce the behavior you're calling into question, that's been how we operate for a while now.

PR #179, opened 2023-02-25
adding version using expo-application

yuliadub:
showed some weird numbers when I was using a simulator - want to try to see what the preview will look like if I do it form the build. 

https://docs.expo.dev/versions/latest/sdk/application/#applicationnativebuildversion


yuliadub:
well that didnt do what I expected , need to dig into this more :D 

File

yuliadub:
looks like expo go/emulator wont display versions right since that gets set at build time - tested this by running a preview build of android and installing the apk directly only my emulator and the right thing happened :D 

image

PR #187, opened 2023-02-26
forecast: show observations for the current zone

stevekuznetsov:


stevekuznetsov:
In adding the recent observations for the zone into the obs tab in the forecast, I had to make a different part of the view scrollable - note that the header with the zone name & icon, as well as the tabs, do not scroll off the screen any longer. Is this OK? It doesn't look like it's possible to have one scrolling view (the list of obs) inside of another (the whole screen).

https://user-images.githubusercontent.com/7328370/221436193-5b92e304-13ff-4f58-b5b5-8114722d1c57.MP4

PR #192, opened 2023-02-27
*: use a polyfill for timezones

stevekuznetsov:
@yuliadub this might fix the Android timezone issue - I will need to build local APKs and test it out.


stevekuznetsov:
Can't hurt to try, we can revert later ...

PR #193, opened 2023-02-27
Observation submission

floatplane:
This gets submission working on the happy path. There's still quite a few TODOs in here, but they're relatively minor. 

A TODO that's not in here is: the submitted obs can't be viewed in the app, because they are so minimal that they don't validate 😬


floatplane:
Fixed the validation error 😄 

Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-27 at 13 50 59

floatplane:
not sure why the image is in the observation 3 times but we'll figure that out

App.tsx
```diff
@@ -140,6 +141,8 @@ const asyncStoragePersister = createAsyncStoragePersister({
   storage: AsyncStorage,
 });
 
+clearUploadCache();
```
stevekuznetsov, line 144: Is this a TODO? In the final version would we, on startup, read this cache and re-try any uploads that had not succeeded?

components/observations/submitObservation.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const extensionToMimeType = (extension: string) => {
+  switch (extension.toLowerCase()) {
+    case 'jpeg':
+    case 'jpg':
+      return 'image/jpeg';
+    case 'png':
+      return 'image/png';
+    default:
+      throw new Error(`Unknown mime type for extension: ${extension}`);
```
stevekuznetsov, line 19: I imagine most folks will be using HEIC, it's standard on both iOS and Android - do we need something special to handle that?

components/observations/submitObservation.ts
```diff
@@ -0,0 +1,115 @@
...
+
+// TODO: should put an expiration time on these cache entries and clear them less aggressively.
+export const clearUploadCache = async () => {
+  const keys = await AsyncStorage.getAllKeys();
+  keys.filter(k => k.startsWith(imageUploadCachePrefix)).map(async k => await AsyncStorage.removeItem(k));
+};
+
+const uploadImage = async ({apiPrefix, uri, name, center_id}: {apiPrefix: string; center_id: AvalancheCenterID; uri: string; name: string | undefined}): Promise<MediaItem> => {
+  // This weird use of `slice` is because the version of Hermes that Expo is currently pinned to doesn't support `at()`
+  const filename = uri.split('/').slice(-1)[0];
```
stevekuznetsov, line 33: Do windows phones use the other slash? :trollface:

components/observations/submitObservation.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  const response = await axios.post<MediaItem>(`${apiPrefix}/v2/public/media`, payload, {
+    headers: {
+      // Public API uses the Origin header to determine who's authorized to call it
+      Origin: 'https://nwac.us',
+    },
+  });
+  await AsyncStorage.setItem(imageCacheKey, JSON.stringify(response.data));
```
stevekuznetsov, line 73: Why couldn't these also be `useMutation`s?

components/observations/submitObservation.ts
floatplane: don't know yet! need to test on my phone to see if this is transparently handled by the ImagePicker or not.

components/observations/submitObservation.ts
floatplane: URIs can't have backslashes but nice try 😉

components/observations/submitObservation.ts
floatplane: They don't play well with hooks: there are from 0-8 of them, and also we need to sequence these calls: the media have to be uploaded before the observation. I think it's cleaner this way.

PR #196, opened 2023-02-28
Minor obs submission fixes

floatplane:
- remove min length validations from text areas, to align with NAC forms
- add some missing max length validations on `cracking_description` and `collapsing_description`
- don't set a default location when the map has never been opened


components/observations/ObservationFormData.ts
```diff
@@ -92,6 +85,12 @@ export const simpleObservationFormSchema = z
           message: required,
           path: ['instability', 'cracking_description'],
         });
+      } else if (cracking_description.length > 1024) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: tooLong,
+          path: ['instability', 'cracking_description'],
```
stevekuznetsov, line 92: IIRC these map into enums - `'isolated', 'widespread'` - so should be a toggle?

floatplane:
also adding tab bar icons 
image

PR #208, opened 2023-03-01
eas.json: we don't want draft releases on android

stevekuznetsov:


yuliadub:
huh weird, in the past it wouldn't let me submit unless draft was there.

stevekuznetsov:
It's some weird thing - I think if you have the internal build, it has to be a draft, but if you don't it doesn't need to be - and if it is, you need to manually go click the publish button

PR #209, opened 2023-03-01
app.json: ios: we don't use export-restricted encryption

stevekuznetsov:


yuliadub:
yay! thats helpful to not have to click the encryption thing every time :D

PR #218, opened 2023-03-02
Update observation detail view styling

floatplane:
There's not a fully baked design here, so I did a bit of improvising. Generally went for tabular data reading across (label -> value) as it felt easier to read in portrait mode.

Known issue: NAC observations are not including location, so I'm hiding the map in that case. I expect we'll get that working today!

simulator_screenshot_2AB59CFA-ACEA-4246-8799-87766843AF8C
simulator_screenshot_E5B38C93-AB8B-4ED3-91A4-246CC6795FD2
simulator_screenshot_413C38E7-59F2-4B1E-A79D-1F1AEE357C10

### DIFF
--- a/components/observations/ObservationDetailView.tsx
+++ b/components/observations/ObservationDetailView.tsx
@@ -153,7 +153,7 @@ export const ObservationCard: React.FunctionComponent<{
               </View>
               <Card borderRadius={0} borderColor="white" header={<Title3Black>Summary</Title3Black>}>
                 <VStack space={8} width="100%">
-                  {observation.location_point && (
+                  {observation.location_point?.lat && observation.location_point?.lng && (
                     <ZoneMap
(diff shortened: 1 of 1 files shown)


floatplane:
Probably need to hide fields in the avalanche section if we don't have the data:

simulator_screenshot_BEFA4F99-2FB7-401A-9AD7-13C1DD9CF2A3

stevekuznetsov:
dundefined-rundefined is my fav

stevekuznetsov:
Looking awesome! Do we want to annotate the "yes" bits for avalanches/cracking/collapsing with the red + orange icons from the list view for consistency?

floatplane:
> Looking awesome! Do we want to annotate the "yes" bits for avalanches/cracking/collapsing with the red + orange icons from the list view for consistency?

that makes sense, I'll add that to the next PR!

PR #228, opened 2023-03-03
Image zoom and pan

floatplane:
This integrates a package for rendering the images in a modal and allowing zoom and pan. My only gripe with it is that it allows you to pan too far - you can pan the bottom of image almost to the top of the screen, for example. But other than that, it's a vast improvement over what we had before and makes the image browsing experience so much more useful.

https://user-images.githubusercontent.com/101196/222795836-68f93a58-f4d9-4d28-8957-18035a1751c3.mov

### DIFF
--- a/components/content/carousel/ImageViewerModal.tsx
+++ b/components/content/carousel/ImageViewerModal.tsx
@@ -1,6 +1,6 @@
 import React, {PropsWithChildren} from 'react';
 
-import {ScrollView} from 'react-native';
+import {ScrollView, TouchableOpacity} from 'react-native';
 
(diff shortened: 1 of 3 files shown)


stevekuznetsov (commented):
It's not immediately clear to me - what happens on load errors?

floatplane:
> It's not immediately clear to me - what happens on load errors?

interesting question, my guess is you end up with blank space but let me see

floatplane:
@stevekuznetsov infinite spinner when I pass it a URL that doesn't exist - seems appropriate

image

PR #231, opened 2023-03-03
Fix weather card rendering to deal with various permutations of missing data

floatplane:
This is a quick follow up to #229, which tried to be clever about only rendering the pieces of the weather section that actually had data. Now it's not clever, but is provably more correct, and has tests attached to prove it.

### DIFF
--- a/components/observations/ObservationDetailView.tsx
+++ b/components/observations/ObservationDetailView.tsx
@@ -125,6 +125,16 @@ export const WeatherCard = ({observation, ...props}: {observation: Observation} 
   );
 };
 
+export const withUnits = (value: string | number | null | undefined, units: string) => {
+  if (value == null) {
+    return 'Unknown';
+  } else if (Number.isNaN(Number(value))) {
+    return value;
+  } else {
+    return `${value}${units}`;
+  }
+};
+
 export const ObservationCard: React.FunctionComponent<{
   observation: Observation;
   mapLayer: MapLayer;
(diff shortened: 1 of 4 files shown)


components/observations/ObservationDetailView.tsx
```diff
@@ -93,13 +95,44 @@ export const TableRow = ({label, value}: {label: string; value: string}) => (
...
+};
+
 export const ObservationCard: React.FunctionComponent<{
   observation: Observation;
   mapLayer: MapLayer;
 }> = ({observation, mapLayer}) => {
   const navigation = useNavigation<ObservationsStackNavigationProps>();
   const {avalanches_observed, avalanches_triggered, avalanches_caught} = observation.instability;
 
+  log.info(observation);
```
stevekuznetsov, line 135: Likely want to remove this before merge?

components/observations/ObservationDetailView.tsx
```diff
@@ -93,13 +95,44 @@ export const TableRow = ({label, value}: {label: string; value: string}) => (
...
+    return null;
+  }
+
+  const {cloud_cover, air_temp, recent_snowfall, rain_elevation, snow_avail_for_transport, wind_loading} = weather;
+
+  return (
+    <Card borderRadius={0} borderColor="white" header={<BodyBlack>Weather</BodyBlack>} {...props}>
+      <VStack space={8} width="100%">
+        {/* Using Boolean() here so that we don't end up rendering empty strings inline */}
+        {Boolean(weather_summary) && <HTML source={{html: weather_summary}} />}
```
stevekuznetsov, line 114: TIL `Boolean()` for when you want _extra_ truthiness!

components/observations/ObservationDetailView.tsx
```diff
@@ -93,13 +95,44 @@ export const TableRow = ({label, value}: {label: string; value: string}) => (
...
+        {Boolean(cloud_cover) && <TableRow label={'Cloud Cover'} value={FormatCloudCover(cloud_cover as CloudCover)} />}
+        {Boolean(air_temp) && <TableRow label={'Temperature (F)'} value={air_temp || 'Unknown'} />}
```
stevekuznetsov, line 116: nit: do you think we can add something to `TableRow` that knows how to default? Perhaps a prop for the default value, which itself defaults to `'Unknown'` ?

stevekuznetsov (commented):
Nice, test approach looks good.

components/observations/ObservationDetailView.tsx
floatplane: it's all about what you want the result to be when `weather_summary` is an empty string...

```
❯ node
Welcome to Node.js v19.4.0.
Type ".help" for more information.
> '' && "something"
''
> Boolean('') && "something"
false
```

components/observations/ObservationDetailView.tsx
floatplane: well for this particular line, `|| 'Unknown'` doesn't actually do anything! we've already checked that `air_temp` has a value, so we don't need a default. I'll just rip that out.

components/observations/ObservationDetailView.tsx
floatplane: do you want `false`, or just falsy? 😵

PR #241, opened 2023-03-08
Update button styles

floatplane:
- pre-process our RGBA color definitions to remove the alpha channel, so that the color is consistent no matter what it renders on top of
- update button styles to match figma (font, colors, borders)
- add a button style view for easier debugging

simulator_screenshot_2D2F321E-A75A-496E-BC2E-A2466E06F38F


floatplane:
image

Use `primary` for the selected icon color and `text.secondary` for the deselected ones

PR #242, opened 2023-03-08
Revert "Update button styles"

floatplane:
Reverts stevekuznetsov/avalanche-forecast#241

I observed one instance of a color going wonky as a part of this PR (the 24 hour label in the weather station detail page). I need to double-check things before I commit this again!


floatplane:
Actually, reverting this is overkill - I have the fix

Issue #243, opened 2023-03-08
Chinook Pass (and maybe others) crashing


floatplane:
I think this may be related to the accumulated snowfall column but I need to dig in a bit

floatplane:
This is actually related to one of the 2 weather stations at Chinook Pass not having data points for each time 😬

floatplane:
The root cause of this issue (and the thing that's still broken) is that `TimeSeriesTable` assumes that the `date_time` arrays will be the same when fetching observations for multiple weather stations at once. That's not the case! 🤮 

This is a url that demonstrates the issue: `curl 'https://api.snowobs.com/wx/v1/station/data/timeseries/?stid=32%2C33&source=nwac&start_date=202303080200&end_date=202303090200&units=default&output=mesowest&calc_diff=false&raw_data=false&token=71ad26d7aaf410e39efe91bd414d32e1db5d'`. The "Chinook Pass Summit" station has 24 entries but the "Chinook Pass Base" station is missing 12 entries in the middle.

I guess one solution is to build a complete list of date_time entries in advance, set up the rows and then start iterating over the observations for each station. Maybe there's a more elegant way though.

PR #251, opened 2023-03-17
weather_data: handle unaligned time series

stevekuznetsov:
Different weather stations may have different times at which they record data; we need to handle the case where our columns are not entirely 'full'.


stevekuznetsov:
Fixes #243

PR #283, opened 2023-06-05
Looking into setting up mixpanel for app

yuliadub:
Found this implementation: https://www.npmjs.com/package/@bothrs/expo-mixpanel-analytics, added and tested it to send an event on when app starts (via expo on IOS, havent tried Android yet) The package I am using is basically wrapping around simple http calls, I'd say it likely makes sense once we have a better idea of what we want to track to just build our own vs taking a dependency (https://developer.mixpanel.com/reference/import-events)

Since the expected path seems to not work for expo (tried it and it comes back null on IOS, havent dug too much into it but likely due to some specific platform setting) 
https://github.com/mixpanel/mixpanel-react-native

Error I got when I tried mixpanel-react-native (possible that this error only happens in expo go and wouldnt happen in Test Flight, may be worth checking)
image

test event that this PR produces can be seen [here](https://mixpanel.com/project/3000673/view/3519389/app/events). 

p.s. added the Mixpanel token to our expo secrets.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -78,7 +78,7 @@ if(ExpoMixpanelAnalytics != null)
   const mixpanel_token = Constants.expoConfig.extra!.mixpanel;
   logger.info('Mixpanel is initializing.');
   // Only initialize Mixpanel if we can find the correct env setup
-  if (mixpanel_token === 'LOADED_FROM_ENVIRONMENT') {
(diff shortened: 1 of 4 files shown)


floatplane:
We need to recreate this PR on top of latest main, closing

PR #284, opened 2023-06-05
Attempt to fix observation pages so they load

yuliadub:
simply taking out NAC obs did the trick. Need to dig into what changed, but basically they always seem to come in as undefined and cause all kinds of fun null calls


stevekuznetsov:
@yuliadub one thing you could try if you want to fix them - run the [generators](https://github.com/stevekuznetsov/avalanche-forecast/blob/67068e757860280b2a121c81f07ac2a684ebdfff/package.json#L11-L13) in case they changed their schema

Issue #281, opened 2023-05-24
Observations views currently broken


yuliadub:
They all load a blank screen - I checked the NWAC obs additions Steve made and those are working as expected. I have not yet checked the NAC obs api but my guess is it is the culprit

yuliadub:
https://github.com/stevekuznetsov/avalanche-forecast/pull/284

PR #286, opened 2023-06-05
hooks: explicitly handle axios errors

stevekuznetsov:
We were doing mostly OK by letting axios just throw exceptions, but that eats the error messages and does not disambiguate between errors on a response, errors sending the request, and errors creating the reqeust object. Axios could just publish a helpful error, but whatever.


stevekuznetsov:
Fixes #281 
Obsoletes #284

PR #314, opened 2023-09-06
Show avalanche center selector on first startup, implement selector in settings

floatplane:
- Added preference storage and a `usePreferences` hook for storing things like `hasSeenCenterPicker` as well as the user's preferred avalanche center, persists across restarts
- Created an `AvalancheCenterList` component
- Created the startup dialog based on current Figma designs
- Didn't spend a lot of time making the debug version of this pretty - it just scales the icons way way down and uses the avalanche center IDs. But the code is almost entirely shared now.

dialog | settings menu | settings menu (debug)
--- | --- | ---
simulator_screenshot_76906B6C-EE6B-490C-8B56-5AC34B4813B1 | simulator_screenshot_65FF0C7E-BC14-41D6-B76A-216975400CD5 | simulator_screenshot_C05787A4-5D38-42BC-84AD-E15676309A58


Preferences.tsx
```diff
@@ -0,0 +1,61 @@
...
+  children?: ReactNode;
+}
+
+export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({children}) => {
+  const [preferences, setFullPreferences] = useState<Preferences>(defaultPreferences);
+
+  useAsyncEffect(async () => {
+    const storedPreferences = await AsyncStorage.getItem('preferences');
+    if (storedPreferences) {
+      setPreferences(merge({}, defaultPreferences, JSON.parse(storedPreferences) as Preferences));
```
stevekuznetsov, line 41: We should likely use `zod` here as we do elsewhere and invalidate the file if it's not correct. The `as Preferences` cast will just cause runtime failures later if it's wrong.

App.tsx
```diff
@@ -235,7 +239,14 @@ const BaseApp: React.FunctionComponent<{
 }> = ({staging, setStaging}) => {
   const colorScheme = useColorScheme();
   const {logger} = React.useContext<LoggerProps>(LoggerContext);
-  const [avalancheCenterId, setAvalancheCenterId] = React.useState((Constants.expoConfig?.extra?.avalanche_center as AvalancheCenterID) ?? 'NWAC');
+  const {preferences, setPreferences} = usePreferences();
+  const avalancheCenterId = preferences.center;
+  const setAvalancheCenterId = useCallback(
```
stevekuznetsov, line 244: Out of curiosity, why did you need to `useCallback` here and elsewhere?

Preferences.tsx
floatplane: that's a good idea!

App.tsx
floatplane: `useCallback` guarantees that the value of the function object is stable if the dependencies don't change. that helps you avoid unnecessary re-renders when you're passing the function as a prop to subcomponents. without `useCallback` the function gets a new value on every render pass.

stevekuznetsov: Gotcha, makes sense - thanks!

PR #316, opened 2023-09-07
Zod preferences

floatplane:
cc @stevekuznetsov, as suggested in #314

### DIFF
--- a/Preferences.tsx
+++ b/Preferences.tsx
@@ -5,13 +5,16 @@ import AsyncStorage from '@react-native-async-storage/async-storage';
 import {merge} from 'lodash';
 import React, {createContext, ReactNode, useContext, useEffect, useState} from 'react';
 
-import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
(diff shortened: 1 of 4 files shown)


Preferences.tsx
```diff
@@ -36,14 +41,20 @@ export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({childre
   const [preferences, setFullPreferences] = useState<Preferences>(defaultPreferences);
 
   useAsyncEffect(async () => {
-    const storedPreferences = await AsyncStorage.getItem('preferences');
+    let storedPreferences = {};
+    try {
+      storedPreferences = preferencesSchema.parse(JSON.parse((await AsyncStorage.getItem(PREFERENCES_KEY)) ?? '{}'));
+    } catch (_e) {
```
stevekuznetsov, line 47: Will Sentry pick up this exception and send it to us? If not, can we log it? I imagine we will want to dig into those

stevekuznetsov (commented):
Looks good - just one question

floatplane:
now logging unexpected errors to Sentry

PR #317, opened 2023-09-07
Settings

floatplane:
Updates to the settings page! Brings things into line with current Figma designs.

- app feedback button - currently opens an email to @charlotteguard 🙈 
- uses different links in General for different centers
- doesn't render "General" section for centers other than NWAC and SNFAC
- move debugging tools into a "Secret Menu" section so they don't clutter the view for non-engineers. this section is collapsed by default, but remembers its state across app runs

SNFAC | NWAC | secret menu collapsed | secret menu expanded
--- | --- | --- | ---
simulator_screenshot_DE55C182-DE19-476B-BA38-2DA39ED84A50 | simulator_screenshot_0B45E4AC-07CE-44B0-BBCC-1A7938CE22B8 | simulator_screenshot_F673D29D-7DEF-4D83-BF55-F417F037E80B | simulator_screenshot_7DF6D910-E7D6-45A5-ADAF-C37E6DACB644

### DIFF
--- a/Preferences.tsx
+++ b/Preferences.tsx
@@ -13,16 +13,14 @@ import {z} from 'zod';
 export const PREFERENCES_KEY = 'PREFERENCES';
 
 const preferencesSchema = z.object({
-  center: avalancheCenterIDSchema,
-  hasSeenCenterPicker: z.boolean(),
+  center: avalancheCenterIDSchema.default('NWAC'),
+  hasSeenCenterPicker: z.boolean().default(false),
+  secretMenuCollapsed: z.boolean().default(true),
 });
 
 export type Preferences = z.infer<typeof preferencesSchema>;
 
-const defaultPreferences: Preferences = {
-  center: 'NWAC',
-  hasSeenCenterPicker: false,
-};
+const defaultPreferences = preferencesSchema.parse({});
 
 interface PreferencesContextType {
   preferences: Preferences;
--- a/components/content/Card.tsx
+++ b/components/content/Card.tsx
@@ -37,11 +37,16 @@ export const Card: React.FunctionComponent<PropsWithChildren<CardProps>> = ({hea
 
 export interface CollapsibleCardProps extends CardProps {
   startsCollapsed: boolean;
+  collapsedStateChanged?: (collapsed: boolean) => void;
 }
 
-export const CollapsibleCard: React.FunctionComponent<PropsWithChildren<CollapsibleCardProps>> = ({startsCollapsed, header, children, ...props}) => {
+export const CollapsibleCard: React.FunctionComponent<PropsWithChildren<CollapsibleCardProps>> = ({startsCollapsed, collapsedStateChanged, header, children, ...props}) => {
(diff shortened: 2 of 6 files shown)


Preferences.tsx
```diff
@@ -13,16 +13,14 @@ import {z} from 'zod';
 export const PREFERENCES_KEY = 'PREFERENCES';
 
 const preferencesSchema = z.object({
-  center: avalancheCenterIDSchema,
-  hasSeenCenterPicker: z.boolean(),
+  center: avalancheCenterIDSchema.default('NWAC'),
+  hasSeenCenterPicker: z.boolean().default(false),
+  secretMenuCollapsed: z.boolean().default(true),
```
floatplane, line 18: before, you had to add a field here and then remember to add it to `defaultPreferences`. now you just add a field once (and if you forget to set a default, line 23 will throw an exception immediately)

components/content/Card.tsx
```diff
@@ -37,19 +37,24 @@ export const Card: React.FunctionComponent<PropsWithChildren<CardProps>> = ({hea
 
 export interface CollapsibleCardProps extends CardProps {
   startsCollapsed: boolean;
+  collapsedStateChanged?: (collapsed: boolean) => void;
```
floatplane, line 40: there wasn't a way to get notified when the collapsed state of a card changed - now there is

data/settingsMenuItems.ts
```diff
@@ -0,0 +1,46 @@
...
+      url: 'https://friends.sawtoothavalanche.com/make-a-donation/',
+    },
+    {
+      title: 'News and Events',
+      url: 'https://friends.sawtoothavalanche.com/news-events/',
+    },
+  ],
+};
+
+export const getMenuItemsForCenter = (avalancheCenterId: AvalancheCenterID) => settingsMenuItems[avalancheCenterId] || ([] as {title: string; url: string}[]);
```
stevekuznetsov, line 46: Isn't this also an invalid TypeScript cast? The array items do *not* have title and url. I think the correct type is 

```ts
export const getMenuItemsForCenter = (avalancheCenterId: AvalancheCenterID): {title: string, url: string}[] | undefined => settingsMenuItems[avalancheCenterId]
```
```

floatplane: I intentionally didn’t want to return undefined, so this cast means that the empty array is narrowed to the same type as the non-empty one. Not sure what is invalid here?

stevekuznetsov: `[]` does not have `title` or `url` properties in its members - yes, you'd never be able to tell since it has no items in it, but it's not a correct type. 

Also it looks like without any type gymnastics, the inferred type here is correct:

Screenshot from 2023-09-08 07-41-59

Just seems like if we're going to use TypeScript, using the escape hatch should be something we do sparingly and carefully.

stevekuznetsov: I don't know that I did a good job communicating it -

> Not sure what is invalid here?

Using the literal ` as ` to override inferred types instead of changing the program to be correct as per `tsc`.

PR #321, opened 2023-09-08
A few miscellaneous things

floatplane:
- avalanche center picker elements weren't center-aligned (icon/text/checkbox)
- observation list feedback from [Figma](https://www.figma.com/file/Bai2w7x0vUakL5WRjPGRg7/Native-App?type=design&node-id=1312-174816&mode=design&t=kFWAMAmCxjIJCpB3-0)
- added a `busy` prop to `Button` for the common case of needing to append an `ActivityIndicator` to the text within
  - replace existing usage in observation submission
  - switch to this pattern when loading more observations

https://github.com/stevekuznetsov/avalanche-forecast/assets/101196/0f1eccee-8e1e-4100-81a1-371fc1611c44

### DIFF
--- a/components/content/AvalancheCenterList.tsx
+++ b/components/content/AvalancheCenterList.tsx
@@ -18,8 +18,8 @@ const AvalancheCenterListItem: React.FC<AvalancheCenterListItemProps> = ({data, 
   const center_id: AvalancheCenterID = data.center.id as AvalancheCenterID;
   return (
     <TouchableOpacity onPress={() => setSelected(center_id)} activeOpacity={0.8}>
-      <HStack justifyContent="space-between" alignItems="flex-start" space={4}>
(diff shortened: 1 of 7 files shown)


floatplane:
also cleaned up some empty state weirdness here
- now using `ListEmptyComponent` which is cool
- fixed typo: `could not find the any`
- add some padding to the bottom of the `Outcome` component
- hide the "Load more" button since I think it's a no-op in the case of an empty view? @stevekuznetsov correct me if that's wrong. but if that's wrong then the button probably needs a different label, like "Search deeper" or something. or we need to implement infinite scroll...

before | after
--- | ---
simulator_screenshot_CEBB7812-CCBD-4881-8323-2CB4E611BDDB | simulator_screenshot_D8849B61-50D1-4CA0-9DFB-F24C6DBFDBB1

PR #322, opened 2023-09-08
Add time machine for debugging

floatplane:
It's really hard to work on stuff in the summertime, so I made a time machine!

screenshot | in action
--- | ---
 |

Notes

- I wanted a simpler way to flip *all* the parts of the app to a new day (and hopefully it's useful to browse particular days when investigating bugs). The existing mechanism in the screens section didn't change the query date for Weather Data, for example
- Observations list is still querying starting at `now`, I haven't looked into that but I assume/hope it's pretty easy to fix
- I didn't make the selected date sticky across runs
- There's nothing in the UI that tells you that you're browsing a particular date (which is one reason I didn't make it sticky)
- There's nothing to stop you from picking a stupid date like 1855, so don't do that
- I'll keep building the collection of good days/times to view
  - This could be especially useful when exploring those weird times around midnight UTC / local midnight / local forecast update time 😬

### DIFF
--- a/components/observations/ObservationsFilterForm.tsx
+++ b/components/observations/ObservationsFilterForm.tsx
@@ -42,30 +42,30 @@ const observationFilterConfigSchema = z
   .deepPartial();
 export type ObservationFilterConfig = z.infer<typeof observationFilterConfigSchema>;
 
-const matchesDates = (dates: z.infer<typeof observationFilterConfigSchema.shape.dates>, observation: ObservationFragment): boolean => {
+const matchesDates = (currentDate: Date, dates: z.infer<typeof observationFilterConfigSchema.shape.dates>, observation: ObservationFragment): boolean => {
   let matches = true;
   if (!dates) {
     return matches;
   }
   switch (dates.value) {
     case 'past_day':
-      dates.to = new Date();
+      dates.to = currentDate;
       dates.from = sub(dates.to, {days: 1});
       break;
     case 'past_3_days':
-      dates.to = new Date();
+      dates.to = currentDate;
       dates.from = sub(dates.to, {days: 3});
       break;
     case 'past_week':
-      dates.to = new Date();
+      dates.to = currentDate;
       dates.from = sub(dates.to, {weeks: 1});
       break;
     case 'past_month':
-      dates.to = new Date();
+      dates.to = currentDate;
       dates.from = sub(dates.to, {months: 1});
       break;
     case 'past_season':
-      dates.to = new Date();
+      dates.to = currentDate;
       // the season starts Nov 1st, that's either this year or last year depending on when we are asking
       if (getMonth(dates.to) <= 11) {
         dates.from = new Date(`${getYear(sub(dates.to, {years: 1}))}-11-01`);
@@ -119,7 +119,7 @@ const matchesAvalancheInstability = (instability: avalancheInstability, observat
   throw new Error(`Unknown instability: ${invalid}`);
 };
 
-export const filtersForConfig = (mapLayer: MapLayer, config: ObservationFilterConfig): ((fragment: ObservationFragment) => boolean)[] => {
+export const filtersForConfig = (mapLayer: MapLayer, config: ObservationFilterConfig, currentDate: Date): ((fragment: ObservationFragment) => boolean)[] => {
   if (!config) {
     return [];
   }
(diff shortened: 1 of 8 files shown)


components/observations/ObservationsFilterForm.tsx
```diff
@@ -42,30 +42,30 @@ const observationFilterConfigSchema = z
...
 
-const matchesDates = (dates: z.infer<typeof observationFilterConfigSchema.shape.dates>, observation: ObservationFragment): boolean => {
+const matchesDates = (currentDate: Date, dates: z.infer<typeof observationFilterConfigSchema.shape.dates>, observation: ObservationFragment): boolean => {
   let matches = true;
   if (!dates) {
     return matches;
   }
   switch (dates.value) {
     case 'past_day':
-      dates.to = new Date();
```
floatplane, line 52: using `new Date()` here breaks date filtering when we're pinned to a non-default date

components/observations/ObservationsListView.tsx
```diff
@@ -95,7 +96,7 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
           </HStack>
         }
         ListFooterComponent={
-          displayedObservations.length > 0 && (nacObservationsResult.hasNextPage || nwacObservationsResult.hasNextPage) ? (
+          nacObservationsResult.hasNextPage || nwacObservationsResult.hasNextPage ? (
```
floatplane, line 99: this line brings the button back when the list is empty

hooks/useNWACObservations.ts
```diff
@@ -17,30 +17,34 @@ import {ZodError} from 'zod';
 export const useNWACObservations = (center_id: AvalancheCenterID, endDate: RequestedTime) => {
   const {nwacHost} = React.useContext<ClientProps>(ClientContext);
   const {logger} = React.useContext<LoggerProps>(LoggerContext);
-  const key = queryKey(nwacHost, center_id);
+  const key = queryKey(nwacHost, center_id, endDate);
```
floatplane, line 20: the `queryKey` function was ignoring dates, so it was impossible to query for observations in a different date range - you'd always get whatever preloaded observations had been cached. 

note that `endDate` can be `'latest'` here, which is also what the preloaded query uses. that means that when viewing the latest info, we'll get the preloaded obs from the cache even though there's a delta between the time of preloading and the actual time that the obs screen is viewed.

hooks/useNWACObservations.ts
```diff
@@ -17,30 +17,34 @@ import {ZodError} from 'zod';
...
   const thisLogger = logger.child({query: key});
   thisLogger.debug('initiating query');
-  const fetchNWACObservationsPage = async ({pageParam = formatRequestedTime(endDate)}): Promise<ObservationsQueryWithMeta> => {
-    const endDate: Date = requestedTimeToUTCDate(parseRequestedTimeString(pageParam));
-    const startDate = sub(endDate, {weeks: 2});
-    return fetchNWACObservations(nwacHost, center_id, startDate, endDate, thisLogger);
+  const fetchNWACObservationsPage = async (props: {pageParam?: unknown}): Promise<ObservationsQueryWithMeta> => {
```
floatplane, line 23: there was a bug here - the pageParam variable didn't get a default assignment because it wasn't undefined, it was null.

floatplane (commented):
@stevekuznetsov you should take a look at the most recent commit in particular and make sure it makes sense to you. I added some notes below:

floatplane:
here's a movie showing that the obs list keeps working at startup, but now it also works when viewing a different date. 🎉 

https://github.com/stevekuznetsov/avalanche-forecast/assets/101196/4e34cc57-5b2a-4814-abce-61008c16b6b1

hooks/useNWACObservations.ts
stevekuznetsov: Good catch. I think this is the behavior we want anyway, that's the point of `'latest'` - if you're trying to get the latest, we want to cache something for you but we can't use the current time.

hooks/useNWACObservations.ts
stevekuznetsov: Gotta love JavaScript :melting_face:

stevekuznetsov (approved):
Awesome work! This is what I had in mind exactly when I did all the work leading up to this to pass around the requested time :)

PR #323, opened 2023-09-11
Observation list tweaks, fix some react warnings

floatplane:
This is a mixed bag of changes, but each commit is standalone. In order:

- Observation list feedback from Figma
- Use React.memo to make list item components pure - should (hopefully) address console warnings of the form `VirtualizedList: You have a large list that is slow to update`. LMK if they keep popping up
- Calling `Toast.show` directly from a render function triggers errors of the form `Cannot update a component from inside the function body of a different component.`. You have to call it from an input handler or from `useEffect` cc @stevekuznetsov 
- The NetworkImage component wasn't rendering its spinner correctly, fixed that
- Make Observation list cells fixed height, so scrolling is smooooth

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -6,7 +6,6 @@ import {
   FlatList,
   GestureResponderEvent,
   LayoutChangeEvent,
-  ListRenderItem,
   NativeScrollEvent,
   NativeSyntheticEvent,
   PanResponder,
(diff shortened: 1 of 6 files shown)


floatplane:
adding one more quick fix - applying the same change to `useNACObservations` that I did for NWAC in #322, so that we can time travel correctly with NAC obs. tested with Bridger-Teton:
simulator_screenshot_09764866-E7FB-4243-BDA5-A32856B29763

stevekuznetsov:
good stuff!

Issue #325, opened 2023-09-11
Make "Reset the query cache" more selective


floatplane:
Right now it blows away all of local storage, including preferences.

Would be nicer if the query cache entries were nested under a key so this delete action could be more targeted.

stevekuznetsov:
Are we storing in separate files or directories? Maybe there's a simple file browser we can add that would allow for choosing what to delete by iterating over what's in the local dir

floatplane:
I don't think files & directories are the right mental model, AsyncStorage is basically a [key-value store](https://react-native-async-storage.github.io/async-storage/docs/api).

It looks like `createAsyncStoragePersister` takes an optional `key` argument and can nest everything under that. Then we can change `AsyncStorage.clear()` to `AsyncStorage.removeItem(THE_MAGIC_KEY_NAME)`. I'll pick this up later today. https://github.com/stevekuznetsov/avalanche-forecast/blob/11fa07109f4e559e5073a22baec9c8a03e4d7c19/components/screens/MenuScreen.tsx#L220-L230

PR #326, opened 2023-09-12
Infinite scroll in observations list

floatplane:
Big change: infinite scroll working with the observations list. This works with both NAC and NWAC. There is an unfortunate interaction with the date filtering option that I haven't completely figured out how to handle. @stevekuznetsov I'll tag you in a comment next to the thing to figure out.

Two additional small changes: 
- add a background tint to the pressed state of normal buttons, so you can tell when you've pressed them
- make "Reset the query cache" more selective, fixing #325 

Obligatory movie:

https://github.com/stevekuznetsov/avalanche-forecast/assets/101196/9c6753f9-2e2e-4eb9-b97a-41f2a1c3cb4c

### DIFF
--- a/components/observations/ObservationsListView.tsx
+++ b/components/observations/ObservationsListView.tsx
@@ -36,6 +36,10 @@ interface ObservationsListViewProps extends Omit<FlatListProps<ObservationsListV
   initialFilterConfig?: ObservationFilterConfig;
 }
 
+interface ObservationFragmentWithPageIndex extends ObservationFragment {
+  pageIndex: number;
+}
+
 export const ObservationsListView: React.FunctionComponent<ObservationsListViewProps> = ({center_id, requestedTime, initialFilterConfig, ...props}) => {
   const endDate = requestedTimeToUTCDate(requestedTime);
   const originalFilterConfig: ObservationFilterConfig = {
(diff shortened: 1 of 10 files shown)


components/observations/ObservationsListView.tsx
```diff
@@ -50,22 +54,55 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
-  const nacObservationsResult = useNACObservations(center_id, requestedTime);
-  const nacObservations: ObservationFragment[] = nacObservationsResult.data?.pages?.flatMap(page => page.data) ?? [];
-  const nwacObservationsResult = useNWACObservations(center_id, requestedTime);
-  const nwacObservations: ObservationFragment[] = nwacObservationsResult.data?.pages?.flatMap(page => page.data) ?? [];
-  const observations: ObservationFragment[] = nacObservations
+  // TODO brian wire up duration correctly - there's an interaction between useInfiniteQuery (which has a cached
+  // notion of whether there's more data) and this duration, and I'm not sure of the best way to solve it.
+  // Right now, the queryKey in useNACObservations/useNWACObservations doesn't use the duration, so it's not invalidated when the duration changes.
+  // That has the nice property of not re-fetching data when the duration changes, but it also means that if you make the duration longer, you won't get new data.
+  const nacObservationsResult = useNACObservations(center_id, requestedTime, DEFAULT_OBSERVATIONS_WINDOW);
```
floatplane, line 61: @stevekuznetsov this is the bit that I'm not sure how best to solve.

maybe one approach would be to make `DEFAULT_OBSERVATIONS_WINDOW` a constant that's as big a lookback as we'd ever allow, and then change the `fetchMoreData` logic in this component to look at the date of the oldest observation that's been fetched and compare it to the current date filter (instead of just looking at `hasNextPage` the way it does today).

floatplane: I'm going to try and implement that approach as a followup PR (along with any other feedback)

components/observations/ObservationsListView.tsx
```diff
@@ -95,23 +127,26 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
-              </Button>
-            </HStack>
-          ) : null
-        }
+        ListFooterComponent={() => {
+          if (!nacObservationsResult.hasNextPage && !nwacObservationsResult.hasNextPage) {
+            return (
+              <Center height={80}>
+                <Body>
+                  {/* this is super-sketchy, need to implement proper string localization/formatting */}
```
stevekuznetsov: Can we get an issue to document how you would like this to be done in a proper way?

floatplane: Good call, created #329

PR #328, opened 2023-09-12
Fetch window

floatplane:
Following up from #326, this implements the idea I mentioned in https://github.com/stevekuznetsov/avalanche-forecast/pull/326#discussion_r1323193044

> maybe one approach would be to make `DEFAULT_OBSERVATIONS_WINDOW` a constant that's as big a lookback as we'd ever allow, and then change the `fetchMoreData` logic in this component to look at the date of the oldest observation that's been fetched and compare it to the current date filter (instead of just looking at `hasNextPage` the way it does today).

and that actually seems to work well! Now you can switch between date ranges without invalidating any cached results, and you can always scroll to the end of your current date range.

There's also a bonus fix in dfce4b63e21e6981a79871c3bc45451b163fee83 - I just noticed that I'd introduced a warning about a require cycle from App -> MenuScreen -> App, and that commit breaks it.


network/prefetchAllActiveForecasts.ts
```diff
@@ -57,12 +55,14 @@ export const prefetchAllActiveForecasts = async (
   }
 
   const endDate: Date = currentDateTime;
-  const startDate = add(endDate, DEFAULT_OBSERVATIONS_WINDOW);
   if (center_id === 'NWAC') {
-    void NWACObservationsQuery.prefetch(queryClient, nwacHost, center_id, startDate, endDate, logger);
+    // NWAC fetches in pages of 50 working backwards from endDate
```
stevekuznetsov, line 59: @floatplane the days of the NWAC obs API might be numbered but if you'd like I can try to make it do better with 2wk periods if it would make the rest of our code easier to grok here

floatplane: I think it's pretty reasonable to leave it alone - in both cases we prefetch a chunk of results, I can't really picture a scenario where anyone notices the difference in paging behavior!

PR #330, opened 2023-09-13
Observation list filtering UI

floatplane:
Worked with Erica on this design, it is still subject to change. Also, changing the filters seems *very* slow - will need to dig into that.

https://github.com/stevekuznetsov/avalanche-forecast/assets/101196/3a46c046-3bd4-444d-9999-f5b8358dcef0

### DIFF
--- a/components/observations/ObservationsFilterForm.tsx
+++ b/components/observations/ObservationsFilterForm.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import {startCase} from 'lodash';
+
 import {AntDesign} from '@expo/vector-icons';
 import {zodResolver} from '@hookform/resolvers/zod';
 import {SelectModalProvider} from '@mobile-reality/react-native-select-pro';
@@ -24,11 +26,14 @@ import {z} from 'zod';
 const avalancheInstabilitySchema = z.enum(['observed', 'triggered', 'caught']);
 type avalancheInstability = z.infer<typeof avalancheInstabilitySchema>;
 
+const dateValueSchema = z.enum(['past_day', 'past_3_days', 'past_week', 'past_month', 'past_season', 'custom']);
+type DateValue = z.infer<typeof dateValueSchema>;
+
 const observationFilterConfigSchema = z
   .object({
     zone: z.string(),
     dates: z.object({
-      value: z.enum(['past_day', 'past_3_days', 'past_week', 'past_month', 'past_season', 'custom']),
+      value: dateValueSchema,
       from: z.date(),
       to: z.date(),
     }),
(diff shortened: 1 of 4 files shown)


components/observations/ObservationsFilterForm.tsx
```diff
@@ -72,20 +103,11 @@ export const datesForFilterConfig = (dates: z.infer<typeof observationFilterConf
...
-  return (observation: ObservationFragment) => {
-    if (dates.from) {
-      matches = matches && isAfter(parseISO(observation.createdAt), startDate);
-    }
-    if (dates.to) {
-      matches = matches && isBefore(parseISO(observation.createdAt), endDate);
-    }
-    return matches;
-  };
+  return (observation: ObservationFragment) => isAfter(parseISO(observation.createdAt), startDate) && isBefore(parseISO(observation.createdAt), endDate);
```
stevekuznetsov, line 110: Not something you need to change, but I would like to mention that the moment you need to debug any of this or add a logging line, the one-liner starts to be something to shake your fist at :)

components/observations/ObservationsFilterForm.tsx
```diff
@@ -123,26 +145,54 @@ const matchesAvalancheInstability = (instability: avalancheInstability, observat
...
+    filterFuncs.push({filter: matchesDates(currentDate, config.dates), label: dateLabelForFilterConfig(config.dates)});
   }
 
-  if (config.dates && config.dates.value) {
-    filterFuncs.push(matchesDates(currentDate, config.dates));
+  if (config.zone) {
+    filterFuncs.push({
+      filter: observation => config.zone === matchesZone(mapLayer, observation.locationPoint?.lat, observation.locationPoint?.lng),
+      label: config.zone,
+      removeFilter: config => ({...config, zone: undefined}),
```
stevekuznetsov, line 167: Is there a reason to choose a spread here (which would allocate a whole new object & copy all data, right?) over just mutating in place?

components/observations/ObservationsFilterForm.tsx
floatplane: well, the old version was actually incorrect (dates.to/dates.from aren't always set) so this is an improvement in that sense 😂 

prettier is the thing that aggressively turns these into one-liners, which I do have mixed feelings about

components/observations/ObservationsFilterForm.tsx
floatplane: Yeah, it's intentional, the input comes from a call to `useState` (so should be treated as immutable) and the return value needs to be passed to `setFilterConfig`.

Compared to the number of object copies we do in the observation list proper, I guarantee you this is not worth worrying about from a perf perspective 🙈 

But I do take the point that the naming here (`removeFilter`) is super-confusing, because it's not actually removing the filter. Let me know if you have better ideas!

components/observations/ObservationsFilterForm.tsx
stevekuznetsov: Didn't the previous checks (` if (dates.from) {`) handle the unset case?

components/observations/ObservationsFilterForm.tsx
stevekuznetsov: Nope, your explanation makes sense, nothing needs to change here

components/observations/ObservationsFilterForm.tsx
floatplane: no, the logic around those fields changed. `dates.from` / `dates.to` are now only set for custom date ranges. `datesForFilterConfig` wraps up the logic of extracting them as needed.

https://github.com/stevekuznetsov/avalanche-forecast/blob/0ed7d41314b81a2edbcae4f9786222bbd601f72f/components/observations/ObservationsFilterForm.tsx#L109-L110

floatplane:
Updated so that you can't clear the zone filter when looking at the zone forecast: 
image

PR #337, opened 2023-09-18
Import the correct ScrollView object

### LINKED ISSUE #332: Select Avalanche Center (debug) menu does not scroll vertically
Problem is in title.

cc @floatplane

floatplane:
Fixes #332


floatplane:
Updated with ESLint error to catch this sort of thing:
image

PR #338, opened 2023-09-19
Upgrade to Expo 48

floatplane:
- Apply necessary package upgrades
- Fix bugs uncovered by new packages/new type definitions
  - Event handlers that seemed to be incorrect
  - Missing/invalid color arguments for Text instances
- Simplify `VStack/HStack` implementation to use new `columnGap/rowGap` style props
- Get Expo Doctor passing and run it in CI (hopefully 🤞)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -174,9 +174,9 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
             mt={8}
             px={4}
             flex={1}
-            onLayout={(event: LayoutChangeEvent) => {
+            onLayout={() => {
               // onLayout returns position relative to parent - we need position relative to screen
-              event.currentTarget.measureInWindow((x, y, width, height) => {
+              topElements.current?.measureInWindow((x, y, width, height) => {
```
stevekuznetsov, line 179: Are we guaranteed to have the most up-to-date reference in `topElements.current` when the `onLayout` fires?

patches/@react-native-segmented-control+segmented-control+2.4.0.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/node_modules/@react-native-segmented-control/segmented-control/index.d.ts b/node_modules/@react-native-segmented-control/segmented-control/index.d.ts
```
stevekuznetsov, line 1: Why did we need this patch?

stevekuznetsov:
New `expo doctor` failed:

```

components/AvalancheForecastZoneMap.tsx
floatplane: The deal with component refs is that it's either the value, or it hasn't been set yet (in which case `measureInWindow` just doesn't get invoked)

patches/@react-native-segmented-control+segmented-control+2.4.0.patch
stevekuznetsov: Sounds good! I'll open an issue to undo this when we bump for the next Expo release

PR #371, opened 2023-09-22
Make pull request closed action more robust

floatplane:
Only try to delete the channel if it doesn't exist

### DIFF
--- a/.github/actions/expo/channel/delete/action.yaml
+++ b/.github/actions/expo/channel/delete/action.yaml
@@ -23,5 +23,8 @@ runs:
       id: eas-update
       shell: bash
       run: |
-        eas channel:delete --non-interactive ${{ inputs.channel }}
-        eas branch:delete --non-interactive ${{ inputs.channel }}
+        eas channel:view ${{ inputs.channel }}  --non-interactive --limit 1
+        if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


floatplane:
guess we'll find out if i did this right

.github/actions/expo/channel/delete/action.yaml
```diff
@@ -23,5 +23,8 @@ runs:
       id: eas-update
       shell: bash
       run: |
-        eas channel:delete --non-interactive ${{ inputs.channel }}
-        eas branch:delete --non-interactive ${{ inputs.channel }}
+        eas channel:view ${{ inputs.channel }}  --non-interactive --limit 1
+        if [ $? -eq 0 ]; then
```
stevekuznetsov, line 27: you can also

```sh
if eas channel:view ${{ inputs.channel }}  --non-interactive --limit 1; then
```

stevekuznetsov: Which you'd need to do if you had `set -o errexit` set

floatplane: it seems like that may be implicitly set by the GH job runner, since my approach didn't work!

floatplane: thanks for the tip :)

Issue #8, opened 2022-12-08
Node version errors in Actions


floatplane:
Action runs are throwing errors ([example](https://github.com/stevekuznetsov/avalanche-forecast/actions/runs/3651928599)):
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-node@v2
```

The references to those actions reference Node 16: 
```yaml
      - name: 🏗 Setup Node
        uses: actions/setup-node@v2
        with:
          node-version: 16.x
          cache: yarn
```
but that doesn't seem to be enough. maybe update to `actions/setup-node@v3`? 🤷‍♂️

PR #377, opened 2023-09-25
Fix "The following actions uses node12 which is deprecated and will be forced to run on node16"

yuliadub:
fixes https://github.com/NWACus/avy/issues/8. Just needed to update actions/checkout@v2->v3 and ctions/setup-node@v2 -> v3, based on documentation that is where node 16 is supported. Tried on the update and seemed to not break anything. 

https://github.com/NWACus/avy/actions/runs/6294661816 -> no more warnings :D


floatplane:
yay thank you @yuliadub!

PR #381, opened 2023-09-25
Update NWAC text and links in About page

floatplane:
This is still pending review by @charlotteguard 

image


floatplane:
get rid of `(Learn)`

Issue #383, opened 2023-09-25
Logs not being saved to `logFilePath` anymore


floatplane:
#275 broke the facility that piped logs to a file in `logFilePath`. It should be pretty straightforward to bring this back, just going to require some Bunyan doc reading.

Issue #382, opened 2023-09-25
Mail sending seems broken on iOS


floatplane:
Filed https://github.com/expo/expo/issues/24613 to dig into this further

stevekuznetsov:
This looks like an issue with phone setups and apple mail composer, not sure we have something we can do here?

yuliadub:
the issue referenced above gives a clear explanation and seems like nothing to do for us here.

PR #384, opened 2023-09-25
Add sendMail wrapper, pick different mailing strategies on iOS / Android

floatplane:
Android seems happy to use MailComposer, but iOS doesn't. I'm hoping that gets fixed and the `sendMail` wrapper becomes simpler, but this seems to work on both types of device for now.

Also, disabling the "Email log file" button since logging to file is currently broken.

Related issues:
#382 
#383 
https://nwac.sentry.io/issues/4499867030
https://nwac.sentry.io/issues/4498064340


floatplane:
Creating https://github.com/NWACus/avy/pull/385 to fix this build break

components/screens/MenuScreen.tsx
```diff
@@ -128,7 +128,7 @@ export const MenuScreen = (queryCache: QueryCache, avalancheCenterId: AvalancheC
                 </Body>
               </Card>
               <View py={16} px={32}>
-                <Button buttonStyle="primary" onPress={() => void Linking.openURL(`mailto:charlotte@nwac.us?subject=I%20have%20thoughts`)}>
+                <Button buttonStyle="primary" onPress={() => void sendMail({to: 'charlotte@nwac.us', subject: 'NWAC app feedback', logger})}>
```
stevekuznetsov, line 131: nit: can we send this to `developer@nwac.us`? No need to bottle-neck with Charlotte if we don't need to

network/sendMail.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as FileSystem from 'expo-file-system';
+import * as Linking from 'expo-linking';
+import * as MailComposer from 'expo-mail-composer';
+
+interface SendMailProps {
+  to: string;
+  subject: string;
+  body?: string;
+  attachments?: string[];
+  logger?: Logger;
```
stevekuznetsov, line 14: Why should we allow this to be undefined?

components/screens/MenuScreen.tsx
floatplane: I'll do `developer+app-feedback@nwac.us` so we can filter it more easily

network/sendMail.ts
floatplane: no real reason I guess. might make testing easier but since I didn't write any tests... 🙈

PR #392, opened 2023-09-26
observations/filter: react to zone changing

### LINKED ISSUE #391: Switching Zone Using Drop-Down Doesn't Impact Observations Filter
When you have the avalanche forecast, weather forecast or conditions blog tabs open and switch the zone using the drop-down at the top of the screen, the tab re-renders with the appropriate data. However, when the observations tab is open and the zone is changed using the drop-down, the filters set is not updated to the new zone.

https://github.com/NWACus/avy/assets/7328370/0780a993-ec7d-4a93-87b4-7c99bc8a3496

stevekuznetsov:
Fixes #391

### DIFF
--- a/components/observations/ObservationsListView.tsx
+++ b/components/observations/ObservationsListView.tsx
@@ -6,12 +6,12 @@ import {FontAwesome, MaterialCommunityIcons} from '@expo/vector-icons';
 import {useNavigation} from '@react-navigation/native';
 import {colorFor} from 'components/AvalancheDangerPyramid';
 import {Card} from 'components/content/Card';
+import {NotFound, QueryState, incompleteQueryState} from 'components/content/QueryState';
 import {NetworkImage} from 'components/content/carousel/NetworkImage';
-import {incompleteQueryState, NotFound, QueryState} from 'components/content/QueryState';
-import {Center, Divider, HStack, View, VStack} from 'components/core';
+import {Center, Divider, HStack, VStack, View} from 'components/core';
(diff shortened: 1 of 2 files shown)


components/observations/ObservationsListView.tsx
```diff
@@ -49,17 +50,25 @@ interface ObservationFragmentWithPageIndexAndZoneAndSource extends ObservationFr
...
+    [initialFilterConfig],
+  );
   const [filterConfig, setFilterConfig] = React.useState<ObservationFilterConfig>(originalFilterConfig);
   const [filterModalVisible, setFilterModalVisible] = React.useState<boolean>(false);
   const mapResult = useMapLayer(center_id);
   const mapLayer = mapResult.data;
 
+  // when the initial filter inputs change, we should honor those and try to overlay the user's requests on top of the new base
+  React.useEffect(() => {
+    setFilterConfig(current => _.assign({}, current, originalFilterConfig));
```
floatplane, line 69: I don't think you need lodash here? Also the comment is not right I think...or at least the code is not doing what the comment says. It's laying the new base on top of the user's requests, not vice-versa.

```suggestion
    setFilterConfig(current => {
      ...current,
      ...originalFilterConfig
    });
```

floatplane (commented):
ptal @stevekuznetsov

stevekuznetsov:
@floatplane updated to use destructuring

floatplane:
looks like prettier wants a change, but LGTM!

floatplane (approved):

PR #397, opened 2023-09-26
Embed current git hash in the build

floatplane:
Make sure `EXPO_PUBLIC_GIT_REVISION` shows up in all build variants...or at least try to 😵


floatplane:
Verified that this works in development and in EAS update/preview builds.

PR #398, opened 2023-09-26
Making logging to file work again

### LINKED ISSUE #383: Logs not being saved to `logFilePath` anymore

floatplane:
~~Putting up as a draft but I think I should make at least one more change: not logging to file in release builds.~~

Logging to file in this implementation is stupidly expensive, because AFAICT you can only write whole files with the [Expo FileSystem API](https://docs.expo.dev/versions/latest/sdk/filesystem). I don't really want to keep all log lines in memory but maybe that would actually be better? At least then I wouldn't need to read the whole file before writing 🤷. Anyway, I try to minimize the badness by buffering up writes and flushing at most once per second. I'm guessing that our old logging package was doing something similar but I don't know.

The other change I made was to centralize creation of the logger object and make it a singleton.

Now, the good news: you can get badass views of these log files. Here's one showing filtering by log level, colored output and nice object formatting:

image

This PR will resolve #383, but emailing log files on iOS won't be supported until #382 is resolved.

### DIFF
--- a/logger.ts
+++ b/logger.ts
@@ -1,6 +1,7 @@
-import {Logger, createLogger, stdSerializers} from 'browser-bunyan';
+import {StreamOptions, createLogger, stdSerializers} from 'browser-bunyan';
 import Constants from 'expo-constants';
 import * as FileSystem from 'expo-file-system';
+import * as Updates from 'expo-updates';
 import {ConsoleFormattedStream} from 'logging/consoleFormattedStream';
 import {FileStream} from 'logging/fileStream';
 import {LogBox} from 'react-native';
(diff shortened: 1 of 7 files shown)


logger.ts
```diff
@@ -15,3 +19,26 @@ if (process.env.NODE_ENV !== 'test') {
 if (process.env.EXPO_PUBLIC_DISABLE_LOGBOX) {
   LogBox.ignoreAllLogs(true);
 }
+
+let theLogger: Logger | undefined = undefined;
```
stevekuznetsov, line 23: Why can't you do the `createLogger()` call right here on assignment and just export `theLogger` and refer to it?

stevekuznetsov:
Seems reasonable. The state of the ecosystem for logging client-side and doing things like flushing and log rotation is really bad. Not a blocker, but you might consider some naive log rotation code to ensure we don't eat up too much filesystem space on the user's device if they use the app all season.

floatplane:
> Seems reasonable. The state of the ecosystem for logging client-side and doing things like flushing and log rotation is really bad. Not a blocker, but you might consider some naive log rotation code to ensure we don't eat up too much filesystem space on the user's device if they use the app all season.

yeah, there's naive rotation here, where the file gets truncated to 0 bytes by the write:

https://github.com/NWACus/avy/blob/bdf37b96a0b357716af1ddb7103576e76c84353f/logging/fileStream.ts#L9-L13

might be a little subtle, I'll add a comment calling out what's happening there.

floatplane:
(but also, if we don't log to file in the actual production builds then we should be ok)

stevekuznetsov:
> yeah, there's naive rotation here, where the file gets truncated to 0 bytes by the write:

So the file will only ever contain 1s of logs? Seems like it might be of limited utility then.

> (but also, if we don't log to file in the actual production builds then we should be ok)

I understood the main reason we want to log to the file is to allow users to submit the logs with bug reports, right? Would we enable logging in the production builds at some point then?

floatplane:
> > yeah, there's naive rotation here, where the file gets truncated to 0 bytes by the write:
> 
> So the file will only ever contain 1s of logs? Seems like it might be of limited utility then.
No, that's in the constructor for the FileStream - truncating the file at startup, in other words.

> > (but also, if we don't log to file in the actual production builds then we should be ok)
> 
> I understood the main reason we want to log to the file is to allow users to submit the logs with bug reports, right? Would we enable logging in the production builds at some point then?

Mayyyyyybe? It seems like an option. I guess we should see if we think it's useful, and balance that with the potential issues around allowing unrestrained emails to us :)

We can probably get a long way by attaching files to our Sentry issues: https://docs.sentry.io/platforms/react-native/enriching-events/attachments/

stevekuznetsov:
What would be the use of logging to the file if it's not for users to send with bug reports? Could we just remove the feature?>

PR #408, opened 2023-09-29
weather_data: show a map with cards for stations

stevekuznetsov:
Avalanche centers other than NWAC do not have a short list of weather stations that their users frequent, nor do they aggregate multiple weather stations in one page (or co-locate them in physical reality). For these centers, we now show them a map with weather stations on it.

This is a first pass, updates should include:

 - a means by which users can filter the centers they're looking at on the map
 - support for webcams
 - perhaps a list/card view
 - a toggle to allow NWAC users to view this mode (and look at all the stations not in the curated list)


components/AvalancheForecastZonePolygon.tsx
```diff
@@ -63,14 +63,17 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
...
     onPress: (event: PolygonPressEvent) => {
-      onPress(zone);
-      // By calling stopPropagation, we prevent this event from getting passed to the MapView's onPress handler,
-      // which would then clear the selection
-      // https://github.com/react-native-maps/react-native-maps/issues/1132
-      event.stopPropagation();
+      if (onPress) {
+        onPress(zone);
+
+        // By calling stopPropagation, we prevent this event from getting passed to the MapView's onPress handler,
```
stevekuznetsov, line 72: Interestingly the behavior I noted was that if `tappable` was set, the map *would* get the event but the coordinates were `undefined`. Stopping propagation is one way to get rid of that, another would be to check that the location of the tap was not in a zone, in the map handler.

floatplane: It seems like the comment in the code describes what's going on correctly, are you saying something should change?

components/weather_data/WeatherStationDetail.tsx
```diff
@@ -225,7 +225,7 @@ export const WeatherStationDetail: React.FC<Props> = ({center_id, name, stations
     <VStack width="100%" height="100%" alignItems="stretch">
       <VStack width="100%" height="100%" p={16} space={8}>
         <HStack space={8} alignItems="center">
-          <BodyBlack>{name}</BodyBlack>
+          <BodyBlack style={{textTransform: 'capitalize'}}>{name.toLowerCase()}</BodyBlack>
```
floatplane, line 228: could just use lodash

```suggestion
          <BodyBlack>{_.capitalize(name)}</BodyBlack>
```

components/weather_data/WeatherStationList.tsx
```diff
@@ -54,7 +139,193 @@ const StationList: React.FunctionComponent<{
   if (center_id === 'NWAC') {
     return <NWACStationList center={center_id} mapLayer={mapLayer} stations={weatherStations} requestedTime={requestedTime} />;
   }
-  return <Unavailable />;
+
```
floatplane, line 142: Can this code get moved into a separate component, so that we have something more like `return <WeatherStationMap {...whateverPropsItNeeds} />`? 

This component has basically been responsible for fetching the data and instantiating the correct component to display it, or displaying a `QueryState` if it's not ready yet - that seems like a good separation of concerns. Now it's learning about the curvature of the earth which is quite a leap :)

components/weather_data/WeatherStationList.tsx
```diff
@@ -54,7 +139,193 @@ const StationList: React.FunctionComponent<{
...
+WeatherStationCard.displayName = 'WeatherStationCard';
+
+const colorsForSource = (source: WeatherStationSource) => {
+  switch (source) {
+    case WeatherStationSource.NWAC:
+      return {primary: '#0059C8', secondary: '#98CBFF'};
+    case WeatherStationSource.SNOTEL:
+      return {primary: '#006D23', secondary: '#9ED696'};
+    case WeatherStationSource.MESOWEST:
+      return {primary: '#EA983F', secondary: 'rgba(234, 152, 63, 0.2)'};
```
floatplane, line 324: I think these should be named colors in colors.ts (even if they're not final)

components/weather_data/WeatherStationList.tsx
```diff
@@ -37,15 +53,84 @@ export const WeatherStationList: React.FC<Props> = ({center_id, requestedTime})
   return <StationList center_id={center_id} token={metadata.widget_config.stations?.token} requestedTime={requestedTime} />;
 };
 
+const EARTH_RADIUS = 6378.1; // in kilometers
```
floatplane, line 56: I find it helpful to append the units to the variable name when there's potential for confusion
```suggestion
const EARTH_RADIUS_KM = 6378.1;
```
(you did this yourself in `radiusForExtent` 😄)

components/weather_data/WeatherStationList.tsx
```diff
@@ -54,7 +139,193 @@ const StationList: React.FunctionComponent<{
...
+        controller={controller}
+      />
+    </>
+  );
+};
+
+const WeatherStationPoint: React.FunctionComponent<{
+  station: WeatherStation;
+  selected: boolean;
+  radius: number;
```
floatplane, line 226: I would make this `radiusKm` for explicitness

components/weather_data/WeatherStationList.tsx
```diff
@@ -54,7 +139,193 @@ const StationList: React.FunctionComponent<{
...
+
+  const highlight = colorLookup('blue.100');
+  const colors = colorsForSource(station.properties.source);
+  const coordinate = {latitude: station.geometry.coordinates[1], longitude: station.geometry.coordinates[0]};
+
+  return (
+    <MapCircle
+      key={station.properties.stid}
+      center={coordinate}
+      radius={radius * 1000} // in meters
```
floatplane, line 240:
```suggestion
      radius={radiusKm * 1000}
```

components/weather_data/WeatherStationDetail.tsx
stevekuznetsov: We're doing this particular approach in quite a number of places today - I don't really understand why one would pick lodash over css here, or vice versa, and I'd rather keep it similar for now. Let's pull out an issue to follow up and fix them all if one approach is better than the other?

components/AvalancheForecastZonePolygon.tsx
stevekuznetsov: Not quite - in my call for the polygon, I don't set an `onPress` and this code does not fire, but the polygon somehow still eats the substantive part of the event.

components/weather_data/WeatherStationDetail.tsx
floatplane: it just looked weird to me that we're applying one string transform in JS and another in CSS, seems like it would be more succinct to have a a single JS call. but maybe I've done it before and started an unfortunate trend 🤷

stevekuznetsov: Yeah, makes sense. FWIW looking through the docs I think the two are not identical - `_.capitalize()` will not put into title case, just will capitalize the first word. No title case utility seems to exist in JavaScript, which is likely why we do this pattern?

components/AvalancheForecastZonePolygon.tsx
floatplane: Ah ok. I still think it's semi-surprising that the library propagates the events - it definitely doesn't seem like a sane default for them.

components/weather_data/WeatherStationDetail.tsx
floatplane: `/me shakes first at JS ecosystem`

components/observations/ObservationsListView.tsx
```diff
@@ -266,20 +266,7 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
-      return {primary: '#006D23', secondary: '#9ED696'};
-    case 'volunteer':
-    case 'public':
-    case 'other':
-      return {primary: '#EA983F', secondary: 'rgba(234, 152, 63, 0.2)'};
-  }
-  const invalid: never = partnerType;
-  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-  throw new Error(`Unknown partner type: ${invalid}`);
+  return {primary: colorLookup(`observer.${partnerType}.primary`).toString(), secondary: colorLookup(`observer.${partnerType}.secondary`).toString()};
```
floatplane, line 269: a minor note that we can sort out later: I suspect at some point we may want to ban dynamically generated strings for color IDs, so that we can statically typecheck that the ID is valid. I've recently fixed a bug where an invalid color string was getting passed and it fell back to black. 

just a note, no action required - we could pretty easily refactor this when the time comes.

components/weather_data/WeatherStationList.tsx
```diff
@@ -34,27 +35,15 @@ export const WeatherStationList: React.FC<Props> = ({center_id, requestedTime})
...
-    return <QueryState results={[mapLayerResult, weatherStationsResult]} />;
-  }
-
   if (center_id === 'NWAC') {
-    return <NWACStationList center={center_id} mapLayer={mapLayer} stations={weatherStations} requestedTime={requestedTime} />;
+    return <NWACStationList token={metadata.widget_config.stations?.token} requestedTime={requestedTime} />;
   }
-  return <Unavailable />;
+
+  return process.env.EXPO_PUBLIC_ENABLE_WEATHER_MAP ? (
```
floatplane, line 42: I've been adding new environment variables to `.env.sample` so that they're somewhat documented in one place. can you add `EXPO_PUBLIC_ENABLE_WEATHER_MAP` to that file?

https://github.com/NWACus/avy/blob/4b17f4b44ccc7d92cd3f50cdb524d0129cd1041a/.env.sample#L34-L38

components/weather_data/WeatherStationsDetail.tsx
```diff
@@ -0,0 +1,288 @@
...
+        </HStack>
+      </ScrollView>
+    </ScrollView>
+  );
+};
+
+const columnPadding = 3;
+const rowPadding = 2;
+
+export const Column: React.FunctionComponent<{borderRightWidth: number; name: string; units: string; elevation?: string; data: string[]}> = ({
```
floatplane, line 175: lol today I noticed this was still called `Row`...thank you for fixing

floatplane (approved):
One small request, I don't need to re-review!

Issue #313, opened 2023-08-03
Outcomes Not Styled Well


stevekuznetsov:
We have an `<Outcome/>` component and a number of permutations of these, but they are not styled well. When they take up a whole page, they are often not centered correctly. When they are added inline, they do not look like they flow with the page, etc.

floatplane:
@stevekuznetsov do you have some screenshots, or pointers on how to get the app into the states you're describing?

stevekuznetsov:
![File (5)](https://github.com/stevekuznetsov/avalanche-forecast/assets/7328370/0fdca26f-bc4a-427d-935c-e12e41496e01)


@floatplane here's an example of this. The colors look all goofy here.

charlotteguard:
@stevekuznetsov when you say the colors are goofy - do you mean the grey box? Maybe we have design weigh in on what this is supposed to look like?

stevekuznetsov:
Yes - it's just a weird grey box halfway into the screen.

floatplane:
fixed in #457

PR #413, opened 2023-10-02
Fix query state not being centered correctly

### LINKED ISSUE #347: Observations loading state not centered when loading from forecast tab
image

floatplane:
Fixes #347

### DIFF
--- a/components/observations/ObservationsListView.tsx
+++ b/components/observations/ObservationsListView.tsx
@@ -161,7 +161,11 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   );
 
   if (incompleteQueryState(nacObservationsResult, nwacObservationsResult, mapResult) || !observations || !mapLayer) {
-    return <QueryState results={[nacObservationsResult, nwacObservationsResult, mapResult]} />;
+    return (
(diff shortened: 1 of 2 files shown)


components/observations/ObservationsListView.tsx
```diff
@@ -161,7 +161,11 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   );
 
   if (incompleteQueryState(nacObservationsResult, nwacObservationsResult, mapResult) || !observations || !mapLayer) {
-    return <QueryState results={[nacObservationsResult, nwacObservationsResult, mapResult]} />;
+    return (
+      <Center width="100%" height="100%">
```
stevekuznetsov, line 165: Why do we need to center _this_ `QueryState` but not others? Is there a `Center` that some higher-level component expects? Are we `Center`ing all outputs from this ListView?

floatplane: Oh, I think there's no guarantee that others don't need something similar. I'm playing whack-a-mole because I don't have a great way to test these different states. When I tackle #313 I'll build something to force different outcomes more effectively.

Issue #416, opened 2023-10-03
Change `colorLookup()` to only allow input that's in the map.


stevekuznetsov:
a minor note that we can sort out later: I suspect at some point we may want to ban dynamically generated strings for color IDs, so that we can statically typecheck that the ID is valid. I've recently fixed a bug where an invalid color string was getting passed and it fell back to black. 

just a note, no action required - we could pretty easily refactor this when the time comes.

_Originally posted by @floatplane in https://github.com/NWACus/avy/pull/408#discussion_r1343234154_

stevekuznetsov:
cc @floatplane

stevekuznetsov:
Looks like the RN components actually expect `string|undefined` for colors, not `ColorValue` - changing the function signature to `colorLookup = (color: keyof typeof COLORS): ColorValue` leads to 

```
components/modals/AvalancheCenterSelectionModal.tsx:56:51 - error TS2322: Type 'ColorValue' is not assignable to type 'string | undefined'.

56                   <Title3Black textAlign="center" color={colorLookup('NWAC-dark')}>
                                                     ~~~~~
```

@floatplane do we just want to return `string` from `colorLookup`?

PR #417, opened 2023-10-03
colors.ts: ensure correct names are provided

stevekuznetsov:
Fixes https://github.com/NWACus/avy/issues/416

### DIFF
--- a/components/core/View.test.tsx
+++ b/components/core/View.test.tsx
@@ -3,13 +3,14 @@ import {render, screen} from '@testing-library/react-native';
 import React from 'react';
 
 import {View} from 'components/core/View';
+import {colorLookup} from 'theme';
 
 describe('View', () => {
   it("sets top-level properties on the wrapped View's style property", () => {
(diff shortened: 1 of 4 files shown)


components/core/View.test.tsx
```diff
@@ -25,39 +26,39 @@ describe('View', () => {
   });
```
stevekuznetsov, line 26: The test above this shows an edge-case for how we have these factored right now - colors not given as top-level attributes don't get put through `colorLookup`. Getting stuff like this ironed out is likely harder than just using Tamagui.

floatplane (approved):
This seems like a good start! How many things break if you change this line:

https://github.com/NWACus/avy/blob/37c0dcd8642dfa2e9eea8f6a47e1289965734a07/components/core/View.tsx#L9

to:
```js
readonly backgroundColor?: ColorName,
```

components/core/View.test.tsx
stevekuznetsov: Yes, I think the way they do Styled components does entirely make this error case impossible.

PR #422, opened 2023-10-04
forecast: remove pyramid, update triangle to new styling

stevekuznetsov:
File (18)


stevekuznetsov:
Noticing a lot of other stuff on the design ... holding on

floatplane (changes requested):
I know this is WIP but want to flag a couple of things to notice in your screenshots that don't look right:

- ~pyramid~ triangle stripes aren't lined up with the gray background stripes any longer
- elevation band labels are right on the edge of the component, they need to be inset by 4 px or so

stevekuznetsov:
| Closed | Open |
| - | - |
| File (20) |  File (19) |


@floatplane I have a boat load of more questions on the Figma chart but the big stuff should be done.

components/AvalancheDangerTable.tsx
```diff
@@ -50,26 +48,31 @@ export const AvalancheDangerTable: React.FunctionComponent<AvalancheDangerTableP
...
-            <HStack flex={1} justifyContent="space-between" key={index}>
-              <View my={4} px={1} justifyContent="center">
-                <Caption1>{(elevation_band_names[layer] ?? defaultElevationBands[layer]).replace('<br>', '\n')}</Caption1>
-              </View>
+            <HStack flex={1} justifyContent="space-between" flexDirection={elevation_band_names ? 'row' : 'row-reverse'} key={index} paddingHorizontal={4}>
+              {elevation_band_names && (
+                <View my={4} px={1} justifyContent="center">
+                  <Caption1>{(elevation_band_names[layer] ?? defaultElevationBands[layer]).replace('<br>', '\n')}</Caption1>
+                </View>
+              )}
```
floatplane, line 69: It looks like this label is meant to be in a white box with a small corner radius (probably 2px?) and a 1px border (color `#F3F5F6`, not sure if that's a predefined color or an improvisation from the design side 😄)

image

floatplane (approved):
One small tweak I can see but that's it!

components/AvalancheDangerTable.tsx
floatplane: Oh and I guess it's meant to be `<Caption1Semibold>` for the first line

stevekuznetsov:
| Closed | Open |
| - | - |
| File (23) | File (23) |

components/AvalancheDangerTable.tsx
stevekuznetsov: Hm - their designs have always had the white box in them and we've never actually made that happen. Maybe let's confirm with them in a follow-up if that's really the intent? I figured if it were the intent we would have heard about it in design review sometime in the last year :D 

(attached a design iteration from early last season) 

Screenshot from 2023-10-09 07-04-26

PR #424, opened 2023-10-06
Observation upload improvements

floatplane:
First, they actually work again! Had to include the new `obs_source` param.

Second: this PR massively overhauls how we track, schedule and retry the observation jobs. This comment describes what we're doing now:
https://github.com/NWACus/avy/blob/f399cfd4a4cfc53f3e4c037432575fc087185be6/components/observations/uploader/ObservationsUploader.ts#L35-L42

The `submitObservation` method is now very stripped down. It enqueues tasks into the `ObservationUploader`, which is then in turn responsible for scheduling calls to `uploadImage` and `uploadObservation` in sequence. Modules are split apart and much easier to read IMO.

Observation uploads should now be resilient in the face of network errors and even app restarts. There are also a bunch of tests in `ObservationUploader.test.ts` where I had to learn lots of tricks about how to test async code with Jest 😬.

The first enhancement I want to make is to detect online/offline status and pause/resume the task queue, but there's enough code here and enough improvement that I want to get this in first.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -35,7 +35,6 @@ import {PersistQueryClientProvider} from '@tanstack/react-query-persist-client';
 
 import {ClientContext, ClientProps, productionHosts, stagingHosts} from 'clientContext';
 import {ActionToast, ErrorToast, InfoToast, SuccessToast, WarningToast} from 'components/content/Toast';
-import {clearUploadCache} from 'components/observations/submitObservation';
 import {HomeTabScreen} from 'components/screens/HomeScreen';
 import {MenuStackScreen} from 'components/screens/MenuScreen';
 import {ObservationsTabScreen} from 'components/screens/ObservationsScreen';
@@ -217,8 +216,6 @@ const asyncStoragePersister = createAsyncStoragePersister({
   storage: AsyncStorage,
   key: QUERY_CACHE_ASYNC_STORAGE_KEY,
 });
-
-void clearUploadCache();
 
 const TabNavigator = createBottomTabNavigator<TabNavigatorParamList>();
 
--- a/App.tsx
+++ b/App.tsx
@@ -73,14 +73,37 @@ const encodeParams = (params: {[s: string]: string}) => {
     .map(kv => kv.map(encodeURIComponent).join('='))
     .join('&');
 };
+
+const filterBigStrings = (params: unknown): unknown => {
+  if (!params || (typeof params !== 'object' && typeof params !== 'string')) {
+    return params;
+  }
+
+  if (typeof params === 'string') {
+    return params.length > 100 ? params.substring(0, 100) + '...' : params;
+  }
+
+  return Object.fromEntries(
+    Object.entries(params).map(([k, v]) => {
+      if (typeof v === 'string' && v.length > 100) {
+        return [k, v.substring(0, 100) + '...'];
+      } else if (typeof v === 'object') {
+        return [k, filterBigStrings(v)];
+      } else {
+        return [k, v];
+      }
+    }),
+  );
+};
+
 const formatURI = (request: AxiosRequestConfig): string => {
   const method = request.method ?? 'GET';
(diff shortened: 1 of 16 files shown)


App.tsx
```diff
@@ -73,14 +72,43 @@ const encodeParams = (params: {[s: string]: string}) => {
     .map(kv => kv.map(encodeURIComponent).join('='))
     .join('&');
 };
+
+// Sometimes we're sending very large data params like base64-encoded images.
+// Let's not log those, eh?
```
floatplane, line 77: this was EXTREMELY spammy when you set the log level to DEBUG

components/observations/ObservationsFilterForm.tsx
```diff
@@ -372,11 +372,11 @@ export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterF
 
 export const matchesZone = (mapLayer: MapLayer, lat: number | null | undefined, long: number | null | undefined): string => {
   if (!lat || !long) {
-    return 'Not in any Forecast Zone';
+    return 'Unknown Zone';
```
floatplane, line 375: The original text was very log and wrapped
  awkwardly

when used in a title bar :D

components/observations/SimpleForm.tsx
```diff
@@ -28,15 +28,15 @@ import {LoggerContext, LoggerProps} from 'loggerContext';
...
+import {AvalancheCenterID, ImageMediaItem, InstabilityDistribution, MediaType} from 'types/nationalAvalancheCenter';
 
 export const SimpleForm: React.FC<{
   center_id: AvalancheCenterID;
   onClose?: () => void;
 }> = ({center_id, onClose}) => {
   const navigation = useNavigation<ObservationsStackNavigationProps>();
   const {logger} = React.useContext<LoggerProps>(LoggerContext);
-  const formContext = useForm({
+  const formContext = useForm<ObservationFormData>({
```
floatplane, line 39: because defaultValues returns a `Partial<ObservationFormData>`, formContext then decided that it was wrapping that type. this explicitly tells it that no, it's wrapping a full `ObservationFormData` and so that's what methods like `onSubmit` will receive.

components/observations/SimpleForm.tsx
```diff
@@ -63,10 +63,10 @@ export const SimpleForm: React.FC<{
 
   const {nationalAvalancheCenterHost} = React.useContext<ClientProps>(ClientContext);
 
-  const mutation = useMutation<Observation, AxiosError, Partial<ObservationFormData>>({
-    mutationFn: async (observationFormData: Partial<ObservationFormData>) => {
+  const mutation = useMutation<void, AxiosError, ObservationFormData>({
```
floatplane, line 66: `submitObservation` doesn't return an `Observation` now...but to be fair, it never did before either. remember this comment? https://github.com/NWACus/avy/blob/4339d9ba7d5443630a58285bcdb86a64784912c4/components/observations/submitObservation.ts#L143-L147

components/observations/uploader/uploadImage.ts
```diff
@@ -0,0 +1,76 @@
+import axios from 'axios';
```
floatplane, line 1: this code is copied over from `submitObservation`, but basically unchanged otherwise.

components/observations/uploader/uploadObservation.ts
```diff
@@ -0,0 +1,41 @@
+import axios from 'axios';
```
floatplane, line 1: this code is copied over from `submitObservation`, but basically unchanged otherwise.

floatplane (commented):
some notes

floatplane:
rebasing on latest main to get expo doctor check passing...<sigh>

floatplane:
@stevekuznetsov I know you're out for a couple of days. I'm going to merge this as it's a massive stability improvement on what we had before, but happy to process feedback async!

PR #433, opened 2023-10-09
Remove binary file accidentally committed in 35b3b4a

floatplane:


stevekuznetsov:
We should force-push if that's not going to break build pipelines, since a commit & revert will keep this in the `git` history and balloon our repo size forever.

floatplane:
It's 19MB so it's not exactly ballooning. It seems somewhat painful to get rid of a file, it can break open pull requests and invalidate stashes - https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository. I personally wouldn't sweat it but if you feel strongly than we can do that.

stevekuznetsov:
Total `.git` size before:

```
$ du -hs .git
27M	.git
```

Of that, 19MiB is the file you committed:

```
$ git cat-file -s c481154
19664823
```

So, yes - in terms of the size of the repo, about 70% of the disk space is that file right now. Ideally we just remove it from history :)

floatplane:
I will just nuke this from history

floatplane:
Actually, we need to merge this before removing the binary - otherwise the BFG tool won't remove the file because it's part of the current HEAD

Issue #426, opened 2023-10-06
Add online/offline detection to ObservationUploader


floatplane:
Follow-on work from #424 - we should pause the upload queue when offline, and resume it when online.

stevekuznetsov:
@floatplane I think we had this at some point in the past already - see [`useAppState()`](https://github.com/NWACus/avy/blob/main/hooks/useAppState.ts).  The other half would be a configuration option for [`tanstack-query`](https://tanstack.com/query/v4/docs/react/guides/network-mode).

floatplane:
I went with `NetInfo` since it seems to be [blessed by Expo](https://docs.expo.dev/versions/latest/sdk/netinfo/)

PR #437, opened 2023-10-11
Online/offline detection for ObservationUploader

floatplane:
With this change the ObservationUploader will put itself to sleep when the device is offline, and wake up when it's back online. Added a couple of tests as well.

Also updated the Jest config to make `silent` the default so that you don't see log lines (you can override by running `yarn jest --silent=false`).


floatplane:
This fixes #426

Issue #403, opened 2023-09-27
Obs custom filtering - can apply filters when start and end dates are not defined


yuliadub:
When I got to set a custom date range for obs, I can apply filters when start and end dates are not set - this results in a white screen. 

Similarly, I can apply just the start date and then when I open the date picker for end date, it will have todays date selected in blue, if I click it - that wont populate the end date but I can push "apply filters" and that also results in a white error screen. 

The apply filters should only be submittable when both start and end dates have been defined. 

One suggestion: default the value in the date picker for both start and end to be todays date.

A few screenshot references: 
![IMG_9242](https://github.com/NWACus/avy/assets/9418602/df54619c-06a6-4640-b58c-6eb8918ae02f)

![IMG_9243](https://github.com/NWACus/avy/assets/9418602/1bbc90d6-75c1-461f-83a2-ea5c9b2647b5)

Issue #439, opened 2023-10-11
Observation List Loading & Error States Disjoint


stevekuznetsov:
We have two spaces on the screen we currently show that we're loading things and what the outcome is. Ideally we just have one.

| Loading | Loaded |
| - | - |
| ![File (25)](https://github.com/NWACus/avy/assets/7328370/2718a3e9-f79c-4398-aa1d-4836f85c65e3) |  ![File (26)](https://github.com/NWACus/avy/assets/7328370/48449803-1733-4227-95f4-f93c6fa387da) |

cc @floatplane

Issue #335, opened 2023-09-18
Changing Date Range in Observations Does Not Cause Re-Load


stevekuznetsov:
What I did:

1. open app
2. go to observations, "view all"
3. enter a custom date range
4. let the app show me them
5. edit the "published after" date to an earlier time point
6. see no changes

cc @floatplane

stevekuznetsov:
Looks like I needed to reset the cache entirely to fix this, restarting the app did not help.

floatplane:
There's currently a hard-coded limit of 90 days back from the end date, despite whatever you pick in custom dates. The custom date picker doesn't enforce this so it's confusing. (The custom date picker doesn't enforce anything, really)

PR #440, opened 2023-10-12
Observations list view improvements

floatplane:
Quite a few changes to improve the list view experience

- implement the new spec for date filtering - you can view the current season, or a date range within it, but you can't view arbitrary dates
- that said, date filtering respects the time machine, so you can warp to last season for debugging purposes
- fix a bug with dropdown items not being visible in the filter form
- clean up rendering of loading states; should get rid of duplicate loading spinners
- the list view only has one observation query active - either NWAC or NAC, but we don't need to interleave both
- you can now scroll all the way to the beginning of the season

I'll add some comments inline

Fixes: #403, #335, #439

### DIFF
--- a/components/content/QueryState.tsx
+++ b/components/content/QueryState.tsx
@@ -129,8 +129,5 @@ export const Unavailable: React.FunctionComponent = () => {
 
 // incompleteQueryState checks to see if any of the queries are not yet complete - if so, render a <QueryState/>.
 export const incompleteQueryState = (...results: UseQueryResult[]): boolean => {
-  return results
-    .map(result => [result.isError, result.isLoading, isResultNotFound(result)])
-    .flat()
-    .reduce((accumulator, value) => accumulator || value);
+  return results.some(result => result.isError || result.isLoading);
 };
--- a/components/observations/ObservationsFilterForm.tsx
+++ b/components/observations/ObservationsFilterForm.tsx
@@ -58,8 +58,7 @@ const DATE_LABELS: Record<DateValue, string> = {
 
 const currentSeasonDates = (requestedTime: RequestedTime): {from: Date; to: Date} => {
   const endDate = requestedTimeToUTCDate(requestedTime);
-  // Months are zero-based, so September is 8
-  const startDate = new Date(endDate.getUTCMonth() >= 8 ? endDate.getUTCFullYear() : endDate.getUTCFullYear() - 1, 8, 1);
+  const startDate = startOfSeasonLocalDate(requestedTime);
   return {from: startDate, to: endDate};
 };
 
--- a/components/observations/ObservationsFilterForm.tsx
+++ b/components/observations/ObservationsFilterForm.tsx
@@ -21,7 +21,7 @@ import {FieldErrors, FormProvider, useForm} from 'react-hook-form';
 import {KeyboardAvoidingView, Platform, View as RNView, SafeAreaView, ScrollView, TouchableOpacity, findNodeHandle} from 'react-native';
 import {colorLookup} from 'theme';
 import {MapLayer, ObservationFragment, PartnerType} from 'types/nationalAvalancheCenter';
-import {RequestedTime, requestedTimeToUTCDate} from 'utils/date';
+import {RequestedTime, requestedTimeToUTCDate, startOfSeasonLocalDate} from 'utils/date';
 import {z} from 'zod';
 
 const avalancheInstabilitySchema = z.enum(['observed', 'triggered', 'caught']);
@@ -58,16 +58,8 @@ const DATE_LABELS: Record<DateValue, string> = {
 
 export const createDefaultFilterConfig = (requestedTime: RequestedTime, defaults: Partial<ObservationFilterConfig> | undefined): ObservationFilterConfig => {
   const endDate: Date = requestedTimeToUTCDate(requestedTime);
-  // Months are zero-base, so September is 8
+  // Months are zero-based, so September is 8
   const startDate = new Date(endDate.getUTCMonth() >= 8 ? endDate.getUTCFullYear() : endDate.getUTCFullYear() - 1, 8, 1);
-  console.log('createDefaultFilterConfig', {
-    dates: {
-      value: 'past_season',
-      from: startDate,
-      to: endDate,
-    },
-    ...defaults,
-  });
   return {
     dates: {
       value: 'past_season',
@@ -193,6 +185,7 @@ export const filtersForConfig = (
 };
 
 interface ObservationsFilterFormProps {
+  requestedTime: RequestedTime;
   mapLayer: MapLayer;
   initialFilterConfig: ObservationFilterConfig;
   currentFilterConfig: ObservationFilterConfig;
@@ -202,7 +195,14 @@ interface ObservationsFilterFormProps {
 
 const formFieldSpacing = 16;
 
-export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterFormProps> = ({mapLayer, initialFilterConfig, currentFilterConfig, setFilterConfig, setVisible}) => {
+export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterFormProps> = ({
+  requestedTime,
+  mapLayer,
+  initialFilterConfig,
+  currentFilterConfig,
+  setFilterConfig,
+  setVisible,
+}) => {
   const {logger} = React.useContext<LoggerProps>(LoggerContext);
   const formContext = useForm({
     defaultValues: initialFilterConfig,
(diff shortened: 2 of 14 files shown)


components/content/QueryState.tsx
```diff
@@ -129,8 +129,5 @@ export const Unavailable: React.FunctionComponent = () => {
 
 // incompleteQueryState checks to see if any of the queries are not yet complete - if so, render a <QueryState/>.
 export const incompleteQueryState = (...results: UseQueryResult[]): boolean => {
-  return results
-    .map(result => [result.isError, result.isLoading, isResultNotFound(result)])
-    .flat()
-    .reduce((accumulator, value) => accumulator || value);
+  return results.some(result => result.isError || result.isLoading);
```
floatplane, line 132: 1. the check for `result.isError` made the check for `isResultNotFound` redundant, so I removed the latter
2. `results.some` simplifies things

components/observations/ObservationsFilterForm.tsx
```diff
@@ -15,58 +15,68 @@ import {SelectField} from 'components/form/SelectField';
...
+  }),
+  observerType: z.nativeEnum(PartnerType).optional(),
+  instability: z
+    .object({
+      avalanches: avalancheInstabilitySchema.optional(),
+      cracking: z.boolean().optional(),
+      collapsing: z.boolean().optional(),
+    })
+    .optional(),
+});
```
floatplane, line 48: made this object mostly required - easy enough to enforce

components/observations/ObservationsFilterForm.tsx
```diff
@@ -240,6 +228,14 @@ export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterF
...
 
   const onSubmitHandler = (data: ObservationFilterConfig) => {
+    // When the user selects `current_season`, the date fields might be set to a previous custom range.
+    // Make sure they're overridden with the season dates.
+    if (data.dates.value === 'current_season') {
+      data.dates = {
+        value: 'current_season',
+        ...currentSeasonDates(requestedTime),
+      };
+    }
```
floatplane, line 238: this fixes the problem with clearing the custom date range

components/observations/ObservationsListView.tsx
```diff
@@ -47,45 +47,44 @@ interface ObservationFragmentWithPageIndexAndZoneAndSource extends ObservationFr
...
+  // From the 2023-24 season onward, we will only have NAC observations
+  const displayNWACObservations = center_id === 'NWAC' && endDate < new Date('2023-09-01');
+  const nacObservationsResult = useNACObservations(center_id, requestedTime, {
+    enabled: !displayNWACObservations,
+  });
+  const nwacObservationsResult = useNWACObservations(center_id, requestedTime, {
+    enabled: displayNWACObservations,
+  });
+
+  const observationsResult = displayNWACObservations ? nwacObservationsResult : nacObservationsResult;
```
floatplane, line 75: removed the code that deals with arrays of results - we only need to worry about one

components/observations/ObservationsListView.tsx
```diff
@@ -125,22 +127,12 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
-    const {startDate: filterStartDate} = datesForFilterConfig(filterConfig.dates, endDate);
-    return (
-      (nacObservationsResult.hasNextPage && (nacObservations.length === 0 || new Date(nacObservations[nacObservations.length - 1].createdAt) > filterStartDate)) ||
-      (nwacObservationsResult.hasNextPage && (nwacObservations.length === 0 || new Date(nwacObservations[nwacObservations.length - 1].createdAt) > filterStartDate))
-    );
-
-    return nacObservationsResult.hasNextPage || nwacObservationsResult.hasNextPage;
-  }, [nwacObservations, nwacObservationsResult, nacObservations, nacObservationsResult, filterConfig, endDate]);
+  const {from: filterStartDate} = filterConfig.dates;
+  const moreDataAvailable = observationsResult.hasNextPage && (observations.length === 0 || new Date(observations[observations.length - 1].createdAt) > filterStartDate);
```
floatplane, line 131: `moreDataAvailable` didn't need to be a function - it's render state.

components/observations/ObservationsListView.tsx
```diff
@@ -160,10 +152,10 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
     [],
   );
 
-  if (incompleteQueryState(nacObservationsResult, nwacObservationsResult, mapResult) || !observations || !mapLayer) {
+  if (incompleteQueryState(observationsResult)) {
```
floatplane, line 155: We don't need to block on loading `mapResult` - we can make things load even if it fails (but it's always in cache so it doesn't).

components/observations/ObservationsFilterForm.tsx
```diff
@@ -15,58 +15,68 @@ import {SelectField} from 'components/form/SelectField';
...
-  past_month: 'Last month',
-  past_season: 'Last forecast season',
+  current_season: 'Current season',
   custom: 'Custom range',
 };
 
+const currentSeasonDates = (requestedTime: RequestedTime): {from: Date; to: Date} => {
+  const endDate = requestedTimeToUTCDate(requestedTime);
+  // Months are zero-based, so September is 8
+  const startDate = new Date(endDate.getUTCMonth() >= 8 ? endDate.getUTCFullYear() : endDate.getUTCFullYear() - 1, 8, 1);
```
floatplane, line 62: ooh need to change this to `startOfSeasonLocalDate`

hooks/useNWACObservations.ts
```diff
@@ -7,47 +7,51 @@ import * as Sentry from 'sentry-expo';
 
 import {Logger} from 'browser-bunyan';
 import {ClientContext, ClientProps} from 'clientContext';
-import {add, formatDistanceToNowStrict} from 'date-fns';
+import {add, formatDistanceToNowStrict, sub} from 'date-fns';
```
floatplane, line 10: changes in this file look big, but it's just making the fetch strategy mirror the NAC strategy - fetch two week windows at a time.

using `limit` and `offset` over a 1 year period produces really slow performance, so fetching a series of smaller time windows works better

hooks/useNWACObservations.ts
```diff
@@ -85,46 +90,33 @@ export const prefetchNWACObservations = async (queryClient: QueryClient, nwacHos
 
 interface ObservationsQueryWithMeta {
   data: ObservationFragment[];
-  published_before: string;
-  published_after: string;
+  endDate: string;
+  startDate: string;
```
floatplane, line 94: minimize the number of places that we have different names for the same concept

hooks/useObservations.ts
```diff
@@ -392,24 +390,28 @@ export const ObservationDocument = `
...
-) =>
-  useQuery<ObservationsQuery, TError, TData>(
-    ['observations', variables],
-    useFetch<ObservationsQuery, ObservationsQueryVariables>(ObservationsDocument).bind(null, variables),
-    options,
-  );
 
-// No matter what span of time we're looking at, we will never query over more than 3 months of observations
-export const MAXIMUM_OBSERVATIONS_LOOKBACK_WINDOW: Duration = {months: -3};
+// These are unused - can they be deleted?
```
floatplane, line 402: @stevekuznetsov thoughts?

hooks/useObservations.ts
```diff
@@ -392,24 +390,28 @@ export const ObservationDocument = `
...
+//   options?: UseQueryOptions<ObservationsQuery, TError, TData>,
+// ) =>
+//   useQuery<ObservationsQuery, TError, TData>(
+//     ['observations', variables],
+//     useFetch<ObservationsQuery, ObservationsQueryVariables>(ObservationsDocument).bind(null, variables),
+//     options,
+//   );
+
+// No matter what span of time we're looking at, we will never query over more than 12 months of observations
+export const MAXIMUM_OBSERVATIONS_LOOKBACK_WINDOW: Duration = {months: -12};
```
stevekuznetsov, line 417: Isn't this a function of the current date? (`startOfSeasonLocalDate`)

components/observations/SimpleForm.tsx
```diff
@@ -62,6 +63,7 @@ export const SimpleForm: React.FC<{
   const scrollViewRef = useRef<ScrollView>(null);
 
   const {nationalAvalancheCenterHost} = React.useContext<ClientProps>(ClientContext);
+  const today = new Date();
```
stevekuznetsov, line 66: When using the date picker, shouldn't this be something we pass in?

components/observations/ObservationsListView.tsx
stevekuznetsov: We will just pop in the zone names on the cards and into the filter dropdown? Can that induce jitter in the layout?

hooks/useObservations.ts
stevekuznetsov: Autogenerated and can be removed I think

components/observations/ObservationsListView.tsx
floatplane: never happens in practice, since this data has already fetched by the map on startup - but I can revert if it makes you happier.

components/observations/SimpleForm.tsx
floatplane: For a new obs, I didn't think so, unless we want to let people pick dates that aren't today? Seems like a Charlotte question I guess 🤷.

hooks/useObservations.ts
floatplane: At this point, it's just a backstop to keep us from running away and pathologically fetching forever. We don't actually display values earlier than start of season/start of custom date window.

PR #441, opened 2023-10-12
weather_data: implement map with cards, list with filter

stevekuznetsov:
Still a work-in-progress, outstanding items:

 - getting the "list view" button to hover on top of the cards is not working perfectly - I have not found how to bind the pan response events to the animated value for the button, since we need it to be an offset from the value for the cards; I tried event listeners but they don't fire for native driver animations
 - loading the map is super slow. I'm not sure why - we memoize the points and the cards
 - need to thread through an event for the final layout of the map to get the starting diameter of the circles correct, right now they start off with the best guess we can do and don't update until the map pans
 - ~` WARN  initialScrollIndex "-1" is not valid (list has 4 items) ` pops up when filtering, unclear why~

https://github.com/NWACus/avy/assets/7328370/042bf852-961b-4450-88d3-c56d23d82f32

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -48,11 +48,6 @@ IOS_GOOGLE_MAPS_API_KEY=
 # EXPO_PUBLIC_ENABLE_CONDITIONS_BLOG=1
 
 #
-# Set this to enable the weather station map for non-NWAC centers.
-#
-# EXPO_PUBLIC_ENABLE_WEATHER_MAP=1
-
-#
 # Set this to enable one-click creation of fake observations.
 #
 # EXPO_PUBLIC_AUTOFILL_FAKE_OBSERVATION=1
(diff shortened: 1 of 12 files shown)


stevekuznetsov:
> getting the "list view" button to hover on top of the cards is not working perfectly - I have not found how to bind the pan response events to the animated value for the button, since we need it to be an offset from the value for the cards; I tried event listeners but they don't fire for native driver animations

Looks like getting that to work with `useNativeDriver` is not possible. Seems like the best approach will be to animate the cards and the button together in one higher-level `View`. The only tricky part here is that we need to determine the offsets dynamically, since we want the cards to be below the fold and the button to be above the fold when closed (right?). Right now we just have a static offset that is "big enough" to get all of the cards off the bottom of the screen, but the amount to which they are below is dependent on the size of the cards.

> WARN  initialScrollIndex "-1" is not valid (list has 4 items) pops up when filtering, unclear why

This happens on `main` in the observations filter as well.

floatplane:
this was superseded by #472

PR #444, opened 2023-10-13
Disable form controls after submitting

### LINKED ISSUE #428: Observation form should be disabled when submitting
New react-hook-form library allows setting a single `disabled` prop, maybe worth the update? But should be straightforward to just disable all the form fields.

floatplane:
Fixes #428

### DIFF
--- a/components/observations/SimpleForm.tsx
+++ b/components/observations/SimpleForm.tsx
@@ -162,6 +162,7 @@ export const SimpleForm: React.FC<{
   };
 
   const formFieldSpacing = 16;
+  const disableFormControls = mutation.isLoading || mutation.isSuccess;
 
   return (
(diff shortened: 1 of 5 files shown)


components/observations/SimpleForm.tsx
```diff
@@ -162,6 +162,7 @@ export const SimpleForm: React.FC<{
   };
 
   const formFieldSpacing = 16;
+  const disableFormControls = mutation.isLoading || mutation.isSuccess;
```
stevekuznetsov, line 165: Is the lifetime of `mutation` bound to the lifetime of the `SimpleForm`, so the UX ends up being that you close & re-open the form to submit two observations in sequence?

floatplane:
yes

PR #447, opened 2023-10-13
Opt-in unescaping of HTML entities in Text components

### LINKED ISSUE #250: Unexpected HTML in obs author names
A02E6161-13D5-496A-ABA6-8C782B42CE10

floatplane:
Fixes #250 

This is a conservative approach to the problem - only escaping entities when explicitly asked to. It might be ok to escape everywhere, but I'd want to spend some time looking at the app deeply before pushing a change like that out.

before | after
--- | ---
 |


stevekuznetsov:
Should we use this on elevation band names as well?

floatplane:
It only replaces entities like `&lt;`, but not tags like `<br>` - as far as
I know, anyway!

PR #451, opened 2023-10-17
pause/resume upload tasks. upload status api. refactoring.

### LINKED ISSUE #409: We could use a debug tool for looking at the pending observation uploads…it’s clearly too fragile right now

### LINKED ISSUE #443: Control over observation upload tasks: pause/resume/cancel

floatplane:
fixes #443 
fixes #409


components/observations/uploader/ObservationsUploader.ts
```diff
@@ -156,6 +265,65 @@ export class ObservationUploader {
...
+
+  resumeUpload(observationId: string, resumeImages = true) {
+    this.setTaskStatus(observationId, 'pending', resumeImages);
+    this.tryRunTaskQueue();
+  }
+
+  getUploadStatus(observationId: string): UploadStatus {
+    const task = this.findTaskById(observationId);
+    if (!task) {
+      throw new Error(`No task found with id ${observationId}`);
```
stevekuznetsov, line 312: Why do we prefer to throw here (and require the caller to try/catch) as opposed to returning an optional status and having the caller handle the missing case?

components/observations/uploader/ObservationsUploader.ts
```diff
@@ -156,6 +265,65 @@ export class ObservationUploader {
...
+    }
+    return {
+      networkStatus: this.offline ? 'offline' : 'online',
+      observation: {
+        status: task.status,
+        attemptCount: task.attemptCount,
+      },
+      images: this.findTasksByParentId(observationId).map(t => ({
+        status: t.status,
```
stevekuznetsov, line 321: What's the implication of doing a copy for the values here? Do we need to worry that the caller gets a snapshot of the status at call-time? I don't fully grok the runtime here but will the caller ever process stale data after `getUploadStatus` (can the routine be preempted?)

components/observations/uploader/ObservationsUploader.ts
```diff
@@ -172,6 +340,22 @@ export class ObservationUploader {
     this.checkInitialized();
     this.subscribers = this.subscribers.filter(subscriber => subscriber !== callback);
   }
+
+  getState(): UploaderState {
+    // The `map` call is because tsc needs a little help to grok the type of the filter expression
```
stevekuznetsov, line 345: I've always just used normal `for` loops since the type narrowing works for those (and there's no performance penalty AFAICT)

components/observations/uploader/ObservationsUploader.ts
```diff
@@ -140,13 +244,18 @@ export class ObservationUploader {
       this.logger.debug({entry}, `processed task queue entry successfully`);
       this.subscribers.forEach(subscriber => subscriber(entry, true, entry.attemptCount));
       this.taskQueue.shift(); // we're done with this task, so remove it from the queue
+      entry.status = 'success';
+      this.completedTasks.push(entry);
```
stevekuznetsov, line 248: What consumes the contents of this? Just the routine dumping everything to logs? Do we expect this to be pruned ever?

components/observations/uploader/ObservationsUploader.ts
floatplane: I would always prefer to not return optionals unless I have to. And at this point I don't have a use case that motivates returning an optional value here...that may change, in which case it's easy enough to adapt.

stevekuznetsov: Isn't this line (the throw itself) where an option is useful? For the same reason that `.find()` returns one? How can we be certain that all upstream callers knows this can throw and have correctly wrapped it in `try/catch`?

components/observations/uploader/ObservationsUploader.ts
floatplane: it's not saved to storage, so it's pruned from memory at app startup...I guess if someone is a very enthusiastic observation uploader then we might have a problem

components/observations/uploader/ObservationsUploader.ts
floatplane: 1. I am not sure I understand the root of this question. I think you're looking for something more in an answer beyond "it's a copy", but you will have to help me.
2. It's explicitly designed to be a snapshot, so...I guess I'm not worried. 
3. I'm afraid I don't understand this question, sorry - what "stale data", which "routine"? in general no, there is no preemption - but async code flows can mean that data is modified across `await` boundaries.

maybe you can give me an example of a problem you're worried we might hit?

components/observations/uploader/ObservationsUploader.ts
floatplane: we are in philosophically different places

stevekuznetsov: As long as we can be confident that the error is handled (and doesn't crash the app at runtime) it's fine. The nice thing about the option is that `tsc` can enforce we handle it correctly.

components/observations/uploader/ObservationsUploader.ts
stevekuznetsov: I guess I was considering the case where:

1. caller uses `getUploadStatus()`, gets a snapshot/pass-by-value copy
2. preemption occurs, other code runs and one or more tasks in the status change status
3. caller returns to execution and operates over stale data - perhaps using stale status to re-try something etc

Does that matter?

stevekuznetsov: (Or is my mental model of the runtime here just wrong, and this is not important?)

floatplane: the JS runtime doesn't allow preemption in the traditional sense so it's less of a concern. if your code is running, then it runs until you hit the end of your scope, as long as you don't do anything `async`.

that is to say, this is safe

```js
const status = getUploader().getUploadStatus(id);
doSomethingWithoutAwaiting();
// status can't be stale here
```

but this is less safe
```js
const status = getUploader().getUploadStatus(id);
await doSomethingAsync();
// status might be stale at this point, because you've yielded 
// and `processTaskQueue` will get a chance to run if it's waiting
```

I think the class of bugs you're describing are still possible, they're just less likely than in trad multithreaded environments.

stevekuznetsov: Makes sense, so it's on the caller to be careful then and this function does not have to be concerned.

stevekuznetsov: Thanks for the explanation!

PR #457, opened 2023-10-18
Fix NotFound outcome styling for ObservationListView, AvalancheForecastZoneMap, WeatherTab

floatplane:
before | after
--- | ---
image | image
image | image
image) | image

### DIFF
--- a/components/forecast/WeatherTab.tsx
+++ b/components/forecast/WeatherTab.tsx
@@ -182,14 +182,36 @@ export const WeatherTab: React.FC<WeatherTabProps> = ({zone, center_id, requeste
 
   if (metadata?.widget_config.stations?.token) {
     if (incompleteQueryState(weatherStationsResult) || !weatherStations) {
-      return <QueryState results={[weatherStationsResult]} />;
+      return (
+        <QueryState
+          results={[weatherStationsResult]}
+          terminal
+          customMessage={{
+            notFound: () => ({
+              headline: 'Missing forecast',
+              body: 'There may not be a forecast available for today.',
+            }),
+          }}
+        />
+      );
     } else {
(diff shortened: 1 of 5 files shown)


components/forecast/WeatherTab.tsx
```diff
@@ -182,14 +182,36 @@ export const WeatherTab: React.FC<WeatherTabProps> = ({zone, center_id, requeste
...
     if (incompleteQueryState(weatherStationsResult) || !weatherStations) {
-      return <QueryState results={[weatherStationsResult]} />;
+      return (
+        <QueryState
+          results={[weatherStationsResult]}
+          terminal
+          customMessage={{
+            notFound: () => ({
+              headline: 'Missing forecast',
+              body: 'There may not be a forecast available for today.',
```
stevekuznetsov, line 192: This is keyed off of the `weatherStationsResult` - not sure the error message is correct.

floatplane: It has the same effect for end users - either way, they don't get to see anything - so I'm not sure I see the value in a different message. But if you have a suggested phrasing then fire away!

stevekuznetsov: I guess that's a UX question - if we can load the weather forecast but not the list of weather stations, should we full-screen an error message and show nothing? Relates to a card we have in Asana that has not seen much movement but probably deserves some thought from the group.

floatplane: that is a much bigger question. personally think it's better to show an error, than to have a UX that almost always works as expected, but occasionally behaves differently or randomly hides functionality. 

I'm assuming this is a rare error though - if it's not then my take would change.

Issue #429, opened 2023-10-06
"Cancel submission" button doesn't work anymore


floatplane:
Changes in #424 broke this, for better or worse

Issue #462, opened 2023-10-19
Newly uploaded observations don't show up in list, even after refresh


floatplane:
The only way to see them is to restart the app. Have played a bit with `staleTime` and `cacheTime` to no avail.

floatplane:
fixed in #463

PR #463, opened 2023-10-19
Observation upload UX improvements

floatplane:
- show pending observations in the obs list. this UX is still TBD
- when submission is complete, list updates to show the new obs
- show toasts as designed on upload form

Fixes: #429 by not making the cancel button a cancel button anymore. We still need to decide what the UX for cancellation looks like - is it a cancel button on the pending obs in the obs list?

~Still TBD:~
- ~listview doesn't refresh correctly once obs is submitted, so new obs don't show up until restart (tracking in issue #462)~
this is now fixed - the issue is that the date filters were set up to end at the exact time that the list view was loaded, as opposed to midnight on the custom date range or at the end of the season in the default case.

https://github.com/NWACus/avy/assets/101196/05007ef0-c974-471e-95c5-9957e4f28b46

simulator_screenshot_CDFCA733-A138-4FA0-82B8-0FBB8545637F

### DIFF
--- a/components/observations/uploader/usePendingObservations.ts
+++ b/components/observations/uploader/usePendingObservations.ts
@@ -1,17 +1,37 @@
-import {ObservationFragmentWithStatus, getUploader} from 'components/observations/uploader/ObservationsUploader';
+import {useQueryClient} from '@tanstack/react-query';
+import {ObservationFragmentWithStatus, UploaderState, getUploader} from 'components/observations/uploader/ObservationsUploader';
 import {useCallback, useEffect, useState} from 'react';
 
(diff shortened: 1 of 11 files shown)


components/observations/uploader/usePendingObservations.ts
```diff
@@ -0,0 +1,37 @@
...
+          type: 'all',
+          exact: false,
+          queryKey: ['nac-observations'],
+          refetchPage: (_page, index, _allPages) => {
+            return index === 0;
+          },
+          // refetchPage: (_page, index, _allPages) => {
+          //   return index === 0;
+          // },
+        });
```
floatplane, line 24: this is not working as I'd hoped yet

PR #471, opened 2023-10-23
Submit cleanup

floatplane:
1. remove Under Construction screen from weather tab
2. hide weather tab for non-NWAC in preview & release builds
3. remove beta tag from splash screen

new splash screen | SNFAC home view | NWAC home view
--- | --- | ---
image | image | image


floatplane:
confirmed with @charlotteguard in Slack that we'll just hide the weather tab for now and finish off that feature this week

PR #477, opened 2023-10-25
Set minimum Android version to 8 / Android SDK 26

### LINKED ISSUE #475: Android 6 issues in Sentry
assuming there's agreement on bumping our minimum version to Android 8, we can set the `minSdkVersion` to `26` ([source](https://en.wikipedia.org/wiki/Android_Oreo)) via the [`expo-build-properties` plugin](https://docs.expo.dev/versions/latest/sdk/build-properties/#pluginconfigtypeandroid).

if there's no agreement then 🤷 I guess someone has to go find an old Nexus phone and get busy

floatplane:
Resolves #475

### DIFF
--- a/app.json
+++ b/app.json
@@ -53,6 +53,14 @@
       ]
     },
     "plugins": [
+      [
+        "expo-build-properties",
+        {
+          "android": {
+            "minSdkVersion": 26
+          }
+        }
+      ],
       "sentry-expo",
       [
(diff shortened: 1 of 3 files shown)


floatplane:
Did an adhoc build and installed it on my Android device, so I think this works as intended. I don't have an old device to test what happens.

PR #500, opened 2023-10-27
Implement "down for maintenance" checker

floatplane:
- add PostHog integration
- check the `down-for-maintenance` feature flag...
  - ...at startup
  - ...when coming back to the foreground...
    - ...but not more frequently than every 30 minutes
- display a modal when the flag is set
- allow retry button in modal to check the state of the feature flags again
- made `useAppState` hook behavior more idiomatic, based on code from https://reactnative.dev/docs/appstate

the movie is too big to include here but you can see it in https://nwactechcommittee.slack.com/archives/C03BMJDAT9S/p1698441229433689


components/KillSwitchMonitor.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const tryReloadFeatureFlags = (posthog: PostHog) => {
+  logger.debug('fetching feature flags');
+  void posthog.reloadFeatureFlagsAsync();
+};
+
+// In release and preview mode, When returning to foreground, don't fetch feature flags more frequently than every 30 minutes
+// In development mode, refresh every time we return from foreground
+const FEATURE_FLAG_REFRESH_INTERVAL_MS = Updates.channel ? 30 * 60 * 1000 : 0;
```
stevekuznetsov, line 26: Is `Updates.channel` set on "vanilla" versions of the app - published once to the app stores and not sent to the user over OTA/EAS build?

stevekuznetsov (commented):
Looks good! One question

components/KillSwitchMonitor.tsx
floatplane: yes, channel is defined in `eas.json` for each build flavor. that value defines which channel the vanilla build would look at for its updates.

PR #529, opened 2023-11-02
Implement the 10 second startup delay when loading updates

floatplane:
This stuff is hard to test without an actual build and update channel, so I'm going to merge and party on the `preview` channel today.

Changes:
- Render a loading screen that seamlessly replaces the splash screen, includes a spinner
- Render that screen while we check for updates
- Add the 10 second delay when checking for and downloading updates at startup

Not implemented yet:
- Watching for updates while the app is running and prompting for restart

Movie showing the splash screen with spinner:

https://github.com/NWACus/avy/assets/101196/f21581cc-741d-474d-940e-540b8be6eb5b


floatplane:
This implements flow 2 from the update flowcharts
image

floatplane:
I've pushed at least 10 different updates that include this PR today, and it's updated every time, so I think it's working as intended

PR #537, opened 2023-11-06
Performance improvements

floatplane:
This work is still in progress but this PR will be massive if I don't break it up. Hammering down various sources of inefficiency. Most of our issues seem to be caused by excessive re-rendering, and most of our re-rendering seems to be down to using [fat arrow functions in props](https://stackoverflow.com/questions/36677733/why-shouldnt-jsx-props-use-arrow-functions-or-bind). Ultimately I will turn on an ESLint to check for that error, but I have to fix about 200 call sites first 😵

### DIFF
--- a/hooks/useAvalancheCenterMetadata.ts
+++ b/hooks/useAvalancheCenterMetadata.ts
@@ -13,7 +13,7 @@ import {LoggerContext, LoggerProps} from 'loggerContext';
 import {AvalancheCenter, AvalancheCenterID, avalancheCenterSchema} from 'types/nationalAvalancheCenter';
 import {ZodError} from 'zod';
 
-export const useAvalancheCenterMetadata = (center_id: AvalancheCenterID): UseQueryResult<AvalancheCenter, AxiosError | ZodError> => {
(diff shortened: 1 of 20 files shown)


hooks/useAvalancheCenterMetadata.ts
```diff
@@ -77,7 +77,9 @@ const fetchAvalancheCenterMetadata = async (nationalAvalancheCenterHost: string,
         url,
       },
     });
-    throw parseResult.error;
+    // If we throw here, we'll just trigger retries forever based on react-query's retry policy.
+    // Since we know we're getting an unparseable result, we should stop asking.
+    return null;
```
stevekuznetsov, line 82: Doing this in one hook - but no others - is going to surprise someone who is not expecting that. Perhaps update the schema if we're simply out-of-date or use a deny-list of nonconforming centers in the list?

floatplane: The right solution is not immediately obvious to me. Centers are conforming one day and non-conforming the next (MSAC is the current offender) so I think maintaining a deny list in code will be toilsome. A cloud-based approach might work but I don't have time for bigger solutions this week...

I observed infinite retry in the logs, otherwise I wouldn't have bothered making this change.

Turning this into an issue so it can get resolved more solidly.

PR #539, opened 2023-11-06
Fixing up more fat arrow functions in render

floatplane:
Mostly does what it says on the tin. Also:
- added warning to ActionList to watch for this case
- tweaked the `useToggle` interface to make it easier to use


floatplane:
tested on Android and iOS devices, seems good

PR #551, opened 2023-11-07
Set up Mixpanel integration, plumb through keys to all build steps

### LINKED ISSUE #282: Set up mixpanel

floatplane:
Includes a fake mixpanel object that only logs if mixpanel is disabled. Resolves #282


floatplane:
can't really test the workflow file change without merging :(. it runs the workflow file from `main`. 

https://github.com/NWACus/avy/actions/runs/6789751199/workflow

PR #552, opened 2023-11-07
Add screen tracking and user ID to Mixpanel

floatplane:
- Create an analytics ID for each install and store it in preferences
- Track screen visits


floatplane:
not ready yet: this implementation will trigger multiple events because the props object changes every time.

Issue #557, opened 2023-11-09
`releaseChannel` on PostHog is always `default`


floatplane:
This makes it impossible to write rules that target test builds vs. release builds. Culprit:

https://github.com/NWACus/avy/blob/8b12cf9fa2146df56842febe2b004b705fb5a753/components/KillSwitchMonitor.tsx#L44

Should be `Updates.channel || 'development'`

stevekuznetsov:
Yep, looks like `releaseChannel` is the wrong value - `channel` is for updates.

stevekuznetsov:
Might be good to name the PostHog value `channel` as well for clarity

PR #558, opened 2023-11-09
Fix channel sent to posthog

floatplane:
Tested in development to make sure the correct value showed up there:

image


floatplane:
@stevekuznetsov good call on the rename. but note that `releaseChannel` still shows up because those values are attached to the unique users that posthog infers.

image

probably possible to clean those up, just something to watch out for

floatplane:
fixes #557

PR #574, opened 2023-11-16
App.tsx: limit maximum font size

stevekuznetsov:


stevekuznetsov:
Quick summary of what happened here:

I tried to use the `defaultProps` approach to set the maximum scaling factor, but doing that made the app hang on the splash-screen forever for some reason. I couldn't root-cause that issue, and `tsc` really hates that approach, so I tried to piece-meal font size by adding it to our `Text` wrapper and the HTML rendering elements. This works fine, but it doesn't look like `react-native-navigation` yet supports this prop (and may never), so if we take the piece-meal approach not only do we need to be vigilant in the future when we add new text elements, but we also need to either find a way to get the tab navigators and what-not to honor our scaling factor, or call it good that most of the app doesn't over-scale.

yuliadub:
do we know if the scaling issue has come up much since the last time it was mentioned? trying to gauge where this would be on our priority list of things to fix.

stevekuznetsov:
Anecdotally I saw some media posted by the AC to promote the app, but it was done on a user's phone with large font and a small screen and I was really not impressed with how poorly we handled it. Can't say it's a big issue but certainly still there.

yuliadub:
gonna close this as its been open for a while. @stevekuznetsov feel free to reopen is you disagree

PR #575, opened 2023-11-17
Version Policy Updates

stevekuznetsov:
app.json: set runtime version policy

Our app contains dependencies outside of the Expo SDK which ship native
code. Therefore, we will be diligent to bump the runtime version of the
app whenever one of those dependencies changes. We futhermore require
that updates we ship only target the version of the app that they were
built on top of, to ensure that users on older native versions do not
receive OTA updates that are incompatible with their native layer.

---

.github: turn on OTA update publishing for release branch

We turned this off when we had to update code in the release branch that
changed native code but did not yet have a mechanism to conditionally
build OTA bundles only when no native code changed.

---

### DIFF
--- a/.github/workflows/publish-build.yaml
+++ b/.github/workflows/publish-build.yaml
@@ -1,10 +1,6 @@
 name: publish-build
 on:
   workflow_dispatch: {}
-  push:
-    branches:
-      - preview
-      - release
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


yuliadub (approved):
Based on expo docs, as long as appVersion actually does what it does we won’t need multiple expo channels which is exciting

PR #579, opened 2023-11-17
observations: require phone number

stevekuznetsov:
observations: make it clear that email is not shown to the public

---

Observations: require a phone number on input

---


stevekuznetsov:
We need a decision from @charlotteguard about whether e-mail should be optional or required to move forward here.

charlotteguard:
Email should be required.

charlotteguard:
But phone number should not be required!

stevekuznetsov:
Reading comprehension 0/10. Would you like us to clarify that phone number is optional in the parenthetical? `"(optional, never shared with the public)"`

PR #581, opened 2023-11-21
.github: bring update publishing in line with builds

stevekuznetsov:
We don't need to run CI before publishing OTA updates any more than we need to run it before publishing binary builds - only passing PRs can merge into the branch in the first place, so this is not useful.

### DIFF
--- a/.github/workflows/publish-update.yaml
+++ b/.github/workflows/publish-update.yaml
@@ -7,11 +7,7 @@ defaults:
     shell: bash
 
 jobs:
-  ci:
-    uses: ./.github/workflows/ci.yaml # use the callable ci job to run CI checks
-
   publish-update-bundle:
-    needs: [ci]
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


floatplane:
@stevekuznetsov I think this accidentally opens up a gap in coverage. Isn't it possible for two pull requests to enter the merge queue, each of them passing in CI but the combination of them not passing? Rare, but definitely not unheard of.

Not saying this PR should get reverted, but rather that we should be running CI checks on `main` after every merge and I don't think we currently are.

floatplane:
Never mind that comment, now I see how this is implemented

stevekuznetsov:
@floatplane it's a good call-out but I think the merge queue is grouping change-sets together [link](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#how-merge-queues-work) so that particular failure mode is not present here

PR #585, opened 2023-11-26
Support `specific` instability distribution

floatplane:
This was causing parse errors, which in turn led the obs list to not load.

This is a NAC form option that we never supported:

image


floatplane:
@stevekuznetsov not sure what you want to do with the expo doctor failure. this PR is an ideal candidate for OTA, but upgrading the listed packages almost surely requires a new app store build.

stevekuznetsov:
@floatplane I've been ignoring the Expo Doctor failure for now - we can OTA these and fix the doctor bits when we're ready to push a new binary build.

PR #595, opened 2023-11-28
Sticky headers on the weather table

floatplane:
This PR changes the weather table - not massively, but noticeably. The new behavior is **off by default** but can be enabled in two ways:
1. set `EXPO_PUBLIC_NEW_WEATHER_TABLE=true` in your `.env` file while developing
2. or flip the `new-weather-table` feature flag in posthog

The changes are:
1. row and column headers are sticky
2. ~column widths are uniform. This might not be a long-term requirement, but it's definitely gonna be more challenging to make the column header widths line up with the data widths when they're no longer lined up in a single `VStack`.~ now applying a rough heuristic size based on label length

Marking as draft for now, I want to do a little more side-by-side testing. Feel free to give feedback/ask questions though.

Still TODO:
- [x] support font scaling
- [x] use same approach for NAC weather data table

before | after 
--- | ---
Simulator Screen Recording - iPhone 14 Pro Max - 2023-11-28 at 12 13 15 | Simulator Screen Recording - iPhone 14 Pro Max - 2023-11-28 at 12 12 39

### DIFF
--- a/components/weather_data/WeatherStationDetail.tsx
+++ b/components/weather_data/WeatherStationDetail.tsx
@@ -1,18 +1,21 @@
-import React, {useState} from 'react';
+import React, {useCallback, useMemo, useState} from 'react';
 import {ScrollView} from 'react-native';
 
 import * as Sentry from 'sentry-expo';
 
 import {ButtonBar} from 'components/content/ButtonBar';
+import {DataGrid} from 'components/content/DataGrid';
 import {InfoTooltip} from 'components/content/InfoTooltip';
 import {incompleteQueryState, NotFound, QueryState} from 'components/content/QueryState';
-import {Divider, HStack, View, VStack} from 'components/core';
-import {BodyBlack, bodySize} from 'components/text';
+import {Center, Divider, HStack, View, VStack} from 'components/core';
+import {BodyBlack, bodySize, BodyXSm, BodyXSmBlack} from 'components/text';
 import {Column, formatDateTime} from 'components/weather_data/WeatherStationsDetail';
 import {compareDesc} from 'date-fns';
 import {useAvalancheCenterMetadata} from 'hooks/useAvalancheCenterMetadata';
 import {useWeatherStationTimeseries} from 'hooks/useWeatherStationTimeseries';
-import {AvalancheCenterID, StationNote, Variable, WeatherStationSource} from 'types/nationalAvalancheCenter';
+import {useFeatureFlag} from 'posthog-react-native';
+import {colorLookup} from 'theme';
+import {AvalancheCenterID, StationNote, Variable, WeatherStationSource, WeatherStationTimeseries} from 'types/nationalAvalancheCenter';
 import {NotFoundError} from 'types/requests';
 import {formatInTimeZone, parseRequestedTimeString, RequestedTimeString, utcDateToLocalDateString} from 'utils/date';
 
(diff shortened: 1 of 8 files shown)


floatplane:
brought the columns closer together, added support for font scaling

Simulator Screen Recording - iPhone 14 Pro Max - 2023-11-28 at 15 19 54

floatplane:
Fixed awkward label display for Solar Radiation column

image

floatplane:
I've gone through every NWAC weather station and haven't seen any regressions when using the new table (data is where it's supposed to be, no awkward text wrapping in column headers).

stevekuznetsov:
FWIW on SNFAC most of the MESOWEST and SNOTEL are identical sensors to each other so you can get a good indication of how this works without sifting through all 300

floatplane:
Oh I am definitely not going through all 300 😂 Checked a couple of each and they seem fine!

MESOWEST | SNOTEL
--- | ---
Simulator Screen Recording - iPhone 14 Pro Max - 2023-11-29 at 09 13 24 | Simulator Screen Recording - iPhone 14 Pro Max - 2023-11-29 at 09 16 40

floatplane:
I think this should be safe to merge, and should be a no-op from a user perspective until the feature flag is changed.

components/weather_data/WeatherStationDetail.tsx
```diff
@@ -269,6 +259,126 @@ export const orderStationVariables = (stationVariables: Variable[], timezone: st
...
+        flex={1}
+        alignItems="center"
+        justifyContent="flex-start"
+        py={rowPadding}
+        px={columnPadding}
+        bg="blue2"
+        borderBottomWidth={1}
+        borderColor={colorLookup('text.tertiary')}>
+        <BodyXSmBlack color="white">Time</BodyXSmBlack>
+        <BodyXSmBlack color="white">PST</BodyXSmBlack>
```
stevekuznetsov, line 332: The previous behavior here was to show the timezone of the user and to localize times - which is also what the website does outside of the old-school NWAC "NOW" views, and what the other centers get on their weather station maps - can we retain that behavior?

stevekuznetsov (commented):
Looks good! Only one comment

stevekuznetsov:
You might need to rebase on #594 for the falsy display update? Your videos show `Pcp1` column with `'-'` values instead of `0` still

components/weather_data/WeatherStationDetail.tsx
floatplane: good callout! see 5801a30

components/weather_data/WeatherStationsDetail.tsx
```diff
@@ -164,7 +269,7 @@ const TimeSeriesTable: React.FC<{timeSeries: WeatherStationTimeseries}> = ({time
               name={shortFieldMap[field]}
               units={shortUnits(timeSeries.UNITS[field])}
               elevation={elevation ? elevation.toString() : ''}
-              data={times.map(time => (time in dataByTime ? (dataByTime[time] === null ? '-' : String(dataByTime[time])) : '-'))}
+              data={times.map(time => (dataByTime[time] !== null ? String(dataByTime[time]) : '-'))}
```
floatplane, line 272: @stevekuznetsov in the process of rebasing on #594 it seemed to me that this could be simplified quite a bit - double ternary is hard to parse. Give a shout if this looks wrong to you!

stevekuznetsov:
Sweet! Let's do it

PR #599, opened 2023-11-29
Fix problem with obs date > submitted date

### LINKED ISSUE #584: Observation Submission From App Sets the Starting & Creation Date Incorrectly
image

floatplane:
I tested this fix by manually setting my computer clock to 7 PM MST, which is past midnight in UTC.

without the fix | with the fix
--- | ---
image | image

Fixes #584

### DIFF
--- a/components/observations/uploader/uploadObservation.ts
+++ b/components/observations/uploader/uploadObservation.ts
@@ -8,12 +8,17 @@ import {apiDateString} from 'utils/date';
 export async function uploadObservation(id: string, data: ObservationTaskData): Promise<Observation> {
   const {formData, extraData} = data;
   const {url, ...params} = extraData;
+  // The user only picks a date, not a time. The time defaults to whatever the current time is in the user's timezone.
+  // If the time is late enough, it might be in the next day when converted to UTC. By setting the time to noon in the user's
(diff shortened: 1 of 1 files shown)


components/observations/uploader/uploadObservation.ts
```diff
@@ -8,12 +8,17 @@ import {apiDateString} from 'utils/date';
...
+  // local timezone, we mitigate this problem.
+  const adjustedDate = new Date(formData.start_date);
+  adjustedDate.setHours(12, 0, 0, 0);
   const payload: Partial<Observation> = {
     ...formData,
     ...params,
     obs_source: 'public',
     // Date has to be a plain-old YYYY-MM-DD string
-    start_date: apiDateString(formData.start_date),
+    start_date: apiDateString(adjustedDate),
```
stevekuznetsov, line 21: Could we simply do the formatting without converting to UTC? Might be simpler and in case we ever get observations from the Middle East, would still work ;)

floatplane: yes, that's cleaner - I like it

floatplane:
ptal @stevekuznetsov

Issue #452, opened 2023-10-17
Eliminate use of @ts-ignore wherever possible


floatplane:
Biggest challenge seems to be in View.tsx. Moving this over from Asana.

stevekuznetsov:
This probably relates to #301 or at the least informs whether or not we think that's worth investment.

PR #600, opened 2023-12-01
Fix Typescript issues in View property setting

floatplane:
- Remove warning suppression from View.tsx
- Remove `bgColor` alias - we only need `backgroundColor` or `bg`
- React Native style prop can be an array, so I leverage that - now the original style object is passed through unchanged, and an additional style object is created with any props passed as top-level props

There's a minor downside in that all the style props have to get written twice in View.tsx (so not DRY), but I haven't had to add a new one since I created that file way back when, so I don't think that's a major issue.

#452 remains open because of unrelated other instances of `ts-ignore`

### DIFF
--- a/components/core/View.tsx
+++ b/components/core/View.tsx
@@ -249,12 +249,14 @@ const validateProp = (prop: ViewStyleProp, value: unknown): void => {
   }
 };
 
-// Given an object and a set of keys, return an array of two objects:
-// 1. An object containing only the keys in the keySet
-// 2. An object containing only the keys not in the keySet
-function split<T extends object, K extends keyof T>(obj: T, keySet: Record<K, string>): [Pick<T, K>, Pick<T, Exclude<keyof T, K>>] {
+/**
+ * Given an object and a set of keys, return an array of two objects:
+ * 1. An object containing only the keys in the keySet
+ * 2. An object containing only the keys not in the keySet
+ */
+function split<T extends object, K extends keyof T>(obj: T, keySet: Record<K, string>): [Pick<T, K>, Omit<T, K>] {
   const pick = {} as Pick<T, K>;
-  const unpick = {} as Pick<T, Exclude<keyof T, K>>;
+  const unpick = {} as Omit<T, K>;
   (Object.keys(obj) as (keyof T)[]).forEach(k => {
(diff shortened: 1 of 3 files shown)


components/core/View.tsx
```diff
@@ -152,31 +249,44 @@ const validateProp = (prop: ViewStyleProp, value: unknown): void => {
...
-      if (style) {
-        // TODO(bsharon): this needs to be cleaned up to be accordant with typescript ...
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        style[prop] = value;
-      }
+// Given an object and a set of keys, return an array of two objects:
+// 1. An object containing only the keys in the keySet
+// 2. An object containing only the keys not in the keySet
+function split<T extends object, K extends keyof T>(obj: T, keySet: Record<K, string>): [Pick<T, K>, Pick<T, Exclude<keyof T, K>>] {
```
stevekuznetsov, line 255:
```suggestion
function split<T extends object, K extends keyof T>(obj: T, keySet: Record<K, string>): [Pick<T, K>, Omit<T, K>] {
```

components/core/View.tsx
```diff
@@ -152,31 +249,44 @@ const validateProp = (prop: ViewStyleProp, value: unknown): void => {
...
+  const pick = {} as Pick<T, K>;
+  const unpick = {} as Pick<T, Exclude<keyof T, K>>;
```
stevekuznetsov, line 257:
```suggestion
  const unpick = {} as Omit<T, K>;
```

stevekuznetsov (commented):
Neat - just one suggestion to reduce the generics overload that helped me grok what was going on there - maybe we can add a docstring to `split` like

```js
// split partitions obj into two disjoint sets of properties, one containing any keys present in keySet, the other containing the rest - realizing Pick and Omit at runtime
```

stevekuznetsov:
Checked things out locally and I'm not seeing anything broken - so I think we're good? Agreed that if anything were wrong it would be _real_ wrong

components/core/View.tsx
floatplane: good callout

PR #614, opened 2023-12-07
*: add a helper to transform features to zones

stevekuznetsov:


components/content/ZoneMap.tsx
```diff
@@ -5,13 +5,27 @@ import MapView, {MAP_TYPES, MapViewProps, PoiClickEvent, Region} from 'react-nat
...
 import {useToggle} from 'hooks/useToggle';
-import {AvalancheCenterID, DangerLevel, Geometry} from 'types/nationalAvalancheCenter';
+import {AvalancheCenterID, DangerLevel, Geometry, MapLayerFeature} from 'types/nationalAvalancheCenter';
 
 const defaultAvalancheCenterMapRegionBounds: RegionBounds = {
   topLeft: {latitude: 0, longitude: 0},
   bottomRight: {latitude: 0, longitude: 0},
 };
 
+export const MapViewZoneFor = (center: AvalancheCenterID, feature: MapLayerFeature): MapViewZone => {
```
floatplane, line 15: nit: idiomatic usage would reserve a capitalized name for a React component, or a class. 
```suggestion
export const mapViewZoneFor = (center: AvalancheCenterID, feature: MapLayerFeature): MapViewZone => {
```

components/weather_data/WeatherStationMap.tsx
```diff
@@ -107,17 +107,13 @@ export const WeatherStationMap: React.FunctionComponent<{
...
-    end_date: feature.properties.end_date,
-    fillOpacity: 0.1,
-  }));
+  const zones: MapViewZone[] = mapLayer.features.map((feature: MapLayerFeature) => {
+    const zone = MapViewZoneFor(center_id, feature);
+    zone.hasWarning = false;
+    zone.danger_level = DangerLevel.None;
+    zone.fillOpacity = 0.1;
+    return zone;
+  });
```
floatplane, line 116: you could get more terse here and eliminate the explicit return 😁. I think reasonable people could disagree on that one though - totally up to you.

```suggestion
  const zones: MapViewZone[] = mapLayer.features.map((feature: MapLayerFeature) => ({
    ...mapViewZoneFor(center_id, feature),
    hasWarning: false,
    danger_level: DangerLevel.None,
    fillOpacity: 0.1,
  }));
```

floatplane (changes requested):
the capitalization thing is the only feedback that really needs addressing IMO - the other thing is totally optional!

components/weather_data/WeatherStationMap.tsx
stevekuznetsov: It looks like it would on the surface, but I didn't know if that was the sort of thing that would get JIT-ed out or not! Also was keen to hear if you thought it was more of an academic concern or would have any real impact on the app, thanks for sharing your thoughts!

stevekuznetsov:
@floatplane looks like you need to approve the PR before auto-merge kicks in!

floatplane (approved):

floatplane:
TIL about auto merge!

PR #617, opened 2023-12-07
Miscellaneous Fixes for CBAC

stevekuznetsov:
network: only prefetch NWAC forecasts when center is NWAC

---

observations: submit drafts when ACs request to review obs

---

### DIFF
--- a/components/observations/ObservationFormData.ts
+++ b/components/observations/ObservationFormData.ts
@@ -33,7 +33,7 @@ export const defaultObservationFormData = (initialValues: Partial<ObservationFor
       photoUsage: MediaUsage.Credit,
       private: false,
       start_date: new Date(),
-      status: 'published',
+      status: 'draft',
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
CBAC metadata is broken on staging so I can't test this right now.

stevekuznetsov:
We may also want to change the toasts that show up when we're submitting a draft to reflect that the observation needs to be approved before it's seen in the list.

stevekuznetsov:
@charlotteguard what do you think about getting the app set up to submit draft/review obs as necessary for centers that have requested it, and figuring out if we need to message this differently to users as a follow-up?

stevekuznetsov:
@charlotteguard confirmed we can do this behavioral fix now and wait on getting feedback on UX for other centers - lots of questions for what users should see once the observation is submitted but not reviewed.

PR #618, opened 2023-12-08
components/forecast: show toasts correctly

stevekuznetsov:
network: only prefetch NWAC forecasts when center is NWAC

---

components/forecast: show toasts correctly

For whatever reason, `useTimeout()` needs to be used when calling
`Toast.show()` in a hook. Without that, we weren't seeing any of these
show up, ever. In addition, we now use the `useFocusEffect()` hook from
`react-navigation` for a much more robust approach to showing and hiding
this toast when folks navigate around the app while the toast is open.

---

### DIFF
--- a/components/forecast/AvalancheTab.tsx
+++ b/components/forecast/AvalancheTab.tsx
@@ -1,9 +1,9 @@
-import React, {useCallback, useEffect} from 'react';
+import React, {useCallback} from 'react';
 
 import {addDays, formatDistanceToNow, isAfter} from 'date-fns';
 
 import {Feather, FontAwesome} from '@expo/vector-icons';
-import {useNavigation} from '@react-navigation/native';
+import {useFocusEffect, useNavigation} from '@react-navigation/native';
 import {AvalancheDangerIcon} from 'components/AvalancheDangerIcon';
 import {AvalancheDangerTable} from 'components/AvalancheDangerTable';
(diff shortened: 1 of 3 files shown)


floatplane:
> cc @floatplane I found the useEffect + useTimeout behavior super confusing. I really don't know what I am doing and I can't imagine this is the correct approach ... I stole it from [here](https://stackoverflow.com/a/69236626) but it does seem to function? None of the documentation around react-native-toast-message shows any issue with useEffect.

Were you seeing console warnings of the form "can't update one component while rendering another"? 

My guess would be that the function you pass to `useFocusEffect` gets executed during render, at which point you shouldn't be modifying React render state, in neither your component nor any other. Calling `Toast.show` does exactly that. To avoid modifying state during render, you have to call `Toast.show` off the render loop, either in an event handler or via setTimeout.

stevekuznetsov:
@floatplane right, that warning was what we had in the past with the direct call. You moved it into a `useEffect` to mitigate that, but it meant that the toasts never showed at all. I nested that even further into a `useTimeout` inside of the `useEffect` and it worked. I have no idea why the second nesting was needed or what it means.

floatplane:
Huh. I feel pretty certain I've seen these toasts, off and on, but I never gave them much thought! They definitely weren't showing before?

stevekuznetsov:
Yep - give it a `git checkout` before & after the `useEffect` commit to reproduce.

PR #622, opened 2023-12-09
Don't crash if an observation has a PDF attached

floatplane:
This isn't implemented on the NAC side yet, but they're going to start allowing people to attach PDFs to observations. This PR doesn't add support for displaying PDFs - it just does the bare minimum to keep the app from crashing when it tries to parse one of these new responses. I added a test to verify that it parses correctly.

I also added a fallback path, so if a new type gets added in the future it won't break the app - it will just get parsed as type `unknown`. 

@thechrislundy can you look at `types/nationalAvalancheCenter/tests/mediaitem.test.ts` and make sure that the structure of the object matches what you will be sending?

cc @stevekuznetsov @charlotteguard we'll want to push this out OTA soon

### DIFF
--- a/types/nationalAvalancheCenter/schemas.ts
+++ b/types/nationalAvalancheCenter/schemas.ts
@@ -129,6 +129,7 @@ export enum MediaType {
   Video = 'video',
   External = 'external',
   Photo = 'photo',
+  PDF = 'pdf',
   None = '',
 }
 
@@ -341,7 +342,22 @@ export const externalMediaSchema = imageMediaSchema.extend({
 });
 export type ExternalMediaItem = z.infer<typeof externalMediaSchema>;
 
-export const mediaItemSchema = z.discriminatedUnion('type', [emptyMediaSchema, nullMediaSchema, imageMediaSchema, videoMediaSchema, externalMediaSchema, photoMediaSchema]);
+export const pdfMediaSchema = z.object({
+  type: z.literal(MediaType.PDF),
+  url: z.object({
+    original: z.string().url(),
+  }),
+});
+
+export const mediaItemSchema = z.discriminatedUnion('type', [
+  emptyMediaSchema,
+  nullMediaSchema,
+  imageMediaSchema,
+  videoMediaSchema,
+  externalMediaSchema,
+  photoMediaSchema,
+  pdfMediaSchema,
+]);
 export type MediaItem = z.infer<typeof mediaItemSchema>;
 
 export const avalancheCenterWeatherConfigurationSchema = z.object({
--- a/types/nationalAvalancheCenter/schemas.ts
+++ b/types/nationalAvalancheCenter/schemas.ts
@@ -130,6 +130,7 @@ export enum MediaType {
   External = 'external',
   Photo = 'photo',
   PDF = 'pdf',
+  Unknown = 'unknown',
   None = '',
 }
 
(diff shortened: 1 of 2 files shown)


types/nationalAvalancheCenter/schemas.ts
```diff
@@ -341,7 +342,22 @@ export const externalMediaSchema = imageMediaSchema.extend({
...
+});
+
+export const mediaItemSchema = z.discriminatedUnion('type', [
+  emptyMediaSchema,
+  nullMediaSchema,
+  imageMediaSchema,
+  videoMediaSchema,
+  externalMediaSchema,
+  photoMediaSchema,
+  pdfMediaSchema,
```
floatplane, line 359: I feel like we should have a fallback or catchall schema here for an `unknown` media type. The client currently ignores schemas it isn't set up to display (`photo`, `video`, `external`) so it should be able to ignore new media types by default.

The problem is that I couldn't figure out the magic Zod invocation to make this work... 😬

floatplane:
Merged and deployed OTA update, tested on my iOS device and everything looks as expected

ccorey:
> I also added a fallback path, so if a new type gets added in the future it won't break the app - it will just get parsed as type unknown

Nice, yeah I was going to ask about that. In this case handing it gracefully seems like a good call, so we can be additive with new data. Perhaps the video upload at some point like you've suggested

floatplane:
> > I also added a fallback path, so if a new type gets added in the future it won't break the app - it will just get parsed as type unknown
> 
> Nice, yeah I was going to ask about that. In this case handing it gracefully seems like a good call, so we can be additive with new data. Perhaps the video upload at some point like you've suggested

FWIW video is already supported, @ccorey - though maybe you can only upload videos when using the NAC dashboard? But we don't display videos in the app yet 😢  

https://github.com/NWACus/avy/blob/9a1060045a2ac13fe59790e61d9a3dd3c6cbd00d/types/nationalAvalancheCenter/schemas.ts#L317-L335

thechrislundy:
Hey Brian,

The Dashboard supports adding a YouTube video to the media gallery. The
forecaster pastes the video link in, and the system retrieves a thumbnail
from YouTube and saves that along with the YouTube URL. So what comes back
from the API is a thumbnail image, the YouTube URL (so it can be embedded)
and the other media metadata.

We're not allowing the public to add videos, but it would be good if the
app supported playing the YouTube videos as forecasters often include them
on their obs.

Chris

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -341,7 +343,22 @@ export const externalMediaSchema = imageMediaSchema.extend({
...
+  }),
+});
+
+const unknownMediaSchema = z.object({
+  type: z.literal(MediaType.Unknown),
+});
+
+export const mediaItemSchema = z
+  .discriminatedUnion('type', [emptyMediaSchema, nullMediaSchema, imageMediaSchema, videoMediaSchema, externalMediaSchema, photoMediaSchema, pdfMediaSchema, unknownMediaSchema])
+  .catch({
```
stevekuznetsov, line 359: Nice. I was looking at `.catchall()` at the global scope to see if we could do something like this for all of our enums but that looks like it only applies to unknown keys, not unknown values. Wonder if you think I should add one of these to the rest of our enums? And maybe in debug mode or cron mode we yell instead of falling back so we know when we need to update?

PR #623, opened 2023-12-13
Use same analytics ID for PostHog & Mixpanel, and show it in the About page

### LINKED ISSUE #607: Smooth out the feature flag workflow
Currently, I'm not sure how we could selectively enable a feature flag for any one of us! There's a person ID that's visible on the Posthog side, but I have no idea where it comes from - which probably means the SDK defaults to supplying a UUID for you. No idea if the UUID is sticky (you'd want it to be). 

Tasks:
1. figure out how to pass a UUID of our choice for the posthog person ID
2. use the same UUID as the Mixpanel ID (`preferences.mixpanelUserId`)
3. expose the UUID in the about page so we can set it easily
4. (optional idea) if user ever enters an email address in the obs submit form, then save it and start sending that along as part of the posthog feature flag lookups. that'd be so much easier to use.

floatplane:
This fixes #607. With this, we can get a user ID from someone and add them to a PostHog cohort, which in turn can be used to control a feature flag.

Here's my user ID displayed in the about page:

image

---

And here's that user ID in the "Early access" cohort in PostHog:

image

---

And here's the feature flag rule for a user in the cohort:

image

---

And here's the flag being returned as true for my user:

image

---

And (finally!) here's me running the new weather table because the feature flag is on:

image

### DIFF
--- a/components/KillSwitchMonitor.tsx
+++ b/components/KillSwitchMonitor.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect} from 'react';
+import React, {useCallback, useEffect, useState} from 'react';
 import {Modal} from 'react-native';
 
 import * as Application from 'expo-application';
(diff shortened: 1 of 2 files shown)


floatplane:
blocking this one until #624 goes in

PR #624, opened 2023-12-14
Preserve user ID when clearing preferences

floatplane:
We use the "Reset preferences" option in the Secret Menu when we're testing and we want to go through startup flows again. But the unique user ID assigned when the user first runs the app is not a thing we should clear every time - we should keep it if possible! That way, we can consistently map events, feature flags etc to the same ID even after preferences are cleared.

### DIFF
--- a/Preferences.test.tsx
+++ b/Preferences.test.tsx
@@ -91,6 +91,7 @@ describe('Preferences', () => {
       const {result, waitForNextUpdate} = renderHook(() => usePreferences(), {wrapper: PreferencesProvider});
       expect(result.current.preferences.mixpanelUserId).toBeUndefined();
       await waitForNextUpdate();
+      // The center is invalid, so preferences parsing will fail. The previous user id is lost, and we get a different UUID in its place.
       expect(result.current.preferences.mixpanelUserId).toMatch(/^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$/);
       expect(result.current.preferences.mixpanelUserId).not.toEqual(userId);
     });
--- a/Preferences.test.tsx
+++ b/Preferences.test.tsx
@@ -2,7 +2,7 @@ import AsyncStorage from '@react-native-async-storage/async-storage';
 import {renderHook} from '@testing-library/react-hooks';
 
 import {PREFERENCES_KEY} from 'data/asyncStorageKeys';
-import {clearPreferences, PreferencesProvider, usePreferences} from 'Preferences';
+import {PreferencesProvider, resetPreferencesForTests, usePreferences} from 'Preferences';
 
 // Mock out AsyncStorage for tests
 // eslint-disable-next-line @typescript-eslint/no-unsafe-return
@@ -21,10 +21,10 @@ describe('Preferences', () => {
 
   it('updates after preferences are loaded', async () => {
     await AsyncStorage.setItem(PREFERENCES_KEY, JSON.stringify({center: 'BAC', hasSeenCenterPicker: true}));
-    const {result, waitForValueToChange} = renderHook(() => usePreferences(), {wrapper: PreferencesProvider});
+    const {result, waitForNextUpdate} = renderHook(() => usePreferences(), {wrapper: PreferencesProvider});
     expect(result.current.preferences.center).toEqual('NWAC');
 
-    await waitForValueToChange(() => result.current.preferences.center);
+    await waitForNextUpdate();
     expect(result.current.preferences.center).toEqual('BAC');
   });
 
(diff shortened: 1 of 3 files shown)


stevekuznetsov (commented):
Do you think we could break out user-provided preferences from user-opaque metadata entirely? Would using two files on `AsyncStorage` make this difficult to code against? Seems like we'd end up adding any other opaque items to the omit list if we gain more.

floatplane:
> Do you think we could break out user-provided preferences from user-opaque metadata entirely? Would using two files on `AsyncStorage` make this difficult to code against? Seems like we'd end up adding any other opaque items to the omit list if we gain more.

Yeah, we definitely could/should. I gotta think a bit about how to manage migrating the data to a new location, deprecating the old one, etc.

stevekuznetsov:
@floatplane do you want to merge this PR as-is while you figure out those bits?

floatplane:
Yeah, I think taking some more time to search for a better solution might be good. Undoing this PR is pretty quick, but I would like to tidy things up properly.

One thing that is quite annoying about the `usePreferences` interface is that it returns the `defaultPreferences` until the async file read is completed, after which it returns the _actual_ preferences. This results in defensive coding like this, where we can't ever assume that `mixpanelUserId` is set: https://github.com/NWACus/avy/blob/9a1060045a2ac13fe59790e61d9a3dd3c6cbd00d/App.tsx#L296-L303 

It would be nice if we could always assume that preferences were initialized before the app starts doing anything meaningful. I'd like to figure out a better way to handle these kind of async initializations in the startup flow (load preferences, load fonts, etc). I think there's a few other places where we kick off async tasks at startup and just kind of assume that they'll complete before anything meaningful happens.

floatplane:
@stevekuznetsov do you want to give this a 👍 / 👎 before I merge

Preferences.test.tsx
```diff
@@ -21,10 +21,10 @@ describe('Preferences', () => {
 
   it('updates after preferences are loaded', async () => {
     await AsyncStorage.setItem(PREFERENCES_KEY, JSON.stringify({center: 'BAC', hasSeenCenterPicker: true}));
-    const {result, waitForValueToChange} = renderHook(() => usePreferences(), {wrapper: PreferencesProvider});
+    const {result, waitForNextUpdate} = renderHook(() => usePreferences(), {wrapper: PreferencesProvider});
```
floatplane, line 24: using `waitForNextUpdate` is simpler and captures the spirit of what we're trying to do here. I made that change throughout.

Preferences.tsx
```diff
@@ -26,31 +26,33 @@ const defaultPreferences = preferencesSchema.parse({});
...
+  clearPreferences: () => undefined,
 });
 
 interface PreferencesProviderProps {
   children?: ReactNode;
 }
 
 export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({children}) => {
-  const [preferences, setFullPreferences] = useState<Preferences>(defaultPreferences);
+  const [preferences, setPreferences] = useState<Preferences>(defaultPreferences);
```
floatplane, line 43: `setFullPreferences` was a terrible name choice on my part

Preferences.tsx
```diff
@@ -67,16 +69,20 @@ export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({childre
     void AsyncStorage.setItem(PREFERENCES_KEY, JSON.stringify(preferences));
   }, [preferences]);
 
-  const setPreferences = (newPreferences: Partial<Preferences>) => {
-    setFullPreferences(merge({}, preferences, newPreferences));
+  const setPartialPreferences = (newPreferences: Partial<Preferences>) => {
```
floatplane, line 72: this is the function that gets publicly exposed, so that callers can easily set a single preference. no behavior change, it's just renamed internally to reflect that it accepts `Partial<Preferences>`.

Preferences.tsx
```diff
@@ -67,16 +69,20 @@ export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({childre
...
+
+  return <PreferencesContext.Provider value={{preferences, setPreferences: setPartialPreferences, clearPreferences}}>{children}</PreferencesContext.Provider>;
 };
 
 export const usePreferences = () => useContext(PreferencesContext);
 
-export const clearPreferences = async () => {
-  await AsyncStorage.removeItem(PREFERENCES_KEY);
+export const resetPreferencesForTests = async () => {
   await AsyncStorage.setItem(PREFERENCES_KEY, JSON.stringify(defaultPreferences));
```
floatplane, line 87: no need to remove an item that we're about to overwrite

stevekuznetsov:
All this looks good!

PR #626, opened 2023-12-14
Set 'Date taken' field for images uploaded to NAC

floatplane:
When we resize images before uploading, we lose EXIF metadata (AFAIK this statement is true for the Avy app _and_ for the web). A general solution for preserving the metadata is going to be hard and would require a *lot* of testing, but luckily the NAC API exposes a param for passing `date_taken` when uploading the image. So we grab the capture date from the original image's EXIF data (which is returned from the Expo image picker) and then we set the `date_taken` param appropriately.

The `captureDateFromExif` function is intentionally very conservative and failure-tolerant (and has tests to back that up). The worst case is that we fall back to the current behavior, where the "Date taken" field isn't set at all.

Here's an image I uploaded that has the "Date taken" field set correctly. It's a pretty great picture, I know.

image

cc @ccorey @stevekuznetsov @thechrislundy

### DIFF
--- a/components/observations/uploader/ObservationsUploader.ts
+++ b/components/observations/uploader/ObservationsUploader.ts
@@ -185,9 +185,8 @@ export class ObservationUploader {
               height: image.height,
               exif: image.exif
                 ? {
-                    // The type of ImagePickerAsset.exif is (unfortunately) Record<string, any>
-                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-                    Orientation: image.exif.Orientation,
+                    Orientation: typeof image.exif.Orientation === 'string' || typeof image.exif.Orientation === 'number' ? image.exif.Orientation : undefined,
+                    DateTimeOriginal: typeof image.exif.DateTimeOriginal === 'string' ? image.exif.DateTimeOriginal : undefined,
                   }
                 : undefined,
             },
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
Is there a downside the wholesale pulling the previous EXIF forward and adding it to the new image? I also remember some chatter over e-mail that would not require us to resize anything on-device - was that something NAC will do / has done?

thechrislundy:
Hey Steve - we're still expecting the client to resize the image, although
on the web widget we've increased the max dimension to 4000px.

Chris

floatplane:
> Is there a downside to wholesale pulling the previous EXIF forward and adding it to the new image when we resize it? 

The downside is that this functionality doesn't exist 😄. It is _theoretically_ possible to write an image file parser, extract the EXIF block and apply it to the new image. But it's a pretty big lift and you'd want a _lot_ of test coverage to be sure it worked.

Here's a [4 year old package](https://github.com/hMatoba/piexifjs) that looks like it could do the thing, but it would need massaging to pull it into TS. I played with it for a minute before giving up. Even if it worked I would have huge reservations about jamming it in to a new build!

> I also remember some chatter over e-mail that would not require us to resize anything on-device - was that something NAC will do / has done?

I believe NAC increased the allowed size but it's always an arms race. My iPhone produces 48 MP images that would still need to get resized. Also, we need to pre-rotate portrait images on the device, because the orientation field is in that EXIF block that gets dropped.

stevekuznetsov:
Here I was thinking that the image picker library would both provide EXIF parsing _and_ marshalling since they should be inverse transforms ... oh well :facepalm:

PR #633, opened 2023-12-15
Add CampaignManager

floatplane:
This sets up campaign tracking:
- campaigns are defined in code, see `data/campaigns/campaigns.ts`
- each campaign has the following properties: startDate, endDate, viewsPerDay, enabled
- CampaignManager exposes two methods
  - shouldShowCampaign: whether the campaign should be shown
  - recordCampaignView: record the campaign as having been seen

There's nothing actually rendered with these changes; it's merely setting up the data storage and adding tests.


data/campaigns/campaignManager.ts
```diff
@@ -0,0 +1,111 @@
...
+export type CampaignViewsSchema = z.infer<typeof campaignViewsSchema>;
+
+export interface ICampaignManager {
+  shouldShowCampaign(campaignId: CampaignId, atDate?: Date): boolean;
+  recordCampaignView(campaignId: CampaignId, atDate?: Date): Promise<void>;
+}
+
+class CampaignManager implements ICampaignManager {
+  private initialized = false;
+  private campaignViews: Record<CampaignId, CampaignViewSchema> = {} as Record<CampaignId, CampaignViewSchema>;
```
stevekuznetsov, line 28: @floatplane would it make sense to further index this by center ID? While I imagine we can deconflict names, more than one center having a campaign in the fourth quarter might like `campaign-q4-2023`

floatplane: overall I think it's best to keep it as a single string. they share a namespace with PostHog feature flags and Mixpanel tracking events (via `useCampaign`) so I would say we should deconflict names. now is probably a good time to do that, just to establish the pattern.

bigger problem is that we probably don't want `nwac-campaign-q4-2023` showing up when browsing Taos Avalanche Center. we could consider adding a list of allowed centers to the campaign definition, or I think you could do center filtering at the PostHog layer. having it explicit in code feels safer though.

floatplane: I guess different centers may want to see Mixpanel/Google Anayltics/whatever events flow into their _own_ dashboards? 🤔 there's going to be some thought required here.

PR #636, opened 2023-12-16
Show the fundraising campaign banner in the observation list

floatplane:
This shows a banner in the observation list, if and only if:

1. the `campaign-q4-2023` campaign is active
2. the `campaign-q4-2023` feature flag is set (currently it's on in `development`, off otherwise)
3. the active center is NWAC

The banner is sticky, sitting above the list. I'd expect we need to either make it smaller or let it be scrolled out of the way. Still waiting for final design on that one. 

Simulator Screen Recording - iPhone 14 Pro Max - 2023-12-15 at 16 27 04


components/observations/ObservationsListView.tsx
```diff
@@ -346,6 +361,13 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
         </ScrollView>
       </HStack>
       <Divider />
+      {showCampaign && (
+        // banner.png is 2000 x 729, which is an aspect ratio of 2.74348422
+        <Pressable onPress={openCampaignLink}>
+          {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment */}
+          <Image source={require('assets/campaigns/campaign-q4-2023/banner.png')} style={{width: '100%', height: undefined, aspectRatio: 2.74348422}} resizeMode="contain" />
```
stevekuznetsov, line 368: @floatplane if you look at the impl for avalanche center logo assets you will see the nonsense I had to do in order to get cached images to show for the dev server (`expo go`) - which is required to make asset loading work correctly when on the dev server but offline - and IIRC for images not to stutter for layout on a first render in prod - if it's not too much can we also do that here, too?

floatplane: > IIRC for images not to stutter for layout on a first render in prod

This feels very surprising to me - the asset should be bundled at that point, why would it stutter? Can we push a preview build and test this out in a prod-like environment?

My expectation is that a stutter might be visible on the dev server, since you're referencing an image that the dev server feeds you on demand. This behavior is what enables hot reloading of changed assets and generally seems like a good thing - I personally don't find it that troublesome...?

stevekuznetsov: Looking at the comments in https://github.com/NWACus/avy/commit/834efb81ff7a80cfe95b5ee7ddc1d695231af45f

> For whatever reason, `Image.prefetch()` does not work. Also, in
developer mode our local assets are served via localhost network and in
the future some assets might come from the Expo CDN, so in general we do
not know if our assets and images come from the filesystem or from a
local file. So, we can use a cache in the app's cache directory to hold
on to these images and display only from the local filesystem.

Looks like in production if we ship assets via OTA they come from a CDN.

stevekuznetsov: I think at a minimum making assets load correctly for dev is reasonable enough to have the couple LoC cost - WDYT? Reload still works as far as I can tell, I think they get unique URIs.

floatplane: ack, I'll take a look!

PR #638, opened 2023-12-17
Implement final(?) campaign designs, get modal working over map

floatplane:
This should be basically ready to go now, I think. Feature flag is still off so merging it should be safe.

- map modal displays at most every 8 hours (this can be tuned in `campaigns.ts`)
- campaign banner in obs list is smaller, can be scrolled away
- useCampaign hook returns a stable value as long as the active screen remains focused, so you can render with it

All the buttons and whatnot are wired up and sending events.

map | list
--- | ---
Simulator Screen Recording - iPhone 14 Pro Max - 2023-12-17 at 14 45 40 | Simulator Screen Recording - iPhone 14 Pro Max - 2023-12-17 at 14 54 10

### DIFF
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -148,7 +148,7 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
   useFocusEffect(
     useCallback(() => {
       // Bake in a delay before showing the campaign modal, so that it doesn't pop up before the map is drawn
-      const timeout = setTimeout(() => setScreenFocused(true), 2000);
+      const timeout = setTimeout(() => setScreenFocused(true), 200);
       return () => {
         clearTimeout(timeout);
         setScreenFocused(false);
--- a/components/AvalancheForecastZoneMap.tsx
+++ b/components/AvalancheForecastZoneMap.tsx
@@ -1,12 +1,15 @@
 import React, {useCallback, useRef, useState} from 'react';
 
-import {useNavigation} from '@react-navigation/native';
+import {useFocusEffect, useNavigation} from '@react-navigation/native';
 import {View as RNView, StyleSheet, Text, TouchableOpacity, useWindowDimensions} from 'react-native';
 import AnimatedMapView, {PoiClickEvent, Region} from 'react-native-maps';
+
+import * as Linking from 'expo-linking';
 
 import {useBottomTabBarHeight} from '@react-navigation/bottom-tabs';
 import {AvalancheDangerIcon} from 'components/AvalancheDangerIcon';
 import {colorFor} from 'components/AvalancheDangerTriangle';
+import {CampaignModal} from 'components/content/CampaignModal';
 import {incompleteQueryState, QueryState} from 'components/content/QueryState';
 import {defaultMapRegionForGeometries, MapViewZone, mapViewZoneFor, ZoneMap} from 'components/content/ZoneMap';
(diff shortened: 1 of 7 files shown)


components/AvalancheForecastZoneMap.tsx
```diff
@@ -132,6 +136,27 @@ export const AvalancheForecastZoneMap: React.FunctionComponent<MapProps> = ({cen
     [setPreferences, navigation],
   );
 
+  // Begin Q4 2023 campaign code
+  const [showCampaign, trackCampaign] = useCampaign('campaign-q4-2023', 'map-view');
+  const openCampaignLink = useCallback(() => {
+    trackCampaign();
+    const url = 'https://give.nwac.us/campaign/nwacs-year-end-fundraiser/c536433';
+    Linking.openURL(url).catch((error: Error) => logger.error('Error opening URL', {error, url}));
```
stevekuznetsov, line 144: Should we use the `Linking` library in the rest of the repo, like on the menu? If so - let's open an issue to track that?

components/content/CampaignModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const CampaignModal: React.FC<CampaignModalProps> = ({onAction = () => undefined}) => {
+  const [showModal, {off: hideModal}] = useToggle(true);
+  const activateCampaign = useCallback(() => {
+    hideModal();
+    onAction();
+  }, [hideModal, onAction]);
+
+  return (
+    // Pressing the Android back button dismisses the modal
```
stevekuznetsov, line 23: We had some feedback that the swipe-to-go-back gesture was not working to exit our observation filter modals - does the approach you're using here fix that as well? If so - let's capture that fix in an issue to update the other modals as well!

components/AvalancheForecastZoneMap.tsx
floatplane: I think we do already use it everywhere that's appropriate (places where we want to open the installed browser, as opposed to opening an in-app browser).

components/content/CampaignModal.tsx
floatplane: #640

components/AvalancheForecastZoneMap.tsx
stevekuznetsov: Ah, makes sense - for my reference, how do we choose which of the two to use?

floatplane: I think we use the Linking library everywhere that makes sense from a product perspective - places where we think we should open an external app, as opposed to showing the page inline in the app. 

The assumption here is that a donation page should probably open in the real browser, so that things like credit card autofill work most effectively.

floatplane:
thanks for taking a look! merging now

components/AvalancheForecastZoneMap.tsx
floatplane: there's probably a fuzzy element to this 😄. I think the product question is do we think this is essentially a "peek" kind of behavior - that the user wants to see something on a single page and then go back to the app? that's a great case for an in-app browser. on the other hand, do we think this might take them off on a series of links? that's probably a good time to open the real browser.

some things, like `mailto:` links, we know *have* to get passed off to the system, so we use `Linking.openURL` for that.

stevekuznetsov: Sweet, thanks for explaining that - the "peek" analogy really cleared up how to think about this for me!

PR #641, opened 2023-12-18
Add center filtering and tracking to campaigns

floatplane:
Implement some of the feedback from #638:

- namespace the campaign id (`nwac-campaign-q4-2023`)
- support filtering a campaign to a specific center (or list of centers)
- include the center id when sending campaign events to Mixpanel


floatplane:
I made it a draft because I thought there was something wrong with this, but I had forgotten to rename the matching feature flag to `nwac-campaign-q4-2023`.  Going to add checking for that case and warn appropriately.

floatplane:
added image caching as requested in https://github.com/NWACus/avy/pull/636#discussion_r1430490790

stevekuznetsov (approved):

floatplane:
here's a mock of the image loading. I kinda think that we should not show the campaign when we're offline

Simulator Screen Recording - iPhone 14 Pro Max - 2023-12-18 at 12 23 54

stevekuznetsov:
Although - if we're offline and it's cached, we can show it no problem. Guess that's a fine price to pay in order to now show that loading state ...

floatplane:
Yeah, I’m more just thinking of the experience- showing a modal where the button won’t do anything because they’re offline is not ideal. 

Feels like a P2 issue - I think the first campaign might be slightly scrappy given the compressed timeline 😄

stevekuznetsov:
Oh, haha - I didn't even think of the button! You're totally right.

PR #643, opened 2023-12-19
Use `avy_app` as source type for observations

floatplane:


types/nationalAvalancheCenter/schemas.ts
```diff
@@ -971,6 +971,8 @@ export const FormatPartnerType = (value: PartnerType): string => {
 const ObsSourceType = {
   Public: 'public',
   Dashboard: 'dashboard',
+  Widget: 'widget',
+  Avy: 'avy_app',
```
floatplane, line 975: @stevekuznetsov can you remind me again why this `const object as const` pattern exists instead of using enums? I genuinely can't remember!

floatplane: I am changing it to a regular enum for now, it doesn't seem like it introduces any issues and removes a hardcoded string elsewhere!

stevekuznetsov: If I remember correctly the `as const` thing with a POJO seemed like the preferred thing since:

 - it has no runtime cost (enums generate a runtime value, a namespace, some other bits)
 - the structure is immutable (apparently you can runtime mutate an enum? lol)
 - it allows `'avy_app'` to be used in addition to `ObsSourceType.Avy`

floatplane: > If I remember correctly the `as const` thing with a POJO seemed like the preferred thing since:
> 
> * it has no runtime cost (enums generate a runtime value, a namespace, some other bits)

According to [this](https://stackoverflow.com/a/66864086) they both compile to the same thing 🤷 . I buy that - namespace isn't a JS concept.

> * the structure is immutable (apparently you can runtime mutate an enum? lol)

I don't think you can mutate them! At least not naively so?

image

(I'd guess either one is probably mutable if you cast it as `any` first?)

> * it allows `'avy_app'` to be used in addition to `ObsSourceType.Avy`

This is maybe the bit that's a matter of taste? I actually don't love that unqualified constant and would rather type the enum reference out. OTOH I can see the viewpoint that the typechecker will enforce that you only type a valid string there.

stevekuznetsov: That StackOverflow link is incorrect - if you read the section on constant enums and objects in the [official docs](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) you'll see that they are entirely compile-time:

> To avoid paying the cost of extra generated code and additional indirection when accessing enum values, it’s possible to use const enums. ... Const enums can only use constant enum expressions and unlike regular enums they are completely removed during compilation. Const enum members are inlined at use sites. This is possible since const enums cannot have computed members.

Furthermore, the docs explicitly guide you to use POJO `as const`:

> In modern TypeScript, you may not need an enum when an object with as const could suffice ... 

I think since it gives us the smallest runtime cost, removes all the myriad bizarre gotchas with enums (see also the doc I linked) and it's parallel with every other enum in the types file, let's switch this back please.

stevekuznetsov: https://github.com/NWACus/avy/pull/644

stevekuznetsov: > I don't think you can mutate them! At least not naively so?

I think the argument (which is fairly moot in our codebase but nevertheless terrifies me) is that classic enums end up being literal runtime objects in the transpiled JS and can, in practice, be mutated at runtime without issue. Since the constant approaches inline string literals instead at callsites, that's not possible.

stevekuznetsov: > This is maybe the bit that's a matter of taste? 

Perhaps, but this is (I think) the only place where TypeScript is a nominal type system as opposed to a structural one. 

The language designers [seem to strongly suggest](https://www.youtube.com/watch?v=vBJF0cJ_3G0&t=1012s) that `const { ... } as const` is what we should use and that the enums are perhaps a mistake from early JS.

stevekuznetsov: Anyway, sorry for the ramble. When I made the decision to use `const { ... } as const` I was following the guidance of the docs and a couple of the blogs / thought-pieces I had read. I think if you have a strong reason to _prefer_ enums maybe it makes sense to move the whole file over but unless we have one of those reasons I don't know that it makes sense to move just one declaration ...

PR #647, opened 2023-12-21
Add version to feedback emails

yuliadub:
Also figured we can add to the log ones as well? but can remove if you do not think that is necessary. Tried to consolidate version into a separate file since we use it a few places. Still learning how this is all set up, let me know what you think!

what the email looks like (version is weird and no git revision since running locally in an emulator): 
image

### DIFF
--- a/components/screens/MenuScreen.tsx
+++ b/components/screens/MenuScreen.tsx
@@ -35,14 +35,17 @@ import {
   TimeMachine,
   ToastPreview,
 } from 'components/screens/menu/SecretMenu';
+import {getVersionInfoFull} from 'components/screens/menu/Version';
 import {NWACObservationScreen, ObservationScreen} from 'components/screens/ObservationsScreen';
 import {Body, BodyBlack, Title3Black} from 'components/text';
 import {openCampaignLink} from 'data/campaigns/openCampaignLink';
 import {useCampaign} from 'data/campaigns/useCampaign';
 import {settingsMenuItems} from 'data/settingsMenuItems';
 import {useAvalancheCenterMetadata} from 'hooks/useAvalancheCenterMetadata';
+import {getUpdateGroupId} from 'hooks/useEASUpdateStatus';
 import {LoggerContext, LoggerProps} from 'loggerContext';
 import {sendMail} from 'network/sendMail';
+import {usePreferences} from 'Preferences';
 import {colorLookup} from 'theme';
 import {AvalancheCenterID, userFacingCenterId} from 'types/nationalAvalancheCenter';
 
@@ -93,8 +96,22 @@ export const MenuScreen = (queryCache: QueryCache, avalancheCenterId: AvalancheC
   const {data} = useAvalancheCenterMetadata(avalancheCenterId);
   const menuItems = settingsMenuItems[avalancheCenterId];
 
+  const {
+    preferences: {mixpanelUserId},
+  } = usePreferences();
+  const [updateGroupId] = getUpdateGroupId();
+
   const MenuScreen = function (_: NativeStackScreenProps<MenuStackParamList, 'menu'>) {
-    const sendMailHandler = useCallback(() => void sendMail({to: 'developer+app-feedback@nwac.us', subject: 'NWAC app feedback', logger}), []);
+    const sendMailHandler = useCallback(
+      () =>
+        void sendMail({
+          to: 'developer+app-feedback@nwac.us',
+          subject: 'NWAC app feedback',
+          footer: `Please do not delete, info below helps with debugging.\n\n ${getVersionInfoFull(mixpanelUserId, updateGroupId)}`,
+          logger,
+        }),
+      [],
+    );
     // begin Q4 2023 campaign
     const [showCampaign, trackCampaign] = useCampaign(avalancheCenterId, 'nwac-campaign-q4-2023', 'menu-screen');
     const onCampaignAction = useCallback(() => {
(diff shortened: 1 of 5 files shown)


yuliadub:
aha need to make prettier happy! forgot about that

components/screens/menu/Version.tsx
```diff
@@ -0,0 +1,15 @@
+import * as Application from 'expo-application';
+import * as Updates from 'expo-updates';
+import {getUpdateTimeAsVersionString} from 'hooks/useEASUpdateStatus';
+import {Platform} from 'react-native';
+
+export const getVersionInfoFull = (mixpanelUserId: string, updateGroupId: string): string => {
```
yuliadub, line 6: could either mixpanel or update group Id ever be null and do we care if they are?

components/screens/MenuScreen.tsx
```diff
@@ -93,8 +96,22 @@ export const MenuScreen = (queryCache: QueryCache, avalancheCenterId: AvalancheC
   const {data} = useAvalancheCenterMetadata(avalancheCenterId);
   const menuItems = settingsMenuItems[avalancheCenterId];
 
+  const {
+    preferences: {mixpanelUserId},
+  } = usePreferences();
+  const [updateGroupId] = useState(getUpdateGroupId());
```
stevekuznetsov, line 102: We're not holding onto this state or mutating it in the future, so I don't think you need the `useState()` here:

```ts
const updateGroupId = getUpdateGroupId();
```

components/screens/menu/Version.tsx
stevekuznetsov: I think at least one of them can be, which is why you're getting the error you're getting above:

>  Argument of type 'string | undefined' is not assignable to parameter of type 'string'.

components/screens/menu/Version.tsx
```diff
@@ -0,0 +1,15 @@
+import * as Application from 'expo-application';
```
stevekuznetsov, line 1: nit: this file doesn't contain any JSX, so we usually just use a `.ts` extension for that.

components/screens/MenuScreen.tsx
yuliadub: total newb question, it seems like it wouldn't be needed here either then? https://github.com/NWACus/avy/pull/647/files#diff-93271258411bbda48ad01e2d70e98d5079555570679f1f32dad049f300874dd4L24

components/screens/menu/Version.tsx
yuliadub: so whats the best way to handle this? would we just pass an empty string to the version function if it is null for some reason or have the version function take untyped variable and handle it possibly being null in there? I played around with this a bit, but couldn't find a solution that looked right to me 🤷

components/screens/menu/AboutScreen.tsx
```diff
@@ -25,13 +26,7 @@ export const AboutScreen = (_: NativeStackScreenProps<MenuStackParamList, 'about
   const openUrl = useCallback(({data}: {data: string}) => void WebBrowser.openBrowserAsync(data), []);
   const copyVersionInfoToClipboard = useCallback(() => {
     void (async () => {
-      await Clipboard.setStringAsync(
-        `Avy version ${Application.nativeApplicationVersion || 'n/a'} (${Application.nativeBuildVersion || 'n/a'})
-Git revision ${process.env.EXPO_PUBLIC_GIT_REVISION || 'n/a'}
-Update group ID ${updateGroupId} (channel: ${Updates.channel || 'development'})
-Update ID ${Updates.updateId || 'n/a'} (platform: ${Platform.OS})
-User ID ${mixpanelUserId || 'n/a'}`,
```
stevekuznetsov, line 33: @yuliadub seems like we could make `getVersionInfoFull()` accept the MixPanel ID as `string | undefined` and inside of the function, do this `mixpanelUserId || 'n/a'` transform to handle the case where it's missing.

components/screens/menu/Version.ts
```diff
@@ -0,0 +1,15 @@
...
+import {getUpdateTimeAsVersionString} from 'hooks/useEASUpdateStatus';
+import {Platform} from 'react-native';
+
+export const getVersionInfoFull = (mixpanelUserId: string, updateGroupId: string): string => {
+  const versionTestFull = `Avy version ${Application.nativeApplicationVersion || 'n/a'} (${Application.nativeBuildVersion || 'n/a'})
+Build date ${getUpdateTimeAsVersionString()}
+Git revision ${process.env.EXPO_PUBLIC_GIT_REVISION || 'n/a'}
+Update group ID ${updateGroupId || 'n/a'} (channel: ${Updates.channel || 'development'})
+Update ID ${Updates.updateId || 'n/a'} (platform: ${Platform.OS})
+User ID ${mixpanelUserId || 'n/a'}`;
```
stevekuznetsov, line 12: Ah, looks like you are already doing the `|| 'n/a'` - so you should be able to just update the function signature.

components/screens/menu/Version.ts
```diff
@@ -0,0 +1,15 @@
+import * as Application from 'expo-application';
+import * as Updates from 'expo-updates';
+import {getUpdateTimeAsVersionString} from 'hooks/useEASUpdateStatus';
```
stevekuznetsov, line 3: You'll need to rebase - I changed this up recently, sorry!

yuliadub: whoops, sorry I missed that

yuliadub:
@stevekuznetsov I think this is good to go except for [this.](https://github.com/NWACus/avy/pull/647#discussion_r1436119895)  I kept the useState in here for now. Update groupd ID doesnt show up for me to test on the emulator but if I removed the used state weird things happen, so I am leaning to keep it in there for now.

stevekuznetsov:
@yuliadub sweet! Can you squash you commits down and we'll merge in?

PR #676, opened 2024-02-02
package.json: fix unmet peer dependencies

stevekuznetsov:
The GraphQL schemas we were working with are not useful anyway.

PR #683, opened 2024-02-14
Revert SDK 50 upgrade tsconfig changes

beaucollins:
Changes were made in #674

There was a problem with `ts-node` picking up the tsconfig referred to in `extends`. #676 fixed the problem the `.tsconfig` can be reverted to what it was.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,15 +1,6 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
   "compilerOptions": {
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
I wish this were less inscrutable :|

beaucollins:
VSCode does an ok job of pointing you to the documentation for each item in the config:

image

https://www.typescriptlang.org/tsconfig#outDir

PR #684, opened 2024-02-14
Memoize more ObservationListView state

beaucollins:
The assumption here is that the state from `usePendingObservations` is stable between calls. These changes then will crate stable references for `observationsSection` and `sections` between renders.

### DIFF
--- a/components/observations/ObservationsListView.tsx
+++ b/components/observations/ObservationsListView.tsx
@@ -168,19 +168,22 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   }
 
   const pendingObservations = usePendingObservations();
-  const pendingObservationsSection: Section = {
-    title: 'Pending',
-    data: pendingObservations.map(
-      ({observation}) =>
-        ({
-          id: observation.id,
-          observation: observation,
-          source: 'nac',
-          zone: observation.locationName,
-          pending: true,
-        } as const),
-    ),
-  };
+  const pendingObservationsSection: Section = useMemo(
+    () => ({
+      title: 'Pending',
+      data: pendingObservations.map(
+        ({observation}) =>
+          ({
+            id: observation.id,
+            observation: observation,
+            source: 'nac',
+            zone: observation.locationName,
+            pending: true,
+          } as const),
+      ),
+    }),
+    [pendingObservations],
+  );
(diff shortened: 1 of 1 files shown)


components/observations/ObservationsListView.tsx
```diff
@@ -1,44 +1,44 @@
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
```
stevekuznetsov, line 1: Does our prettier config do this on your machine? Yikes

stevekuznetsov (commented):
Seems reasonable - one question on formatting. Can't wait for `react-forget`.

components/observations/ObservationsListView.tsx
beaucollins: Ha, just reverting this now. This looks like my global prettier config getting opinionated.

components/observations/ObservationsListView.tsx
```diff
@@ -201,13 +204,17 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   );
 
   const hasPendingObservations = pendingObservationsSection.data.length > 0;
```
stevekuznetsov, line 206: Is there some gotcha here where the reference in `pendingObservationsSection` stays stable but the data underneath changes? Any reason not to depend on `pendingObservationsSection.data.length` in the dependency array and do this computation in the memo?

beaucollins: We could do this in the memo. I wasn't looking too closely at the code being memoized. In the end it wouldn't really change the performance characteristics since this boolean value being changed would mean `pendingObservartionsSection` changed which would bust the memo cache dependencies.

In the end, fewer dependencies is better.

beaucollins: Looks like `hasPendingObservations` is used elsewhere in the component, and memoizing that wouldn't be worth it. But we can still clean up the dependencies. `.length` is not an expensive operation.

beaucollins: The hope of this change is perhaps reducing renders of `ObservationListView` which does look like it could be a expensive.

beaucollins:
How is this my first time hearing of `react-forget`!?

Issue #688, opened 2024-02-16
Select Location in Obs is not working after the updrage to SDK 50


yuliadub:
**[Not currently impacting production, just the new dev builds post SDK 50 update]**

Selecting a location when submitting an observation works on Android with no issues, on Apple when you go to select a location in Obs form, nothing happens. If you click on it a bunch of times it may become disabled but that behavior is not consistent. 

If you run it on expo go on Apple, you will actually get an exception: 
Exception thrown while executing UI block: -[AIRGoogleMap setOnClick:]: unrecognized selector sent to instance 0x10c8cac00
RCTSurfaceStageIsPreparing
    RCTUIManager.m:1211
RCTSurfaceStageIsPreparing
5D16936B-4E4C-3276-BA7A-69C9BC760ABA
5D16936B-4E4C-3276-BA7A-69C9BC760ABA
5D16936B-4E4C-3276-BA7A-69C9BC760ABA
_dispatch_main_queue_callback_4CF
A900B459-0127-379E-9CBA-0EAB9C5D559F
A900B459-0127-379E-9CBA-0EAB9C5D559F
CFRunLoopRunSpecific
GSEventRunModal
7D57A1D1-856F-338D-97DB-880C4EC8B02E
UIApplicationMain
main
C3FC2EE4-367F-3086-BEB8-420AD442BF88

This likely sounds to be the issue: https://github.com/react-native-maps/react-native-maps/issues/4962

PR #689, opened 2024-02-16
Fix obs location for iOS

yuliadub:
Confirmed this fixes iOS and Android still works as expected. https://github.com/NWACus/avy/issues/688

for reference: https://github.com/facebook/react-native/issues/10180#issuecomment-298375648


stevekuznetsov:
Weird that we hit an issue from that many years ago but it does fix the issue ...

Fixes #688

Issue #696, opened 2024-03-06
Change Watch behavior 


charlotteguard:
If a watch is in effect there is no indication on the home screen map with all zone polygons. 
Once in the forecast there is a box at the top that looks like the warning that looks like an avalanche warning but the accent color is orange as opposed to red.

charlotteguard:
Asana link: https://app.asana.com/0/1203665433773397/1206593650595302/f

PR #698, opened 2024-03-16
AvalancheForecastZoneMap: update watch/bulletin behavior

stevekuznetsov:
The warnings endpoint can return three different types of products (warnings, watches and special bulletins). We only want the map to flash when there's an active warning.

A small fix to add 1 day to the warning fetch hook for the time machine, since the NAC API assumes a date and appends a time to it.


stevekuznetsov:
Fixes https://github.com/NWACus/avy/issues/696

PR #707, opened 2024-03-18
DO-NOT-MERGE: mapLayer: use the day query for previous layers

stevekuznetsov:
this works but the forecasts seemingly got updated with the newer zones? idk


kevinherdez:
Closing since this is being done in the new `useAllMapLayer` hook

Issue #656, opened 2024-01-24
We don't have a way to clear a location on an observation submission


floatplane:
This was reported as a bug by a user...we should take a look and come up with something, I'd expect it to be simple

PR #708, opened 2024-03-22
Add clearing of obs location

yuliadub:
attempt to solve https://github.com/NWACus/avy/issues/656. This changes the behavior a bit, where to select a location you now push a "check" location and "x" to clear it. The part that is not perfect is clearing the coordinates fully, you can spot that if you clear and then come back to set it again - the point will briefly be visible at the previous location if there was one. 

https://nwactechcommittee.slack.com/archives/C051KR96B2L/p1711086267002079?thread_ts=1711086244.167219&cid=C051KR96B2L

### DIFF
--- a/components/form/LocationField.tsx
+++ b/components/form/LocationField.tsx
@@ -45,13 +45,13 @@ export const LocationField = React.forwardRef<RNView, LocationFieldProps>(({name
 
   const toggleModal = useCallback(() => {
     setModalVisible(!modalVisible);
-     setCleared(false)
+    setCleared(false);
   }, [modalVisible, setModalVisible]);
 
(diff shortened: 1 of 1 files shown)


components/form/LocationField.tsx
```diff
@@ -72,6 +83,7 @@ export const LocationField = React.forwardRef<RNView, LocationFieldProps>(({name
   const onPress = useCallback(
```
stevekuznetsov, line 83: I think with the current implementation if you were to clear your location, the `field` would still have the data in there and you could still submit the observation with the previously chosen point. Do you think we could get clearing behavior by updating the form field to be empty?

yuliadub: ha, not sure why overcomplicated it so much originally. let me know if this looks more like it to you!

stevekuznetsov:
Looks good!

PR #709, opened 2024-03-23
Image caption editing

beaucollins:
Image caption editing for observation form uploads.

### iOS

https://github.com/NWACus/avy/assets/19795/003a6353-cf1a-4ee6-81b9-51b12e625097

### Android

**Note:** Android fails to auto-open when focusing a TextField. A potential hack is to use a `setTimeout` to open it after it's animated in.

https://github.com/NWACus/avy/assets/19795/9c9810b3-0015-462c-a81d-70039cd18f15

### DIFF
--- a/components/observations/ObservationImageEditView.tsx
+++ b/components/observations/ObservationImageEditView.tsx
@@ -1,0 +1,316 @@
+import {zodResolver} from '@hookform/resolvers/zod';
+import {Button} from 'components/content/Button';
+import {TextField, TextFieldComponent} from 'components/form/TextField';
+import {BodySemibold} from 'components/text';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
+import {FormProvider, useForm} from 'react-hook-form';
+import {Animated, Keyboard, KeyboardAvoidingView, LayoutChangeEvent, PanResponder, Platform, Pressable, StyleSheet, View} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+import {z} from 'zod';
+
+interface CaptionFormData {
+  caption: string;
+}
+
+interface Props {
+  onSetCaption: (caption: string) => void;
+  onDismiss: () => void;
+  onViewDismissed: () => void;
+  initialCaption?: string;
+  autoDismiss?: boolean;
+}
+
+const DefaultCaptionData: CaptionFormData = {
+  caption: '',
+};
+
+const CaptionFormTextField = TextField as TextFieldComponent<CaptionFormData>;
+
+const captionForm = z.object({
+  caption: z.string(),
+});
+
+// Behavior for KeyboardAvoidingView
+const BEHAVIOR = Platform.OS === 'ios' ? 'padding' : undefined;
+
+const INITIAL_HEIGHT = 320;
+const AUTO_DISMISS_DRAGGING_HEIGHT = 40;
+const AUTO_DISMISS_VELOCITY = 3;
+
+export const ObservationImageEditView: React.FC<Props> = ({onSetCaption, onDismiss, initialCaption, autoDismiss = true}) => {
+  const ref = useRef<View>(null);
+  const fadeAnim = useRef(new Animated.Value(0)).current; // Initial value for opacity: 0
+
+  // There is only one value, but using <TextField /> requires the use of a <FormProvider />
+  const formContext = useForm<CaptionFormData>({
+    defaultValues: initialCaption ? {caption: initialCaption} : DefaultCaptionData,
+    resolver: zodResolver(captionForm),
+    mode: 'onBlur',
+    shouldFocusError: false,
+    shouldUnregister: true,
+  });
+
+  const dragAnimationRef = useRef(new Animated.Value(INITIAL_HEIGHT, {useNativeDriver: false}));
+  const basisAnimationRef = useRef(Animated.add<number>(INITIAL_HEIGHT, Animated.multiply<number>(-1, dragAnimationRef.current)));
+
+  const edges = useSafeAreaInsets();
+
+  const [isDismissing, setDismissing] = useState(false);
+  const [minMaxHeight, setMinMaxHeight] = useState({min: 0, max: 0});
+
+  /**
+   * When the keyboard is presented of dismissed, onLayout will fire with the containing view
+   * that determines the maximum height the draggable view can be.
+   */
+  const finalHeight = useRef(INITIAL_HEIGHT);
+
+  const onLayout = useCallback(
+    (event: LayoutChangeEvent) => {
+      if (isDismissing) {
+        setMinMaxHeight(current => ({min: 0, max: current.max}));
+        return;
+      }
+      const height = event.nativeEvent.layout.height;
+      finalHeight.current = height;
+      const maxHeight = height - edges.top;
+      setMinMaxHeight({min: 24, max: maxHeight});
+    },
+    [edges.top, isDismissing],
+  );
+
+  const clampedAnimation = useMemo(() => {
+    const {min, max} = minMaxHeight;
+    return Animated.diffClamp(basisAnimationRef.current, min, max);
+  }, [minMaxHeight]);
+
+  const dismissedRef = useRef(false);
+  const onDragDismiss = useCallback((velocity?: number) => {
+    if (dismissedRef.current) return;
+    if (velocity != null && velocity > AUTO_DISMISS_VELOCITY) {
+      Keyboard.dismiss();
+    }
+    ref.current?.measure((_x, _y, _w, h) => {
+      if (h < AUTO_DISMISS_DRAGGING_HEIGHT) {
+        dismissedRef.current = true;
+        Keyboard.dismiss();
+      }
+    });
+  }, []);
+
+  const panResponder = useRef(
+    PanResponder.create({
+      onMoveShouldSetPanResponder: () => true,
+      onStartShouldSetPanResponder: () => true,
+      onPanResponderMove: Animated.event([null, {dy: dragAnimationRef.current}], {useNativeDriver: false}),
+      onPanResponderRelease: (_, gestureState) => {
+        onDragDismiss(gestureState.vy);
+        Animated.decay(dragAnimationRef.current, {
+          velocity: gestureState.vy,
+          deceleration: 0.98,
+          useNativeDriver: false,
+        }).start(() => {
+          dragAnimationRef.current.extractOffset();
+          onDragDismiss();
+        });
+      },
+    }),
+  ).current;
+
+  const onSubmitPress = useCallback(() => {
+    void (async () => {
+      await formContext.trigger();
+
+      const submit = formContext.handleSubmit(values => {
+        onSetCaption(values.caption);
+      });
+
+      await submit();
+
+      if (autoDismiss) {
+        Keyboard.dismiss();
+      }
+    })();
+  }, [formContext, onSetCaption, autoDismiss]);
+
+  const onPressOverlay = useCallback(() => {
+    Keyboard.dismiss();
+  }, []);
+
+  useEffect(() => {
+    const keyboardSubscriptions = [
+      Keyboard.addListener('keyboardWillShow', () => {
+        setDismissing(false);
+        Animated.parallel([
+          Animated.timing(fadeAnim, {
+            toValue: 1,
+            duration: 200,
+            useNativeDriver: true,
+          }),
+          Animated.spring(dragAnimationRef.current, {
+            toValue: 0,
+            useNativeDriver: false,
+          }),
+        ]).start();
+      }),
+
+      Keyboard.addListener('keyboardWillHide', () => {
+        setDismissing(true);
+        Animated.parallel([
+          Animated.timing(fadeAnim, {
+            toValue: 0,
+            duration: 100,
+            useNativeDriver: true,
+          }),
+          Animated.spring(dragAnimationRef.current, {
+            toValue: finalHeight.current,
+            useNativeDriver: false,
+          }),
+        ]).start(() => {
+          onDismiss();
+        });
+      }),
+
+      Keyboard.addListener('keyboardDidHide', () => {
+        onDismiss();
+      }),
+    ];
+
+    return () => {
+      for (const subscription of keyboardSubscriptions) {
+        subscription.remove();
+      }
+    };
+  }, [fadeAnim, onDismiss]);
+
+  return (
+    <KeyboardAvoidingView behavior={BEHAVIOR} style={[styles.flexV, styles.container]}>
+      <Animated.View style={[styles.overlay, {opacity: fadeAnim}]} />
+
+      <View onLayout={onLayout} style={[styles.flexV, {height: '100%'}]}>
+        <FormProvider {...formContext}>
+          <Pressable style={styles.dismissArea} onPress={onPressOverlay} />
+          <Animated.View collapsable={false} ref={ref} style={[styles.flexV, styles.form, {flexBasis: clampedAnimation}]}>
+            <View style={styles.formHeader} {...panResponder.panHandlers}>
+              <BodySemibold>Photo Description</BodySemibold>
+              <View style={styles.grip}>
+                <View style={styles.handle} />
+              </View>
+            </View>
+            <View style={[styles.flexV, styles.formFields]}>
+              <CaptionFormTextField
+                name="caption"
+                style={styles.textField}
+                label="Caption"
+                hideLabel
+                textInputProps={{
+                  placeholder: 'Write a short image description…',
+                  multiline: true,
+                  autoFocus: true,
+                }}
+              />
+              <Button mx={0} mt={8} buttonStyle="primary" onPress={onSubmitPress}>
+                <BodySemibold>Save</BodySemibold>
+              </Button>
+              <Button mx={0} mt={8} buttonStyle="normal" borderWidth={0} onPress={onPressOverlay}>
+                <BodySemibold>Cancel</BodySemibold>
+              </Button>
+            </View>
+          </Animated.View>
+        </FormProvider>
+      </View>
+    </KeyboardAvoidingView>
+  );
+};
+
+const styles = StyleSheet.create({
+  /**
+   * Using these styles is the equivalent of using VStack
+   */
+  flexV: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'stretch',
+  },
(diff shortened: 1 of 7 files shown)


components/observations/ObservationImageEditView.tsx
```diff
@@ -0,0 +1,122 @@
...
+            onSetCaption(values.caption);
+          })()
+          .catch(e => {
+            throw e;
+          });
+      })
+      .catch(e => {
+        throw e;
+      });
+  }, [formContext, onSetCaption]);
```
beaucollins, line 62: I broke the pattern and didn't use `async`/`await` here because technically using `async` for this function is "wrong". The way it's used in this project with callback props results in unhandled rejected promises ([an example](https://github.com/NWACus/avy/blob/d3bcba00a7ead135207ee044b7def919cad1b158/components/screens/menu/AboutScreen.tsx#L30-L32)].

Another pattern I saw was using an immediately evaluated async function.

The `.then` and `.catch` methods of a `Promise` are pretty straight forward but maybe not preferred in this project?

I suppose it's done this because Sentry should be registering an event listener for unhandled rejections and they'll end up being captured there.

I guess I'm just sad that the "no un-awaited promises" rule makes one right code like this:

```ts
voyd (async () => {
  await myPromise;
})()
```

components/core/View.tsx
```diff
@@ -38,6 +38,9 @@ interface ViewStyleProps {
   flexGrow?: RNViewStyle['flexGrow'];
   flexShrink?: RNViewStyle['flexShrink'];
   flexWrap?: RNViewStyle['flexWrap'];
+  gap?: RNViewStyle['gap'];
```
stevekuznetsov: This seems like a leaky abstraction - do you have any suggestions for how we might improve it in the future?

components/observations/SimpleForm.tsx
```diff
@@ -175,7 +195,37 @@ export const SimpleForm: React.FC<{
     },
     [images, setImages],
   );
-  const renderOverlay = useCallback((index: number) => <ImageListOverlay index={index} onPress={removeImage} />, [removeImage]);
+
+  const editImage = useCallback(
+    (index: number) => {
+      const image = images[index];
+      if (image == null) {
```
stevekuznetsov, line 202: nit: here, elsewhere - is a `== null` check more brittle than truthiness? Are we just inferring that `image` can't be `undefined`?

components/screens/ObservationsScreen.tsx
```diff
@@ -46,6 +60,42 @@ const ObservationsListScreen = ({route}: NativeStackScreenProps<ObservationsStac
...
+const ObservationImageEditScreen = ({route, navigation}: ObservationStackProps<'observationEditImage'>) => {
+  const {image, caption} = route.params;
+
+  const returnToScreen = lastScreen(navigation.getState());
+
+  const handleSetCaption = useCallback(
+    (caption: string) => {
+      switch (returnToScreen.name) {
+        case 'observationSubmit': {
+          // Ideally the entire route would refine types based os the returnToScreen.name refinement
```
stevekuznetsov, line 86: `s/os/on/`

components/screens/ObservationsScreen.tsx
```diff
@@ -46,6 +60,42 @@ const ObservationsListScreen = ({route}: NativeStackScreenProps<ObservationsStac
   );
 };
 
+/**
+ * Return the previous screen from the route stack.
+ */
+const lastScreen = (state: StackNavigationState<ObservationsStackParamList>) => state.routes[state.index - 1];
```
stevekuznetsov, line 71: nit: for really simple things & especially as they're just called once, I prefer to inline this

components/screens/ObservationsScreen.tsx
```diff
@@ -46,6 +60,42 @@ const ObservationsListScreen = ({route}: NativeStackScreenProps<ObservationsStac
...
+          // on the name
+          const submitRoute = returnToScreen as ObservationStackProps<'observationSubmit'>['route'];
+          navigation.navigate(submitRoute.name, {
+            ...submitRoute.params,
+            set_image_caption: [image, caption],
+          });
+          break;
+        }
+        default: {
+          // noop
```
stevekuznetsov, line 99: The implication of this is that the app gets stuck in this view without a way to navigate back, right?

stevekuznetsov (commented):
This looks great! One bigger question I had - is it possible to forward the reference to the form field like we do for the `LocationField`? It seems like that might make the state management easier here, allowing us to simply set the caption value in the `ObservationImageEditView` itself as opposed to having to use a navigation prop to back-propagate a setter.

components/observations/SimpleForm.tsx
beaucollins: Using `== null` is a standard JS way of refining out `T | null | undefined` to `T`. There's even a lint for it to prevent you from doing just a truthiness check: https://typescript-eslint.io/rules/strict-boolean-expressions/

And for refining out `null` explicitly truthiness checks have weird edge cases https://eslint.org/docs/latest/rules/no-eq-null

As far as using truthiness, there's enough weird edge cases around `if (image) {` that there's also a lint for no doing truthiness checks unless `image` is strictly a `boolean.` 

Index access of a Record type is also considered unsafe and there's stricter flags to make that the case:
https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

These are all things I have turned on in my projects so this pattern is somewhat muscle memory for me.

And JavaScript is ridiculous.

components/core/View.tsx
beaucollins: `StyleSheet` is probably worth leveraging. But I think we could really benefit from using `styled-components` or `emotion` which both have `react-native` integrations.

components/screens/ObservationsScreen.tsx
beaucollins: Yeah, my thinking here for the `default` case is to either `throw` or use `.goBack()`. It could also `throw` in development builds and `goBack` in production ones.

beaucollins: Another option is to use a function as a param. This works but `react-navigation` warns that it's not serializable so doesn't work with deep linking, which I don't think we'd ever use for this screen anyway.

beaucollins:
> It seems like that might make the state management easier here, allowing us to simply set the caption value in the ObservationImageEditView itself as opposed to having to use a navigation prop to back-propagate a setter.

The tradeoff here is pretty big for the user experience we can deliver. Probably worth syncing up on to see which ones we want to make.

components/observations/SimpleForm.tsx
beaucollins: TL;DR in this case a truthiness check is fine so I can change, just muscle memory for those specific lints.

beaucollins:
I don't think we'd be able to duplicate the form management that `TextField` does, but the end user experience will be more like this where the navigation bar will be on top of everything still.

`react-navigation` doesn't seem to give us any good apis to control the header or tab bar presentation from one of its screens (or child Navigator).

https://github.com/NWACus/avy/assets/19795/d5259da8-3bf4-42e4-b398-2793ef59189b

components/screens/ObservationsScreen.tsx
beaucollins: More docs on how `react-navigation` thinks this should be handled https://reactnavigation.org/docs/params#passing-params-to-a-previous-screen

You _can_ use a callback function and it works. This is the warning you're given

```
 WARN  Non-serializable values were found in the navigation state. Check:

Observations > observationEditImage > params.onSetImageCaption (Function)

This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you can use 'navigation.setOptions' instead. See https://reactnavigation.org/docs/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state for more details. 
```

beaucollins:
Trying to see what we can do with the navigation bar and looks like we should be able to hide it from a screen, but only with Android

```
  /**
   * Sets the visibility of the navigation bar. Defaults to `false`.
   *
   * @platform android
   */
  navigationBarHidden?: boolean;
```

components/observations/SimpleForm.tsx
stevekuznetsov: Thanks as always for the deep insights! If we don't have those stricter flags on in this project, please feel free to turn them on! JavaScript is certainly ridiculous, thanks for helping me learn!

stevekuznetsov:
Hm, that's a huge shame ... point taken on the UX impact for sure. We were able to get a full-screen modal with the `AvalancheCenterSelectionModal` which seems to hide everything from the navigation layer, but I don't know if it would apply here. I don't want to ask you to go on a goose-chase but if you have a chance to look into that overlay, it would be good to know if it would be something we could leverage here for the captions to try to get the best of both worlds.

beaucollins:
Ha, the more obvious answer via your `AvalancheCenterSelectionModal` suggestion: just use `react-native`'s `Modal`.

beaucollins:
#721 updates the form handling so that `images` can be managed in a comparable way to the other fields.

beaucollins:
@stevekuznetsov this now implements everything based on a Modal. 

To make it work with `useForm()` like TextField does this is built on #721

components/observations/SimpleForm.tsx
```diff
@@ -1,65 +1,43 @@
...
-        style={{textAlign: 'center'}}
-        onPress={onPressHandler}
-      />
-    </View>
-  );
-};
-
 const useKeyboardVerticalOffset = () => {
-  return useHeaderHeight() + useSafeAreaInsets().top;
+  return useHeaderHeight(); // + useSafeAreaInsets().top;
```
stevekuznetsov, line 40: Is this a holdover from something?

stevekuznetsov (commented):
This looks great! Looks like something went sideways in our preview action so I'll take this for a spin I'm back to my desk on Monday - I think we'll be good to merge then!

components/observations/SimpleForm.tsx
```diff
@@ -67,10 +45,12 @@ const useKeyboardVerticalOffset = () => {
...
 };
 
-export const SimpleForm: React.FC<{
+interface SimpleFormProps {
   center_id: AvalancheCenterID;
   onClose?: () => void;
-}> = ({center_id, onClose}) => {
+}
+
+export const SimpleForm: React.FC<SimpleFormProps> = ({center_id, onClose}: SimpleFormProps) => {
```
beaucollins, line 53: This same lint false negative has happened a couple times, not sure what's going on.

```
/avy/components/observations/SimpleForm.tsx
  53:56  error  'center_id' is missing in props validation  react/prop-types
  53:67  error  'onClose' is missing in props validation    react/prop-types

✖ 2 problems (2 errors, 0 warnings)
```

components/observations/SimpleForm.tsx
beaucollins: It was the same fix as #720

stevekuznetsov:
When you rebase, if you don't mind squashing, it would help review if you'd like to have each commit get some attention.

beaucollins:
> When you rebase, if you don't mind squashing, it would help review if you'd like to have each commit get some attention.

Would it be useful to make a fresh PR since this has so much noise in it now related to all the others I've spun out of these changes?

stevekuznetsov:
Up to you! I don't mind keeping it here.

components/observations/ObservationImageEditView.tsx
```diff
@@ -0,0 +1,421 @@
...
+      return;
+    }
+    actions.dismiss();
+
+    transitionAnimationRef.current?.stop();
+    textInputRef.current?.blur();
+    transitionAnimationRef.current = Animated.parallel([
+      Animated.timing(fadeAnim.current, {
+        toValue: 0,
+        duration: 400,
```
stevekuznetsov, line 172: I'm not sure if this is indicative of what users will experience, but once I accidentally opened the modal and tried to close it before the animation had ended, which caused the input to get lost forever:

```
 LOG   >> ACTION {"action": "presentRequest"}
 LOG   >>  {"keyboardVisible": false, "mode": "resting", "viewBoundary": {"max": 0, "min": 0}}
 LOG   <<  {"keyboardVisible": false, "mode": "presentRequested", "mutatedAt": 1712235746648, "previousMode": "resting", "viewBoundary": {"max": 0, "min": 0}}
 LOG   >> ACTION {"action": "present"}
 LOG   >>  {"keyboardVisible": false, "mode": "presentRequested", "mutatedAt": 1712235746648, "previousMode": "resting", "viewBoundary": {"max": 0, "min": 0}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746660, "previousMode": "presentRequested", "viewBoundary": {"max": 0, "min": 0}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"max": 797}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746660, "previousMode": "presentRequested", "viewBoundary": {"max": 0, "min": 0}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746672, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 0}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 32}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746672, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 0}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746680, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 32}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 54}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746680, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 32}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746688, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 54}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"max": 461}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746688, "previousMode": "presentRequested", "viewBoundary": {"max": 797, "min": 54}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746729, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 53.9998779296875}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746729, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746818, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 53.9998779296875}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 54.0001220703125}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746818, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 53.9998779296875}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746835, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54.0001220703125}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 53.9998779296875}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746835, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54.0001220703125}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746868, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 53.9998779296875}}
 LOG   >> ACTION {"action": "setBoundary", "minMax": {"min": 54}}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746868, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 53.9998779296875}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746885, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   >> ACTION {"action": "dismissRequest"}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746885, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   <<  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746885, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   >> ACTION {"action": "presentCompleted"}
 LOG   >>  {"keyboardVisible": false, "mode": "presenting", "mutatedAt": 1712235746885, "previousMode": "presentRequested", "viewBoundary": {"max": 461, "min": 54}}
 LOG   <<  {"keyboardVisible": false, "mode": "presented", "mutatedAt": 1712235747671, "previousMode": "presenting", "viewBoundary": {"max": 461, "min": 54}}
```

components/observations/ObservationImageEditView.tsx
```diff
@@ -0,0 +1,421 @@
...
+  const formContext = useForm<CaptionFormData>({
+    defaultValues: initialCaption ? {caption: initialCaption} : DefaultCaptionData,
+    resolver: zodResolver(captionForm),
+    mode: 'onBlur',
+    shouldFocusError: false,
+    shouldUnregister: true,
+  });
+
+  // Animated value that corresponds to the height of the non-form area
+  const dragRef = useRef(new Animated.Value(INITIAL_HEIGHT, {useNativeDriver: false}));
```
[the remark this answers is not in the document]
beaucollins: It threw an error that the `flexBasis` can't be animated with the native driver.

I can look into using height instead. I was trying to leverage the flex layout here but height may also do the trick.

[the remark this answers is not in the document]
beaucollins: `height` is also not supported with `useNativeDriver: true`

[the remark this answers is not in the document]
beaucollins: https://reactnative.dev/docs/animations#caveats

PR #710, opened 2024-03-25
Improve <TextField /> usability

beaucollins:
- Allow setting a height on the container
 - Bordered container is now a high area to focus the field
 - Track focus state and change border color

https://github.com/NWACus/avy/assets/19795/388cbbfc-35d4-463f-8905-188ad0705923

Pulled out of #709


components/form/TextField.tsx
```diff
@@ -36,26 +36,69 @@ export const TextField = React.forwardRef<RNView, TextFieldProps>(({name, label,
...
+  const onFocus = useCallback(() => {
+    setFocused(true);
+  }, []);
+
+  const textFieldRef = useRef<TextInput>(null);
+  const [isFocused, setFocused] = useState(() => textFieldRef?.current?.isFocused ?? false);
+
+  useLayoutEffect(() => {
+    setFocused(textFieldRef?.current?.isFocused ?? false);
+  }, []);
```
stevekuznetsov, line 53: Does this dependency array need `textFieldRef`?

components/form/TextField.tsx
```diff
@@ -36,26 +36,69 @@ export const TextField = React.forwardRef<RNView, TextFieldProps>(({name, label,
...
+
+  const handleFocusField = useCallback(() => {
+    textFieldRef.current?.focus();
+  }, []);
```
stevekuznetsov, line 57: Does this dependency array need `textFieldRef`?

stevekuznetsov (commented):
Looks good! Just one question

components/form/TextField.tsx
beaucollins: No, refs are not to be used as dependencies. Neither is the setter of `useState`. I'll try to dig up the docs, but that's why there's no lint warning here or on line 57.

beaucollins: So I can't find any written docs on this (I know I've read some), but here's the unit tests for the lint rule describing how the `setState` and `useRef` values are considered to be static.

https://github.com/facebook/react/blob/d3037405c4c9ce8158ef9017e960ec97fb089df4/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js#L644-L656

stevekuznetsov: I see, interesting - so they're guaranteed static, so it's not necessary to put them in. Neat, TIL.

PR #716, opened 2024-03-27
Add type safety to fieldRefs

beaucollins:
`fieldRefs` is more accurately represented as a `const` `Record` type. By `as const`ing the strict key and value types are captured by the type system. 

Before, it's an array of objects that get searched through

image

After, it's a Record of explicit keys and ref values:

image

We don't need `getFieldRef` anymore, the keys are statically known. And no more iterating just to look up the correct ref.

### DIFF
--- a/components/observations/SimpleForm.tsx
+++ b/components/observations/SimpleForm.tsx
@@ -110,21 +110,21 @@ export const SimpleForm: React.FC<{
   }, [cracking, formContext]);
 
   const fieldRefs = useMemo(
-    () => [
-      {field: 'name', ref: React.createRef<RNView>()},
-      {field: 'email', ref: React.createRef<RNView>()},
-      {field: 'phone', ref: React.createRef<RNView>()},
-      {field: 'activity', ref: React.createRef<RNView>()},
-      {field: 'location_name', ref: React.createRef<RNView>()},
-      {field: 'location_point', ref: React.createRef<RNView>()},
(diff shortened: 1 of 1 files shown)


components/observations/SimpleForm.tsx
```diff
@@ -110,21 +110,21 @@ export const SimpleForm: React.FC<{
...
-      {field: 'phone', ref: React.createRef<RNView>()},
-      {field: 'activity', ref: React.createRef<RNView>()},
-      {field: 'location_name', ref: React.createRef<RNView>()},
-      {field: 'location_point', ref: React.createRef<RNView>()},
-      {field: 'instability.cracking_description', ref: React.createRef<RNView>()},
-      {field: 'instability.collapsing_description', ref: React.createRef<RNView>()},
-      {field: 'avalanches_summary', ref: React.createRef<RNView>()},
-      {field: 'observation_summary', ref: React.createRef<RNView>()},
-    ],
+    () =>
```
beaucollins, line 113: It might even work better to use `useRef` here and get rid of the `useMemo` closure.

stevekuznetsov:
Love it. I wish we could decorate the actual form context with the refs since we're not enforcing that the keys here match anything in the form, but /shrug. I know you saw some of the other TODOs around typing in here - we originally wrote this without strict tsconfig on, which let us be a little loosey-goosey, and adding types after the fact was tricky since I'm not an expert in TypeScript magic.

PR #721, opened 2024-03-29
SimpleForm treat images as form field

beaucollins:
Instead of managing `image` and `setImage` differently than other form fields, this uses the `useForm` apis to manage the `ObservationFormData`'s `images` field.

Functionally nothing has changed.


stevekuznetsov (approved):
Looks like this collided with one of your other PRs, but great work! This looks good.

PR #722, opened 2024-03-30
Add generics to TextField to fit with react-hook-form

beaucollins:
I saw the `todo:` and knew how to solve it. It's a bit of a doozy though.

The end result is that you can have stricter typing for the form fields. It _could_ be simplified if we only want `TextField` to ever work with `ObservationFormData`.

The end result is:

 - You can't use a `name` prop for a form field that is not a string:
    image
 - It infers the names you _can_ use:
   image

### DIFF
--- a/components/form/TextField.tsx
+++ b/components/form/TextField.tsx
@@ -1,14 +1,22 @@
 import {HStack, View, ViewProps, VStack} from 'components/core';
 import {BodySm, BodySmBlack, BodyXSm} from 'components/text';
-import React, {useCallback, useLayoutEffect, useRef, useState} from 'react';
-import {useController} from 'react-hook-form';
+import React, {Ref, useCallback, useLayoutEffect, useRef, useState} from 'react';
+import {FieldPathByValue, FieldValues, useController} from 'react-hook-form';
 import {Pressable, View as RNView, StyleSheet, TextInput, TextInputProps} from 'react-native';
 import {colorLookup} from 'theme';
 
 export type KeysMatching<T, V> = {[K in keyof T]-?: T[K] extends V ? K : never}[keyof T];
 
-interface TextFieldProps extends ViewProps {
-  name: string;
+/**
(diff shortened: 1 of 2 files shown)


components/form/TextField.tsx
```diff
@@ -80,7 +99,7 @@ export const TextField = React.forwardRef<RNView, TextFieldProps>(({name, label,
             onBlur={onBlur}
             onFocus={onFocus}
             onChangeText={onChangeText}
-            value={field.value as string} // TODO(skuznets): determine why the generics here don't collapse to string itself ...
+            value={field.value}
```
beaucollins, line 102: It now knows the field.value is constrained to a `string` type. Or more specifically `string | null | undefined`.

components/form/TextField.tsx
```diff
@@ -92,7 +111,35 @@ export const TextField = React.forwardRef<RNView, TextFieldProps>(({name, label,
...
+ *          <MyCoolTextField name="age" .../>
+ *         </>
+ *      );
+ *    };
+ */
+export type TextFieldComponent<TFieldValues extends FieldValues> = <TFieldName extends FieldPathByValue<TFieldValues, StringFieldValue>>(
+  props: React.PropsWithRef<TextFieldProps<TFieldValues, TFieldName>> & {ref?: Ref<RNView>},
+) => JSX.Element;
+
+export const TextField = React.forwardRef(_TextField) as (<TFieldValues extends FieldValues, TFieldName extends FieldPathByValue<TFieldValues, StringFieldValue>>(
```
beaucollins, line 139: `forwardRef` "erases" generics so you have to cast them back in with `as`.

stevekuznetsov (approved):
Very cool! Do you think it's possible to have this inferred from the form context? What you've got here is a huge improvement but still requires us to cast a couple things `as` correctly ... I was wondering if there's some possible state where the component can infer what the type of the value is based on the field from the form that was passed to it. Would be a cool update in the future if so.

PR #740, opened 2024-05-07
Initial attempt to get the share button to work

yuliadub:
This code is not pretty and really just is there to check if the universal link behavior will work the way we expect, since to test it we need to generate a developer build.


App.tsx
```diff
@@ -74,8 +74,10 @@ import {filterLoggedData} from 'logging/filterLoggedData';
 import mixpanel from 'mixpanel';
 import PostHog, {PostHogProvider} from 'posthog-react-native';
 import {startupUpdateCheck, UpdateStatus} from 'Updates';
+import * as Linking from 'expo-linking';
 
 logger.info('App starting.');
+const prefix = Linking.createURL('/');
```
stevekuznetsov, line 80: note to self: if this is just for supporting Expo Go, maybe they recommend some other approach? code that sits around only for debugging always feels a bit weird

components/observations/ObservationDetailView.tsx
```diff
@@ -216,34 +217,83 @@ export const ObservationCard: React.FunctionComponent<{
   }, [postHog, observation.center_id, observation.id]);
   useFocusEffect(recordAnalytics);
 
+// route.path will have the link we would put into the share button here
+const route = useRoute(); 
+
+// currently the back button will leave to the list of obs of your current default center
```
stevekuznetsov, line 223: It looks like we should be able to use a built-in hook [`useLinkBuilder()`](https://reactnavigation.org/docs/use-link-builder) to generate the link for this page - that would make sure we're not ever at risk of generating a link that would not be correct as per the linking configuration.

yuliadub: agreed - the link generation can def use work but I think my emphasis was lets just see if the sharing works as we expected, then we will actually clean up all the easier stuff 😄 but happy to clean up some of this now, whatever you think makes the most sense

App.tsx
yuliadub: I mean we can really just take it out, no need to test this in expo once we know it works via universal links since it doesnt actually help us that much anyways

components/observations/ObservationDetailView.tsx
stevekuznetsov: Let's see if we can do this in our first pass?

yuliadub: sounds good to me - I may have time later today or tomorrow, but if you have time sooner go ahead and make the change

Issue #693, opened 2024-02-21
Submit button from obs with a forecast zone not behaving


charlotteguard:
If I click on the submit an ob button (blue button in the lower right hand corner) from the obs view _within_ a fx view, it appears responsive but does not open the obs form. This is happening whether it reads "submit" or as just the "+" when you scroll down.  I don't know if this is helpful but it feels the same as tapping on a card to bring me to the avalanche forecast. Everything else - scrolling, loading more obs, clicking into an ob works... but not the button. No issue if I go to the observations from the global nav and click submit.

charlotteguard:
https://github.com/NWACus/avy/assets/119702342/12c5d383-98d8-4790-b051-a1f1eb48237b

stevekuznetsov:
I can't reproduce this even on my low-powered Android device. Even right after startup when things are pretty slow, pushing the button does get the modal to pop up - with some lag - but even just after one push. I wonder if the total ignoring of input is an S10e issue.

charlotteguard:
@stevekuznetsov this is issue we ran into on the call with Sierra

stevekuznetsov:
This is a dupe of https://github.com/NWACus/avy/issues/530

kevinherdez:
Yulia mentioned this issue to me while getting spun up, and I found a nav issue with the button when first launching the app. I think it could be the same issue.

Repro steps:
1. Open the app on a clean launch
2. Go to a forecast zone and then observations. make sure not to tap the observations tab before doing this
3. Submit button doesn't work

yuliadub:
i feel like Kevin's PR fixed this but we may need to double check with @charlotteguard

yuliadub:
Gonna close this for now and we can always reopen if it pops up again

PR #792, opened 2024-09-19
Issue 693: Fix Submit button not responding

kevinherdez:
This an issue that seems to be what's causing this issue #693 

On a clean launch, it doesn't seem like the navigator for the Observations screens are loaded until the user goes to that tab. If a user goes to a forecast page, without opening that tab, and then hits the Submit button, the app can't navigate to that screen and React Native throws this error:

`The action 'NAVIGATE' with payload {"name":"observationSubmit","params":{"center_id":"NWAC"}} was not handled by any navigator.`

This fix adds the `ObservationSubmitScreen` to the `AvalancheCenterStack.Navigator` in the `HomeScreen`

*Import thing to note*: This changes the behavior so that if the Submit button is tapped on a forecast screen, the `ObservationSubmitScreen` is added to the home screen stack which keeps the user in the Zones tab. It no longer changes to the Observations tab

| Before | After |
|---|---|
|SubmitBefore|SubmitAfter|


.gitignore
```diff
@@ -21,3 +21,5 @@ android
 # Jetbrains IDE files
 .idea
 
+# Ignore changes to the package-lock
```
kevinherdez, line 24: I added this because I noticed that the `package-lock.json` wasn't being tracked

components/screens/ObservationsScreen.tsx
yuliadub, line 1: doesnt look anything got changed in this file, just formatting? that sound right?

yuliadub:
nice catch - your fix makes sense to me! I can 100% repro what you are describing in expo, the preview and the production app (the submit button doesn't work for me if I go to the zone first vs loading the main obs screen from the main without a zone)

looks like we got some formatting spacing differences (I noticed this too when I recently submitted a PR after opening the files from this repo in Vscode). @stevekuznetsov will be back next week so we can decide if we want to revert those or maybe do a change for the whole repo.... I dont feel too strongly one way or another

kevinherdez:
> nice catch - your fix makes sense to me! I can 100% repro what you are describing in expo, the preview and the production app (the submit button doesn't work for me if I go to the zone first vs loading the main obs screen from the main without a zone)
> 
> looks like we got some formatting spacing differences (I noticed this too when I recently submitted a PR after opening the files from this repo in Vscode). @stevekuznetsov will be back next week so we can decide if we want to revert those or maybe do a change for the whole repo.... I dont feel too strongly one way or another

I didn't realize that there was CI checking for this, so I ran `yarn prettify` and it reverted those changes. It seems like there's a decision that was made here already on the styling, and I'm not familiar enough with JS to know which one is "standard" between what I have on VS Code and what `prettier` is doing, so I'll just keep using `prettier` for now

yuliadub:
> > nice catch - your fix makes sense to me! I can 100% repro what you are describing in expo, the preview and the production app (the submit button doesn't work for me if I go to the zone first vs loading the main obs screen from the main without a zone)
> > looks like we got some formatting spacing differences (I noticed this too when I recently submitted a PR after opening the files from this repo in Vscode). @stevekuznetsov will be back next week so we can decide if we want to revert those or maybe do a change for the whole repo.... I dont feel too strongly one way or another
> 
> I didn't realize that there was CI checking for this, so I ran `yarn prettify` and it reverted those changes. It seems like there's a decision that was made here already on the styling, and I'm not familiar enough with JS to know which one is "standard" between what I have on VS Code and what `prettier` is doing, so I'll just keep using `prettier` for now

sounds good! I (in all honesty) am really bad and dont have pre commit hooks set up (I really should) so I always forget and screw up styling as well. But agreed that sticking to what the repo has for now makes the most sense!

stevekuznetsov:
Thanks for the fix! Love to see it. I don't know there is any standard in JS for formatting so each project is free to configure their `.prettierrc`. You should be able to configure VSCode to honor it as well! I am surprised it does not do so by default.

Issue #747, opened 2024-05-23
When Panning Map, Don't Move the Pin Every Time


stevekuznetsov:
This is specifically on the obs submit form, when picking location for where the obs is coming from. Right not the Pin will move each time you touch the map with one finger.

kevinherdez:
I have a fix for this, but when working on it I came across some user experience things that maybe could be improved on. I wanted to get feedback before making the changes though.

Right now, it's not clear that closing the modal with the X button is going to erase the location as well. I think the standard practice is that it should close out, and keep everything as it was. I think we can make this more clear in 2 ways.

1) Add another button to the modal (like an eraser or trash can) that the user presses to clear the pin on the map, they would then hit the checkmark to save the clear

2) Add a "Clear location" button under the location form field, if the user has a location set

Let me know what you think. I'm also fine leaving it as is. It was just something that I found a little wonky when debugging this.

yuliadub:
@kevinherdez these are fair points! 

a little history on the view: at first, we only had the check mark and you could only select the location and there was no way to clear it. when we added the ability to clear, we went back and forth on this quite a bit (if the X made sense/etc, if it should be a different icon, we havent considered having the clear button outside or the map as far as I remember since we didnt want to make major changes to the obs form itself), and in the end the consensus was it would be good to get a design opinion on what user experience is best but we didn't at the time have a designer who can help out. With all that, totally open to changing it! 

if you have a vision of how it may work better, and its not a lot of work, I'd say - put it together so we can play with it. If its more then minimum amount of work, lets hold off for now -- we may have some designer help later on as well, so it is possible this whole experience may need a redesign. will leave a tag for @charlotteguard  to keep me honest :)

kevinherdez:
That sounds good! I'll go ahead and make a separate issue for this.

kevinherdez:
Made the new issue here #793

yuliadub:
this was fixed

PR #794, opened 2024-09-25
Issue 747: Stop pin from moving on map pans

kevinherdez:
This fixes the issue in #747 where the pin moved every time a user panned or zoomed in on the map. This was because no matter what the user did while interacting with the map, it would file an `onPress` that would then update the pin location. 

This fix replaces `<Pressable>` and the `onPress` with a `<View>` and `PanResponder`. The `PanResponder` keeps tracks of movements the user does on the `View` which are then used to determine if the user tapped the view or did a different gesture.

PanFix

### DIFF
--- a/components/form/LocationField.tsx
+++ b/components/form/LocationField.tsx
@@ -43,7 +43,6 @@ export const LocationField = React.forwardRef<RNView, LocationFieldProps>(({name
       onMoveShouldSetPanResponder: () => true,
       onPanResponderRelease: (event, gestureState) => {
         if (gestureState.dx != 0 && gestureState.dy != 0) {
-          console.log('Gesture ignored because the user moved the map');
           return;
         }
 
--- a/components/form/LocationField.tsx
+++ b/components/form/LocationField.tsx
@@ -8,7 +8,7 @@ import {Body, BodySmBlack, BodyXSm, Title3Black, bodySize} from 'components/text
 import {useMapLayer} from 'hooks/useMapLayer';
 import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {useController} from 'react-hook-form';
-import {GestureResponderEvent, Modal, Pressable, View as RNView, TouchableOpacity} from 'react-native';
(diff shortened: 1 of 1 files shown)


components/form/LocationField.tsx
```diff
@@ -37,6 +37,26 @@ export const LocationField = React.forwardRef<RNView, LocationFieldProps>(({name
   const [mapReady, setMapReady] = useState<boolean>(false);
   const mapRef = useRef<MapView>(null);
 
+  const mapPanResponder = useRef(
+    PanResponder.create({
+      onStartShouldSetPanResponder: () => true,
+      onMoveShouldSetPanResponder: () => true,
+      onPanResponderRelease: (event, gestureState) => {
+        if (gestureState.dx != 0 && gestureState.dy != 0) {
+          return;
```
kevinherdez, line 46: I had a `console.log` here, but that's not really anywhere else in the repo. Should I add a log here using the `Logger` class, or leave it as is?

components/form/LocationField.tsx
```diff
@@ -73,35 +93,25 @@ export const LocationField = React.forwardRef<RNView, LocationFieldProps>(({name
...
-          </View>
-          <Center px={8} borderLeftWidth={2} borderColor={colorLookup('border.base')}>
-            <FontAwesome name="map-marker" color={colorLookup('text')} size={bodySize} />
-          </Center>
-        </HStack>
-      </TouchableOpacity>
-      {/* TODO: animate the appearance/disappearance of the error string */}
-      {error && <BodyXSm color={colorLookup('error.900')}>{error.message}</BodyXSm>}
+    <>
+      <VStack width="100%" space={4} ref={ref}>
```
kevinherdez, line 100: I added this fragment to pull out the modal code from the `VStack`. It doesn't seem to change the functionality at all, but I think it makes it easier to read and understand what's going on

yuliadub:
@kevinherdez looks good to me! I'm gonna merge this - sound good to you?

stevekuznetsov (approved):

stevekuznetsov:
Thank you! Looks great.

PR #801, opened 2024-10-09
Share functionality testing

yuliadub:
Currently, if you load the expo preview for any of my attempts from this PR, I get a blue screen but then expo never loads anything. Sadly I dont get a crash or error or anything. When I load the branch in expo go on my computer in an emulator, it works with no issues. 

I have also tried creating a brand new app preview build and submitting it to the preview app, for both ios and android, the app had the same behavior on first launch as I see in expo preview, but for the second launch (if i went and killed it), it would just load the latest preview build with none of my changes. (build 17)

I have also tried creating an OTA update after creating a clean preview build (build 21) and then publishing it. The behavior is similar where when I open the app on first launch it hands on the blue screen, if I open it again after killing it, it will simply load the recent preview build (the one before the OTA). 

another weird thing: when I try to open a link to an obs with either build 17 or 21 installed on IOS, the link does open the app, just not the obs page. On Android this is not the case, the link always opens in the browser still. 

some things I thought may cause the issues that I have tried removing and testing: 
- app config changes
- linking changes in App.tsx
- using a different expo icon (Ant instead of current)

I also thought maybe this is being caused by some misconfiguration in how I create my builds (at first I was doing them from my local machine) but it seems to have the same behavior if I use existing github actions in the repo. At least expo preview should work for starters!


types/nationalAvalancheCenter/schemas.ts
```diff
@@ -25,6 +25,13 @@ export function userFacingCenterId(input: AvalancheCenterID): string {
   return input === 'SNFAC' ? 'SAC' : input;
 }
 
+export const AvalancheCenterWebsiteSchema = {'https://nwac.us': 'NWAC'} as const;
```
stevekuznetsov, line 28: Couple small comments here:
 - this is not really a schema in that it doesn't describe the shape of any data
 - when you have to use `as` it usually means something went wrong - here I think you could more explicitly model this as a `Record<AvalancheCenterID, string>`
 - I don't think you ever want to look up the center based on its website, so the `reverseLookup` is likely not necessary

So this could become:

```ts
export const AvalancheCenterWebsites: Record<AvalancheCenterID, string> = {
  ['BAC']: 'https://bridgeportavalanchecenter.org/',
  ['BTAC']: 'https://bridgertetonavalanchecenter.org/',
  ['CBAC']: 'https://cbavalanchecenter.org/',
  ['CNFAIC']: 'https://www.cnfaic.org/',
  ['COAA']: 'https://www.coavalanche.org/',
  ['ESAC']: 'https://www.esavalanche.org/',
  ['FAC']: 'https://www.flatheadavalanche.org/',
  ['HPAC']: 'https://hpavalanche.org/',
  ['IPAC']: 'https://www.idahopanhandleavalanche.org/',
  ['KPAC']: 'https://kachinapeaks.org/',
  ['MSAC']: 'https://www.shastaavalanche.org/',
  ['MWAC']: 'https://www.mountwashingtonavalanchecenter.org/',
  ['NWAC']: 'https://nwac.us/',
  ['PAC']: 'https://payetteavalanche.org/',
  ['SAC']: 'https://www.sierraavalanchecenter.org/',
  ['SNFAC']: 'https://www.sawtoothavalanche.com/',
  ['TAC']: 'https://taosavalanchecenter.org/',
  ['WAC']: 'https://wallowaavalanchecenter.org/',
  ['WCMAC']: 'https://missoulaavalanche.org/',
};
```
Look-ups to get a website URL are just a map access.

components/observations/ObservationDetailView.tsx
```diff
@@ -214,6 +218,38 @@ export const ObservationCard: React.FunctionComponent<{
   }, [postHog, observation.center_id, observation.id]);
   useFocusEffect(recordAnalytics);
 
+  // currently the back button will leave to the list of obs of your current default center
+  // even if someone shares an observation with you from a different center
+  // the header will still show the current center logo (that likely needs to change as design actually said to replace that with the share icon)
+  // to open in expo: *check url and port, and correct obs ID that exists
+  // example: npx uri-scheme open exp://192.168.1.8:8082/--/observation/866b81db-52b3-4f94-890c-0cae8f162097 --android
+  const url = FormatCenterWebsite(observation.center_id as AvalancheCenterWebsiteSchema) + '/observations/#/view/observations/' + observation.id;
+  const ShareButton = () => {
```
stevekuznetsov, line 227: Go ahead and label this as a component:

```ts
const ShareButton: React.FunctionComponent = () => {...}
```

components/observations/ObservationDetailView.tsx
```diff
@@ -240,6 +276,9 @@ export const ObservationCard: React.FunctionComponent<{
                       {observation.name || 'Unknown'}
                     </AllCapsSm>
                   </VStack>
+                  <VStack space={5} style={{flex: 0}}>
+                    {ShareButton()}
```
stevekuznetsov, line 280: Once it's a component, just embed it in JSX:

```jsx
                  <VStack space={5} style={{flex: 0}}>
                    <ShareButton/>
                 </VStack>
```

components/observations/ObservationDetailView.tsx
kevinherdez: I think the change is fine since it's the standard in the repo. Just wanted to point out that because of the return type (and I think the naming convention too) it's automatically typed to be a `React.FunctionComponent`!

App.tsx
```diff
@@ -474,11 +476,24 @@ const BaseApp: React.FunctionComponent<{
     );
   }
 
+  const linking = {
+    prefixes: [prefix, FormatCenterWebsite(avalancheCenterId as AvalancheCenterWebsiteSchema) + '/observations/#/view/'],
```
kevinherdez, line 480: I need to look more into how this works in RN, but you might not need this here. This seems like it should only be the base URLs that the app accepts

App.tsx
```diff
@@ -474,11 +476,24 @@ const BaseApp: React.FunctionComponent<{
...
+    config: {
+      screens: {
+        Observations: {
+          screens: {
+            observation: 'observations/:id',
```
kevinherdez, line 485: Here is where you could add `observations/#/view/:id` so that the navigation container associates the path to this screen. Not 100% positive about this though

components/observations/ObservationDetailView.tsx
stevekuznetsov: Sorry, not sure what you mean. We almost exclusively use explicit types:

```
$ git grep -E 'React.(FC|FunctionComponent)' | wc -l
132
$ git grep '= () =>' | grep -v use | wc -l
19
```

The inferred return type will be a union of all of the different `return` blocks, so not adding an explicit type will mean it's trivially easy to create whatever mistaken union of types happens to be the set of possible return values and to have one `return` site with a different type than the rest. We're using TypeScript to avoid these kinds of mistakes.

App.tsx
stevekuznetsov: This is documented as the configuration to map paths to screens: https://reactnavigation.org/docs/configuring-links#mapping-path-to-route-names

App.tsx
stevekuznetsov: Both of the following link to an individual observation:

https://nwac.us/observations/#/view/observations/fb5bb19a-2b89-4c9c-91d2-eb673c5ab877
https://nwac.us/observations/#/observation/fb5bb19a-2b89-4c9c-91d2-eb673c5ab877

The former (which is what you're writing about) overlays the observation on the list view, while the latter just renders it stand-alone. We might be able to support both, but if we can't I think what Yulia has now (support for the latter) is likely the better option as it maps more cleanly to the screen we're sending users to.

stevekuznetsov: Ah, actually ... looks like @yuliadub supports only the latter, with `https://nwac.us/observations/#/view/` configured as the accepted prefix and `observations/fb5bb19a-2b89-4c9c-91d2-eb673c5ab877` the path.

stevekuznetsov: In the future we will likely want to support `https://nwac.us/` as the prefix but if we're only trying to route obs pages for now, /shrug

App.tsx
kevinherdez: Oh sorry I meant the `/observations/#/view` might not be needed. In those docs it seems like just the base URLs are being used in the `prefixes`

App.tsx
kevinherdez: oooohhhh okay that makes more sense I see what's going on now. Thanks for clarifying

App.tsx
kevinherdez: I see what's going on here now

components/observations/ObservationDetailView.tsx
kevinherdez: I see and that makes sense. I was just pointing out that based on how this is written that it can already be embedded in the JSX as the type is inferred to be `React.JSX.Element` (I was wrong about it being a `React.FunctionComponent`). I understand wanting to maintain the type safety though

App.tsx
yuliadub: yeah I am also not 100% sure of this mapping and how it will work. From what I tested on expo, it seems like the way I have it set up makes sense....but need to try it on an actual non expo build to see what works -- once we get that working will appreciate any help figuring it out if it doesnt work

types/nationalAvalancheCenter/schemas.ts
yuliadub: lol this is me being lazy and copying and pasting things without using my brain 😄

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -25,6 +25,28 @@ export function userFacingCenterId(input: AvalancheCenterID): string {
   return input === 'SNFAC' ? 'SAC' : input;
 }
 
+export const AvalancheCenterWebsites: Record<AvalancheCenterID, string> = {
+  ['NWAC']: 'https://nwac.us/',
+  // ['BAC']: 'https://bridgeportavalanchecenter.org/',
```
yuliadub, line 30: what would be the best way here? To only populate centers that we support for sharing or implement some sort of other way (a boolean that quantifies if sharing status is on or off?) 

I thought of just having values be uncommented as we add them, but right now the enum wants all AvalancheCenterID values in here. And it made me wonder if that was actually the best way 🤔

App.tsx
yuliadub: > n.b. @yuliadub making these prefixes dynamic based on the currently-focused center means that someone can't open a link to another center's observations page if they aren't already looking at that center's data - is that what we want?

I went back and forth on this....i.e. do we hard code a list in there or load it dynamically based on the center you are in...welcome all thoughts and opinions on this since I dont 100% know what is going to work best (maybe once we figure out the user experience we want with opening a shared obs not from your center it could help guide us too)

types/nationalAvalancheCenter/schemas.ts
yuliadub: thats a good point - we can still share links from the app, they just wont open in the app for the centers that do not yet have support for opening links in the app.

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -17,6 +17,8 @@ export const avalancheCenterIDSchema = z.enum([
   'TAC', // Taos: NM
   'WAC', // Wallowas: OR
   'WCMAC', // West Central Montana: MT
+  'CNFAIC', // Chugach : AK
```
yuliadub, line 20: @stevekuznetsov these are special case ones, right? they were in the link of urls but not in here, I assume we want to add them?

stevekuznetsov: I think you need to rebase, they are now in the types.

Issue #809, opened 2024-10-24
Finish sharing feature


yuliadub:
What still needs to be done for sharing to work correctly e2e: 
- finish back button work
1. don't show the shared obs once navigated away from it to other pages
2. show back button when a shared obs is open with the first launch of the app
3. display the correct logo in the corner when the shared obs is not from the default center a user has selected

PR #814, opened 2024-10-28
Adding behavior to ensure sharing is working correctly

yuliadub:
resolves https://github.com/NWACus/avy/issues/809

A few notes: 
- couldn't quite figure out how to navigate to Home without the navigation reset, given the way navigation is created today as far as I understood. Open to feedback there
- Need to do a whole bunch of testing on the preview app, since this is hard to do in expo to make sure things work as expected (as they did for me in expo on emulator) Specifically I want to make sure we get the universal link as expected and are able to get the "which center the obs was shared from". I didnt want to have to make any sort of API call to the NAC to figure this out, so I based this on url of center but there is not easy way to test the initial Url works right or not (since I can only test it with the expo prefix/link)
- Definitely open to feedback on how to tell the app that it was opened from a link - getStateFromPath with adding params is the easiest way I found to currently do it
- There are some centers that have special links - thinking on how we should handle it: hard code those cases in code since there are not too many? Just include the whole url link expect for obs Id in the links since that is all we are using those for? (likely bad idea once we expand) Make a separate obs specific dictionary that keeps that mapping? Open to other ideas! I added the special cases into schemas as comments for now.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -506,7 +506,7 @@ const BaseApp: React.FunctionComponent<{
       if (initialUrl) {
         // this url contains the whole url, like so: https://nwac.us/observations/#/observations/fb5bb19a-2b89-4c9c-91d2-eb673c5ab877
         const url = new URL(initialUrl);
-        return getStateFromPath(path + '?share=true&share_url=' + url.hostname + '/', opts);
+        return getStateFromPath(path + '?share=true&share_url=' + url.origin + '/', opts);
       }
     },
   };
--- a/App.tsx
+++ b/App.tsx
@@ -485,9 +485,8 @@ const BaseApp: React.FunctionComponent<{
     })
     .catch(() => logger.error('Getting initialUrl failed'));
 
-  const prefix = Linking.createURL('/');
   const linking = {
-    prefixes: [prefix, AvalancheCenterWebsites['NWAC'] + '/observations/#/view/'],
+    prefixes: [AvalancheCenterWebsites['NWAC'] + '/observations/#/view/'],
     config: {
       screens: {
         Observations: {
--- a/App.tsx
+++ b/App.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
 import {
   Lato_100Thin,
@@ -13,13 +13,13 @@ import {
   Lato_900Black_Italic,
   useFonts,
 } from '@expo-google-fonts/lato';
-import {Ionicons, MaterialCommunityIcons} from '@expo/vector-icons';
-import {SelectProvider} from '@mobile-reality/react-native-select-pro';
-import {createBottomTabNavigator} from '@react-navigation/bottom-tabs';
-import {NavigationContainer, RouteProp, useNavigationContainerRef} from '@react-navigation/native';
+import { Ionicons, MaterialCommunityIcons } from '@expo/vector-icons';
+import { SelectProvider } from '@mobile-reality/react-native-select-pro';
+import { createBottomTabNavigator } from '@react-navigation/bottom-tabs';
+import { getStateFromPath, NavigationContainer, RouteProp, useNavigationContainerRef } from '@react-navigation/native';
 import * as SplashScreen from 'expo-splash-screen';
-import {ActivityIndicator, AppState, AppStateStatus, Image, Platform, StatusBar, StyleSheet, UIManager, View} from 'react-native';
-import {SafeAreaProvider} from 'react-native-safe-area-context';
+import { ActivityIndicator, AppState, AppStateStatus, Image, Platform, StatusBar, StyleSheet, UIManager, View } from 'react-native';
+import { SafeAreaProvider } from 'react-native-safe-area-context';
 
 import * as Sentry from '@sentry/react-native';
 import * as Application from 'expo-application';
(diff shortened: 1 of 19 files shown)


App.tsx
```diff
@@ -474,6 +475,16 @@ const BaseApp: React.FunctionComponent<{
...
+  let initialUrl: string | null = null;
+  Linking.getInitialURL()
+    .then(url => {
+      initialUrl = url;
+    })
+    .catch(() => logger.error('Getting initialUrl failed'));
+
   const linking = {
     prefixes: [AvalancheCenterWebsites['NWAC'] + '/observations/#/view/'],
     config: {
```
stevekuznetsov, line 490: Adding this under `Observations` might fix your issue with not having the list:

https://reactnavigation.org/docs/configuring-links/#rendering-an-initial-route

yuliadub: you are right, if I add this --- then the Observation List screen is there if a universal link is open before app has been open/Obs page has been open. The back button will take you back to the list, not the home screen. If that is what we want - we can add that, lets check with Charlotte tomorrow

App.tsx
```diff
@@ -474,6 +475,22 @@ const BaseApp: React.FunctionComponent<{
...
+    .then(url => {
+      if (url) {
+        initialUrl = url;
+      }
+    })
+    .catch(err => {
+      logger.error('An error occurred while getting InitialUrl.', err);
+    });
+
+  Linking.addEventListener('url', event => {
```
yuliadub, line 490: I verified that this listened update the initial Url and things work as expected 😄 (did it in a hacky way, but you know, it was easy to test lol see screenshot) this also made it easy to see what the initialUrl had in it. 
IMG_1031
IMG_1032

components/observations/ObservationDetailView.tsx
```diff
@@ -309,6 +306,7 @@ export const ObservationCard: React.FunctionComponent<{
                       </Marker>
                     </ZoneMap>
                   )}
+                  <TableRow label="Avalanche Center" value={observation.center_id} />
```
yuliadub, line 309: I suggest we add this into the Obs page, regardless of logo in title/etc. I think this makes it easier to know what center the obs is shared from, in case you dont know what other center logos look like/etc since we do take you back to your default center. I am going to see what Charlotte thoughts are on this.

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -31,21 +31,21 @@ export const AvalancheCenterWebsites: Record<AvalancheCenterID, string> = {
   ['NWAC']: 'https://nwac.us/',
   ['BAC']: 'https://bridgeportavalanchecenter.org/',
   ['BTAC']: 'https://bridgertetonavalanchecenter.org/',
-  ['CBAC']: 'https://cbavalanchecenter.org/',
-  ['CNFAIC']: 'https://www.cnfaic.org/',
+  ['CBAC']: 'https://cbavalanchecenter.org/view-', // special link for obs sharing https://cbavalanchecenter.org/view-observations/#/view/observations/ddc9783a-a321-45cd-b656-88b066e5868d
```
yuliadub, line 34: still need your thoughts on what the best way to do this @stevekuznetsov  -- right now I am learning towards just making a file where I hardcode how to construct these to display the right obs link on the Observation page and just leave the links in here as the hostname. Figured that may just be easiest.

stevekuznetsov: React Navigation fundamentally does not allow for more than one route to be mapped to a specific screen, and that is exactly what we need to do. I unfortunately think that either:

- if `getStateFromPath()` gives you the entire path with the host name, we can write a function to transform any given URL for any center into the correct React Navigation screen state
- if it does not, we will need to drop down into the bare-bones listener approach.

App.tsx
stevekuznetsov: This is an interesting approach. It makes me queasy simply because I don't think we have any guarantee in what order the event loop handles your listener versus the `getStateFromPath()` call. The benefit of continuing to use React Navigation linking as opposed to having to hand-roll it might be large enough that the benefits here outweigh the cons, but fundamentally I do worry that out-of-order executions here would totally break links. You know how much we struggled with this sort of thing on the `onLayout` hacks to try to make the map react to the size of the header. 90% of the time, it works every time. The other 10%, it just doesn't and we have nothing we can do to fix it. When the outcome of that is that the map overlaps weirdly, whatever. When the outcome is that opening a link doesn't work, not sure if that's ok.

types/nationalAvalancheCenter/schemas.ts
yuliadub: ah good point, I was only really thinking of this in the context of building out the link - not the prefix part....let me think about that part a bit more

App.tsx
yuliadub: I think your concern is valid - although from testing/reading about this it seems like this is the suggested and trusted way to get the url. 🤷 Fair to say that we wont know when the 10% fail happens until we get a wider use of this.

I am not sure I get how we can use react navigation here: how do you see navigation linking providing the info that can get us center id? Aside from either having the center id in the url or querying NAC using the observation Id I dont see another way. Or after loading the obs page and then updating the logo in the header or something...or just not having the logo in the shared obs altogether.

components/observations/ObservationDetailView.tsx
stevekuznetsov: nit: we need the display name with `userFacingCenterId`

App.tsx
yuliadub: Charlotte to play with this and let us know if it should change or not

App.tsx
yuliadub: chatted about this offline - this works for just obs sharing, but gets way more complex when we will share forecasts, weather, etc. For now, we will keep as is and watch for errors --- but in future may take a different approach (i.e. not having linking be generated for us)

types/nationalAvalancheCenter/schemas.ts
yuliadub: decided that for now we will ignore prefixes and how we build them since none of the other centers are ready to have their obs open in the app anyways.

components/observations/ObservationUrlMapping.tsx
```diff
@@ -0,0 +1,29 @@
+import {AvalancheCenterID, AvalancheCenterWebsites} from 'types/nationalAvalancheCenter';
+
+export const GenerateObservationShareLink = (centerId: AvalancheCenterID, observationId: string) => {
+  let url: string | undefined = AvalancheCenterWebsites[centerId];
```
stevekuznetsov, line 4: `Record<AvalancheCenterID, string>` ensures at compile time that every center ID is in there, so I don't think this is truly `string | undefined` right?

components/content/NavigationHeader.tsx
```diff
@@ -7,18 +7,53 @@ import {Title1Black, Title3Black} from 'components/text';
...
-import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
+import {AvalancheCenterID, AvalancheCenterWebsites} from 'types/nationalAvalancheCenter';
 
 export const NavigationHeader: React.FunctionComponent<
   NativeStackHeaderProps & {
     center_id: AvalancheCenterID;
     large?: boolean;
   }
 > = ({navigation, route, options, back, center_id, large}) => {
+  let share: boolean = false;
```
stevekuznetsov, line 18: If it's easy, could you put the "back means reset everything" behavior in a separate commit so it's easy to `git revert` if/when we decide that we don't want that? Chris seemed to agree with you that the extra jank there would be jarring for users.

components/content/NavigationHeader.tsx
```diff
@@ -7,18 +7,53 @@ import {Title1Black, Title3Black} from 'components/text';
...
+  if (shareParams.share) {
+    share = true;
+    // if back is false, means the obs screens have not been open yet
+    if (!back) {
+      firstOpen = true;
+      // set back to not be null since we want a shared obs to have a back button
+      back = {title: 'Observations'};
+    }
+
+    const flip = (data: Record<AvalancheCenterID, string>) => Object.fromEntries(Object.entries(data).map(([key, value]) => [value, key]));
```
stevekuznetsov, line 32: We already have `reverseLookup` in the repo, can you use that?

stevekuznetsov: I *think* it is also typed correctly so you should not need `as AvalancheCenterID` on the output.

components/observations/ObservationUrlMapping.tsx
```diff
@@ -0,0 +1,29 @@
...
+    // special notes
+    // CNFAIC links their obs to https://www.cnfaic.org/observations/tincan-383/ but https://www.cnfaic.org/observations/#/view/observations/54831e07-3905-46a6-a6d4-de73d5496cd5
+    // also works (they just dont link to the widget anywhere on their site yet)
+    // HPAC is the same,  https://hpavalanche.org/observations/all-observations/ but also https://hpavalanche.org/observations/#/view/observations/07a02101-319f-4fd7-81d2-b82962886719
+    // TAC just wont work, not using AFP https://taosavalanchecenter.org/public-obs/20240621_lake-fork-ne-face/ - obs not loading in app at this time
+
+    url += 'observations/#/view/observations/' + observationId;
+  }
+
+  return url;
```
stevekuznetsov, line 28: `switch/case` and `const invalid: never` pattern ([like this](https://github.com/NWACus/avy/blob/2588ef37f74cafd117315786e1b7e940370906ad/components/AvalancheForecastZoneMap.tsx#L351)) make it so that the compiler ensures that every `centerId` input is covered, good thing to use to make sure we don't ever make any mistakes here.

components/content/NavigationHeader.tsx
yuliadub: still need it or it complains

components/content/NavigationHeader.tsx
yuliadub: because I changed the merge queue to be squash commit, having it in a different commit wont matter. But it is a small enough bit of code that if we do need to take it out, wont be too bad since just reverting it actually wont work - we will need to add initialRoute and make sure that doesnt break other stuff.

Issue #812, opened 2024-10-28
Expose All NAC Content Through Universal Links


stevekuznetsov:
Once we've got sharing figured out, we can support the map, forecast pages, weather data, etc, as links.

Depends on #809

PR #818, opened 2024-10-31
components: generic linking for all centers and pages

stevekuznetsov:
This commit adds a function to map between any avalanche center's URL and the correct set of (nested) screens on the app. Lots of testing to do.


App.tsx
```diff
@@ -492,21 +493,14 @@ const BaseApp: React.FunctionComponent<{
...
-          },
-        },
-      },
-    },
+    prefixes: [AvalancheCenterWebsites['NWAC']],
     getStateFromPath: (path: string, opts: {initialRouteName?: string; screens: PathConfigMap<object>} | undefined) => {
       if (initialUrl) {
-        // this url contains the whole url, like so: https://nwac.us/observations/#/observations/fb5bb19a-2b89-4c9c-91d2-eb673c5ab877
-        const url = new URL(initialUrl);
-        return getStateFromPath(path + '?share=true&share_url=' + url.origin + '/', opts);
```
yuliadub, line 509: the back button navigation depends on this - so if you are going to fully take this out, that logic will need to change

components/linking.ts
```diff
@@ -0,0 +1,473 @@
...
+
+type prefixes = {
+  avalanche?: string;
+  weatherStation?: string;
+  nwacWeatherStation?: string;
+  observations?: string;
+};
+
+// centers use the widgets, and the widgets manage state after some prefix in the URL; not ever center has the same prefix for every widget
+const prefixesForCenter = (avalancheCenter: AvalancheCenterID): prefixes => {
```
yuliadub, line 72: I think it would make sense for this to live in a separate file vs with all this linking

App.tsx
stevekuznetsov: Oh yeah I am hoping we don't go that route ... :crossed_fingers:

stevekuznetsov:
Fixes https://github.com/NWACus/avy/issues/812

components/linking.ts
```diff
@@ -0,0 +1,516 @@
+import {NavigationState, PartialState} from '@react-navigation/native';
+import {Logger} from 'browser-bunyan';
+import {TabNavigatorParamList} from 'routes';
+import {AvalancheCenterID, AvalancheCenterWebsites, reverseLookup} from 'types/nationalAvalancheCenter';
+
+export const getStateFromUrl = (logger: Logger, rawUrl: string): undefined | PartialState<NavigationState<TabNavigatorParamList>> => {
```
stevekuznetsov, line 6: TODO: on unknown, fall back to a new webview in the root stack that just shows the website, since there's no way to programatically reject some parts of the domain if we're registered for the whole thing

Issue #849, opened 2024-12-12
Filters not filtering in observations


charlotteguard:
The date filter does appear to be working but the zone, observer type and avalanches filters are not. you can select the filter option and then once you click apply nothing happens.

PR #850, opened 2024-12-12
observations: fix filtering

stevekuznetsov:
Looks like changes to react-hook-form also changed the allowable shapes that our filter config would have, which we did not capture in tsc.


stevekuznetsov:
Fixes #849

PR #870, opened 2025-01-02
components: handle expired forecasts more correctly

stevekuznetsov:
We were not showing the correct light-blue color for zones with information but no active forecast, nor were we capturing expired forecasts correctly on the map layer.

Map layer shows light blue for current information:
| Before | After |
| - | - |
| File (59) | File (52) |

Map layer shows grey for expired forecasts instead of their previous danger:
| Before | After |
| - | - |
| File (56) | File (53) |

Card hides expired danger rating, travel advice, and publication/expiry times for expired forecasts:
| Before | After |
| - | - |
| File (57) | File (55) |

Card still shows information for currently valid informative posts:
| Before | After |
| - | - |
| File (58) | File (54) |

### DIFF
--- a/components/AvalancheDangerTriangle.tsx
+++ b/components/AvalancheDangerTriangle.tsx
@@ -21,8 +21,9 @@ export const colorFor = (danger: DangerLevel | null | undefined): Color => {
       return Color('rgb(255, 242, 0)');
     case DangerLevel.Low:
       return Color('rgb(80, 184, 72)');
-    case DangerLevel.None:
     case DangerLevel.GeneralInformation:
+      return Color('rgb(110, 164, 219)');
+    case DangerLevel.None:
     default:
       return Color('rgb(147, 149, 152)');
   }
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
One question here w.r.t. the web view - in the past there was a grey polygon on the map for `'no rating'` zones with expired forecasts. Today, they are blue whether there's an active informational bulletin or an expired forecast. The API still shows the `'no rating'` danger level and grey color in the `"color"` field but the map shows "general information" and a blue color. Not sure what the appropriate thing to do here is and what users expect semantically between expired information and active bulletins.

```
$ curl -s https://api.avalanche.org/v2/public/products/map-layer/IPAC | jq '.features[] | select(.properties.name=="Selkirk Mountains") | .properties'
{
  "name": "Selkirk Mountains",
  "center": "Idaho Panhandle Avalanche Center",
  "center_link": "http://www.idahopanhandleavalanche.org/",
  "timezone": "America/Los_Angeles",
  "center_id": "IPAC",
  "state": "ID",
  "off_season": false,
  "travel_advice": "Watch for signs of unstable snow such as recent avalanches, cracking in the snow, and audible collapsing. Avoid traveling on or under similar slopes.",
  "danger": "no rating",
  "danger_level": -1,
  "color": "#888888",
  "stroke": "#104efb",
  "font_color": "#ffffff",
  "link": "https://www.idahopanhandleavalanche.org/forecasts/#/selkirk-mountains/",
  "start_date": null,
  "end_date": null,
  "fillOpacity": 0.5,
  "fillIncrement": 0.1,
  "warning": {
    "product": null
  }
}
```

Issue #141, opened 2023-02-14
App doesn't handle zones composed of multiple polygons (seen in CAIC)


floatplane:
looks like 3 zones missing? not a big v1 priority but would be good to sort out

app | [website](https://avalanche.state.co.us/)
--- | --- 
<img alt="image" src="https://user-images.githubusercontent.com/101196/218854859-f4235a9a-d95a-43ca-85a8-8c962ec1534d.png"> | <img alt="image" src="https://user-images.githubusercontent.com/101196/218854945-9969dcbe-34d0-413a-8413-89f65c2b5a3b.png">

floatplane:
Ah I think maybe we don't handle a zone that's got > 1 boundary

<img width="350" alt="image" src="https://user-images.githubusercontent.com/101196/218855709-e1ac6ee0-20bc-4d22-a16e-b2cc7989641b.png">

charlotteguard:
@floatplane the CAIC has dynamic zones (I don't totally know how it works but the change the area they're forecasting for) - could that be the issue?

stevekuznetsov:
The way they handle dynamic we actually do honor - they just flip ones enabled and disabled from what I can tell. I think the multiple polygons thing is probably right here.

floatplane:
agreed, I'll update the title of this issue

stevekuznetsov:
Looks like this hits Flathead now, we should go ahead and fix it.

PR #873, opened 2025-01-08
components/map: use GeoJSON map marker

stevekuznetsov:
Originally it must not have been clear that we could just shove the Feature for a zone into the GeoJSON marker and support all the varied types of geometries.

Fixes https://github.com/NWACus/avy/issues/141

### DIFF
--- a/components/map/AvalancheForecastZonePolygon.tsx
+++ b/components/map/AvalancheForecastZonePolygon.tsx
@@ -1,6 +1,6 @@
 import React, {useEffect, useRef} from 'react';
-import {Animated, NativeSyntheticEvent} from 'react-native';
-import {LatLng, Point, Polygon} from 'react-native-maps';
+import {Animated} from 'react-native';
+import {Geojson, GeojsonProps, LatLng} from 'react-native-maps';
 
 import {colorFor} from 'components/AvalancheDangerTriangle';
 import {MapViewZone} from 'components/content/ZoneMap';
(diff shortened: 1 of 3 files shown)


components/map/AvalancheForecastZonePolygon.tsx
```diff
@@ -59,20 +59,18 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
...
     zIndex: selected ? 1 : 0,
-    onPress: (event: PolygonPressEvent) => {
+    onPress: (_event: Parameters<Required<GeojsonProps>['onPress']>[0]) => {
       if (onPress) {
         onPress(zone);
-
-        // By calling stopPropagation, we prevent this event from getting passed to the MapView's onPress handler,
-        // which would then clear the selection
-        // https://github.com/react-native-maps/react-native-maps/issues/1132
-        event.stopPropagation();
```
stevekuznetsov, line 75: This new kind of event for the `onPress` on `<Geojson>` doesn't have `stopPropagation()` but from my testing it did not seem like we needed it.

components/map/AvalancheForecastZonePolygon.tsx
```diff
@@ -89,31 +87,10 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
...
+    return <AnimatedGeojson fillColor={fillColor} {...polygonProps} />;
   } else {
     const fillOpacity = renderFillColor ? zone.fillOpacity : 0;
     const fillColor = colorFor(zone.danger_level).alpha(fillOpacity).string();
-    return <Polygon fillColor={fillColor} {...polygonProps} />;
+    return <Geojson fillColor={fillColor} {...polygonProps} />;
   }
 };
-
-// need to update to react-native-maps 1.6 to get rid of this
```
stevekuznetsov, line 100: Note to future self: use `Parameters` and `React.ComponentProps` to extract types from libraries that don't export things.

PR #904, opened 2025-02-26
Issue 889: Add Observation Only Zones

### LINKED ISSUE #889: Adding Observation Zones to MWAC
MWAC had some zones added midseason to their Obs only. They are hoping the app can be updated to reflect what's happening on their website.

Image

wheninseattle:
## Description

This feature incorporates observation only zones into Avy; applying observation zone names to the observation preview cards and adding these zones to the filter widget. 

## Details

- Some centers (MWAC, BTAC, others?) have designated observation only zones outside of their official forecast zones. When available, a url to a KML file is provided in the avalanche center's meta data at: >> metaData.widget_config.observation_viewer.alternate_zones
- This url is passed to the useAlternateObserationZones hook which fetches and parses the KML file, returning features similar to those in the mapLayer.
- If provided, these alternate zones are merged with the features from the mapLayer prior to the flattening of the observations list, so that these zones labels are properly associated with the observations.
- **Optional** This mergedMapLayer is subsequently filtered by the available observations, removing zones that don't contain any observations. I could see this being a positive to reduce clutter in the already full filter widget, but on the other hand it could be advantageous for users to see the full list of zones even if they don't have observations.
- Finally, the resulting, filteredMapLayer is passed to the ObservationsFilterForm.

## Related Issue

Closes #889

## Acceptance Criteria

- [x]  When available, add observation only zones to observation list view and filter
- [x]  Observation preview cards, should be labeled with appropriate zone name, including observation only zones
- [x] When observation only zone is selected in the ObservationFilterForm, the list of observations should be correctly filtered for that zone.

## Considerations

- The map view on the [MWAC](https://www.mountwashingtonavalanchecenter.org/observations/#/view/observations) and [BTAC](https://bridgertetonavalanchecenter.org/observations/#/view/observations) websites does not show the observation only zones, so I decided to decouple the mapLayer and the observation only zones to match this. Additionally,  merging them into the mapLayer was breaking the map view. 
- image
- In the ObservationsFilterForm, do we want to visually separate "{Center Id} Forecast Zones" and "Other Regions" similar to the website form or create collapsible sections? If there are too many zones that push other sections off screen, will those other sections go unnoticed or be difficult to navigate to? 
- image

- As mentioned in "details", do we want to filter out zones that have no observations?
- Currently "unknown zone" still flashing briefly when observation preview cards initially render.
- As you can tell, Typescript is not my daily driver; I'm certainly open to any suggestions for how I can improve my typing!

### Before

- image
- image

### After

- image
- image
- image

## Testing Steps / QA Criteria

1.  Use `git pull wn-889-add-observation-zones`
2. `yarn start`
3. Set avalanche center to BTAC or MWAC
4. Check to ensure that observation preview cards are labeled with alternate zones when appropriate.
5. Check to ensure that the ObservationsFilterForm includes additional, observation only zones


components/observations/ObservationsFilterForm.tsx
```diff
@@ -323,12 +323,13 @@ export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterF
...
                   )}
                   {mapLayer && (
+                    // TODO: Render in two groups with respective labels: <Center_ID> Forecast Zones & Other Regions
                     <CheckboxSelectField
                       name="zones"
                       items={
                         initialFilterConfig.zones.length > 0
                           ? initialFilterConfig.zones.map(z => ({label: z, value: z}))
-                          : mapLayer.features.map(feature => ({label: feature.properties.name, value: feature.properties.name}))
+                          : [...mapLayer.features.map(feature => ({label: feature.properties.name, value: feature.properties.name}))]
```
stevekuznetsov, line 332: Why do we need the array destructuring here?

components/observations/ObservationsFilterForm.tsx
```diff
@@ -323,12 +323,13 @@ export const ObservationsFilterForm: React.FunctionComponent<ObservationsFilterF
                     </View>
                   )}
                   {mapLayer && (
+                    // TODO: Render in two groups with respective labels: <Center_ID> Forecast Zones & Other Regions
```
stevekuznetsov, line 326: @rustynwac let's pin down the UX we want for the two sets of zones with an eye to what the websites are doing.

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,28 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   const originalFilterConfig: ObservationFilterConfig = useMemo(() => createDefaultFilterConfig(additionalFilters), [additionalFilters]);
   const [filterConfig, setFilterConfig] = useState<ObservationFilterConfig>(originalFilterConfig);
   const [filterModalVisible, {set: setFilterModalVisible, on: showFilterModal}] = useToggle(false);
+  const {data: metaData} = useAvalancheCenterMetadata(center_id);
```
stevekuznetsov, line 77: Please keep the result pattern and add this to the `incompleteQueryState` block.

components/observations/ObservationsListView.tsx
```diff
@@ -298,7 +336,7 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
     [filterConfig, setFilterConfig],
   );
 
-  if (incompleteQueryState(observationsResult, mapResult) || !mapLayer) {
+  if (incompleteQueryState(observationsResult, mapResult) || !filteredMapLayer) {
```
stevekuznetsov, line 339: The check for `!mapLayer` is to guard against that request returning nothing - let's check here for the responses, not computed values from the responses, e.g. `|| !mapLayer || !centerMetadata || !alternateZones`

components/observations/ObservationsListView.tsx
```diff
@@ -171,7 +206,10 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
 
   // the displayed observations need to match all filters - for instance, if a user chooses a zone *and*
   // an observer type, we only show observations that match both of those at the same time
-  const resolvedFilters = useMemo(() => (mapLayer ? filtersForConfig(mapLayer, filterConfig, additionalFilters) : []), [mapLayer, filterConfig, additionalFilters]);
+  const resolvedFilters = useMemo(
```
stevekuznetsov, line 209: I think you need to run `yarn prettify` and/or configure your editor to use our prettier config for formatting diffs like this.

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+import axios, {AxiosError} from 'axios';
+import {safeFetch} from 'hooks/fetch';
+import {LoggerContext, LoggerProps} from 'loggerContext';
+import {useContext} from 'react';
+import {xml2json} from 'xml-js';
+interface KMLPoint {
+  coordinates: string;
+}
+
+interface KMLPlacemark {
```
stevekuznetsov, line 12: Please write a `zod` schema (or generate one from the interface here, or a known good KML library like [togeojson](https://www.npmjs.com/package/@tmcw/togeojson)) so we can parse and validate our incoming data.

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+  properties: {
+    name: string;
+  };
+}
+
+interface UseAlternateObservationZonesOptions {
+  enabled?: boolean;
+}
+
+export const useAlternateObservationZones = (url?: string, options?: UseAlternateObservationZonesOptions): UseQueryResult<AlternateObservationZones[] | null, AxiosError> => {
```
stevekuznetsov, line 67: `url === undefined` always means `options.enabled === false`. It would be a mistake to pass `useAlternateObservationZones(somethingReal, false)` - let's change the signature to `(url?: string)` and compute enabled directly as `!!url`.

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+  const {logger} = useContext<LoggerProps>(LoggerContext);
+  const key = ['alternateObservationZone-', url];
```
stevekuznetsov, line 69: Could we please mirror the pattern in other hooks to allow for pre-fetching?

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+  const thisLogger = logger.child({query: key});
+
+  return useQuery<AlternateObservationZones[] | null, AxiosError>({
+    queryKey: key,
+    queryFn: (): Promise<AlternateObservationZones[] | null> => fetchAlternateObservationZones(url!, thisLogger),
```
stevekuznetsov, line 74:
```suggestion
    queryFn: (): Promise<AlternateObservationZones[] | null> => fetchAlternateObservationZones(url, thisLogger),
```

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+    enabled: !!url && options?.enabled,
+    cacheTime: 60, // TODO: Change cache to one day after testing
+    staleTime: 60,
+    initialData: null,
+  });
+};
+
+export const fetchAlternateObservationZones = async (url: string, logger: LoggerProps['logger']): Promise<AlternateObservationZones[] | null> => {
```
stevekuznetsov, line 82:
```suggestion
export const fetchAlternateObservationZones = async (url?: string, logger: LoggerProps['logger']): Promise<AlternateObservationZones[] | null> => {
```

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const kmlData = JSON.parse(kmlDataJson) as KMLData;
+  if (!kmlData || !kmlData.kml || !kmlData.kml.Document || !kmlData.kml.Document.Folder || !kmlData.kml.Document.Folder.Placemark) {
+    thisLogger.error('Invalid KML data structure');
+    return null;
+  }
+
+  const alternateZones: AlternateObservationZones[] = kmlData.kml.Document.Folder.Placemark.map(placemark => {
+    let coordinates: [number, number][] = [];
+    const parseCoordinates = (coordinateString: string): [number, number][] => {
```
stevekuznetsov, line 103: This is a good candidate for a set of unit tests!

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,28 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
   const mapLayer = mapResult.data;
 
+  const mergedMapLayer = useMemo(() => {
+    if (alternateZones && mapLayer) {
+      const zonesNotInMapLayer: MapLayerFeature[] = (alternateZones as unknown as MapLayerFeature[]).filter(
+        (zone: MapLayerFeature) => !mapLayer.features.some((feature: MapLayerFeature) => feature.properties.name === zone.properties.name),
+      );
+      return {
+        ...mapLayer,
+        features: [...mapLayer.features, ...zonesNotInMapLayer],
```
stevekuznetsov, line 93: When we display observations to the user in card form, are we grabbing the first item from this list that matches for labelling? I think we should try to be explicit - do users expect to see an observation matching a custom zone and a forecast zone labelled with the former or the latter?

components/observations/ObservationsListView.tsx
```diff
@@ -136,10 +159,22 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
         .map(observation => ({
           ...observation,
-          zone: mapLayer ? matchesZone(mapLayer, observation.locationPoint.lat, observation.locationPoint.lng) : '',
+          zone: mergedMapLayer ? matchesZone(mergedMapLayer, observation.locationPoint.lat, observation.locationPoint.lng) : '',
         })),
-    [flatObservationList, mapLayer, displayNWACObservations],
+    [flatObservationList, mergedMapLayer, displayNWACObservations],
   );
+
+  const filteredMapLayer = useMemo(() => {
```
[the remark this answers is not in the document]
rustynwac: I don't think we should hide a zone if there is no data.

components/observations/ObservationsFilterForm.tsx
wheninseattle: Great question! This is a holdover from an earlier attempt of passing the alternate zones separately, and spreading the two arrays here.

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,28 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   const originalFilterConfig: ObservationFilterConfig = useMemo(() => createDefaultFilterConfig(additionalFilters), [additionalFilters]);
   const [filterConfig, setFilterConfig] = useState<ObservationFilterConfig>(originalFilterConfig);
   const [filterModalVisible, {set: setFilterModalVisible, on: showFilterModal}] = useToggle(false);
+  const {data: metaData} = useAvalancheCenterMetadata(center_id);
+  const alternateZonesUrl = metaData?.widget_config?.observation_viewer?.alternate_zones || '';
+  const {data: alternateZones} = useAlternateObservationZones(alternateZonesUrl, {
```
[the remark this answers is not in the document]
wheninseattle: Will do - that incompleteQueryState block is pretty sweet!

components/observations/ObservationsListView.tsx
wheninseattle: @stevekuznetsov, sorry, I'm not entirely sure what you mean here. Are you saying that we want to make sure that if there are multiple matches, we want to be explicit about prioritizing the forecast zone?

stevekuznetsov: I'm actually not sure what avalanche center staff would expect in this situation - I've reached out to the NAC and I'll follow up here once I hear back. However, regardless of which path is taken, I was hoping we could have explicit logic `(observation, forecast_zones, custom_zones) => label` that would encode the choice, as opposed to having implicit logic (return the first match from a list containing both types of zones that might or might not be sorted in a specific order).

stevekuznetsov: Got some clarity from the NAC, we should prioritize the custom zones if present.

wheninseattle: Thanks for getting clarification on that!

hooks/useAlternateObservationZones.ts
wheninseattle: @stevekuznetsov are you speaking specifically about the coordinate parsing function or the broader KML parsing functionality?

stevekuznetsov: Any of the logic here which takes static data as input and spits out other static data in some other format!

components/observations/ObservationsFilterForm.tsx
wheninseattle: @rustynwac @stevekuznetsov did we decide how we want to represent the zones?

hooks/useAlternateObservationZones.ts
wheninseattle: @stevekuznetsov I added a few unit tests for this. It looks like a recent package change may have caused a version conflict for "strip-ansi", causing the Jest tests to fail with "TypeError: stripAnsi is not a function". It looks like a dependency of "xml-js" may have changed the "strip-ansi" version.

I'm curious about the trade offs of breaking out smaller functions like parseCoordinates for testing. On the one hand this exposes them for individual unit test, but on the other they no longer have access to objects in the parent function (like logger).

hooks/useAlternateObservationZones.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+      [-121.7, 47.5],
+    ]);
+  });
+
+  it('should return an empty array for an empty coordinates string', () => {
+    const coordinatesString = '';
+    const result = parseCoordinates(coordinatesString);
+    expect(result).toEqual([]);
+  });
+});
```
stevekuznetsov, line 56: Were you able to write a test for `parseKmlData()`?

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+};
+
+function queryKey(url: string) {
+  return ['alternateZoneKML-', url];
+}
+
+export function parseKmlData(response: string): AlternateObservationZones {
+  const kmlDataJson = xml2json(response, {compact: true, spaces: 2});
+  const kmlData = JSON.parse(kmlDataJson) as KMLData;
```
stevekuznetsov, line 62: What we're doing in the rest of the codebase is using Zod to make sure that `JSON.parse()` spits out an object with the fields we're actually expecting. I think you already have a schema for `KMLData`, so we can `zod.safeParse()` the input and catch errors here instead of failing at runtime when we incorrectly access a field that we expected to exist but is missing.

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,104 @@
...
+  const placemarks = kmlData.kml.Document.Folder.Placemark;
+  const alternateZones: AlternateObservationZones = (Array.isArray(placemarks) ? placemarks : [placemarks]).map(placemark => {
+    let coordinates: number[][] = [];
+
+    const coordinateString = getCoordinateString(placemark);
+    coordinates = parseCoordinates(coordinateString);
+    const feature: AlternateObservationZones = {
+      type: 'Feature',
+      geometry: {
+        type: Array.isArray(coordinates[0][0]) ? 'MultiPolygon' : 'Polygon',
```
stevekuznetsov, line 72: I might be looking at an older version of this - did you decide if it would work to collapse this to `MultiPolygon` in every case?

types/nationalAvalancheCenter/schemas.ts
```diff
@@ -1427,3 +1427,81 @@ export const allAvalancheCenterCapabilitiesSchema = z.object({
...
+
+// Define a schema for KMLData
+export const KMLDataSchema = z.object({
+  kml: z.object({
+    Document: KMLDocumentSchema,
+  }),
+});
+export type KMLData = z.infer<typeof KMLDataSchema>;
+
+const AlternateObservationZonesSchema = z.union([
```
stevekuznetsov, line 1482: Ooops, sorry - finished the review before realizing you had not yet updated the code 🤦

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,26 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
   const originalFilterConfig: ObservationFilterConfig = useMemo(() => createDefaultFilterConfig(additionalFilters), [additionalFilters]);
   const [filterConfig, setFilterConfig] = useState<ObservationFilterConfig>(originalFilterConfig);
   const [filterModalVisible, {set: setFilterModalVisible, on: showFilterModal, off: hideFilterModal}] = useToggle(false);
+  const avalancheZoneMetadata = useAvalancheCenterMetadata(center_id);
```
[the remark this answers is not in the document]
wheninseattle: Oh, that is a smart convention. Added it!

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,26 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
+  const alternateZonesUrl: string = (avalancheZoneMetadata.data?.widget_config?.observation_viewer?.alternate_zones as string) || '';
```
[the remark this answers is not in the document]
wheninseattle: Oh, good point. Thanks!

components/observations/ObservationsListView.tsx
```diff
@@ -71,9 +74,26 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
+  const {data: alternateObservationZones} = useAlternateObservationZones(alternateZonesUrl);
+
   const mapResult = useMapLayer(center_id);
   const mapLayer = mapResult.data;
 
+  const mergedMapLayer = useMemo(() => {
+    if (alternateObservationZones && mapLayer) {
+      const zonesNotInMapLayer: AlternateObservationZones = alternateObservationZones.filter(
```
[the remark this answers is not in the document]
wheninseattle: Gosh, I went around and around on this; in the end, the only way I was able to prevent the mergedMapLayer from throwing a bunch of type errors, was to create a new type MergedMapLayer, that supports a union ` z.union([mapLayerFeatureSchema, observationZonesFeatureSchema]);`. Totally open to other solutions if I'm missing something!

hooks/useAlternateObservationZones.test.tsx
wheninseattle: Yes, I just committed them!

hooks/useAlternateObservationZones.ts
wheninseattle: This is a really cool capability, thanks for pointing me towards it; I added it here and in my transform function. One approach at getting the feature collections to merge involved populating the feature properties with default values to make them compatible with the mapLayer. For this the safeParse was super helpful!

types/nationalAvalancheCenter/schemas.ts
wheninseattle: Yeah, shoot - I should have let you know I was still grappling with the type issues. I've picked up the comments above, and committed all of my changes!

components/observations/ObservationsListView.tsx
```diff
@@ -71,8 +84,31 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
...
+  const alternateZonesUrl: string = avalancheZoneMetadataResult.data?.widget_config?.observation_viewer?.alternate_zones || '';
+  const alternateObservationZonesResult = useAlternateObservationZones(alternateZonesUrl, center_id);
+
   const mapResult = useMapLayer(center_id);
   const mapLayer = mapResult.data;
+  const observationOnlyZones = alternateObservationZonesResult.data;
+
+  const mergedMapLayer = useMemo((): MergedMapLayer => {
+    if (!mapLayer || !observationOnlyZones || !observationOnlyZones.features || observationOnlyZones.features.length === 0) {
+      return mapLayer as MergedMapLayer;
```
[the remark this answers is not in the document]
stevekuznetsov: (so, we should try to never use `as` and see what needs to be done to remove the uses that we have)

hooks/useAlternateObservationZones.test.tsx
stevekuznetsov: Awesome :sunglasses:

hooks/useAlternateObservationZones.ts
```diff
@@ -0,0 +1,164 @@
...
+export const useAlternateObservationZones = (url: string, center_id: AvalancheCenterID): UseQueryResult<ObservationZonesFeatureCollection, AxiosError> => {
+  const {logger} = useContext<LoggerProps>(LoggerContext);
+  const key = queryKey(url);
+  const thisLogger: Logger = logger.child({query: key});
+
+  return useQuery<ObservationZonesFeatureCollection, AxiosError>({
+    queryKey: key,
+    queryFn: (): Promise<ObservationZonesFeatureCollection> => fetchAlternateObservationZones(thisLogger, url, center_id),
+    enabled: !!url,
+    cacheTime: 60, // TODO: Change cache to one day after testing
```
stevekuznetsov, line 28: Couple TODOs left over?

stevekuznetsov (commented):
Looking awesome! Just a couple small comments and I think we're good to go!

components/observations/ObservationsListView.tsx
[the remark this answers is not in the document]
wheninseattle: Thanks - I'll file that away _as_ something to look out for in the future!

hooks/useAvalancheCenterMetadata.ts
```diff
@@ -25,6 +25,7 @@ export const useAvalancheCenterMetadata = (center_id: AvalancheCenterID): UseQue
   return useQuery<AvalancheCenter, AxiosError | ZodError>({
     queryKey: key,
     queryFn: async (): Promise<AvalancheCenter> => fetchAvalancheCenterMetadata(nationalAvalancheCenterHost, center_id, thisLogger),
+    enabled: !!center_id,
```
yuliadub, line 28: basically in the obs detail view, we do not have the center_id - we get it from the observation, which means there is a moment in time where it is null. trying to think what is the best way to handle this - should we handle the time when it is null vs trying to propogate the selected center_id into the obs details page? also possible I am missing something

yuliadub: @wheninseattle I already added it in, but I do think I want to get some context from Steve if that makes sense since I cant remember all the context there. I think at this time, if you want to just take a look and make sure I am not missing anything that'd be helpful but otherwise I think this should mostly be in a good place!

I decided to merge the expo updates and get a preview build out. Once we merge this in we can get another preview build going.

yuliadub: okay, Steve and I chatted about this and I am reverting the change to propogate center_id into the obs page. One of the main reasons is the possible mismatch of the center from observation vs the passed in center and no way to truly validate having just those two ids if the center is correct.

kevinherdez:
This PR got recreated in #1218 which has been checked in. Closing this PR

PR #905, opened 2025-02-28
Fix Modal Back Gestures

### LINKED ISSUE #640: Make sure Android back gesture is handled in all modals
We had some feedback that the swipe-to-go-back gesture was not working to exit our observation filter modals - does the approach you're using here fix that as well? If so - let's capture that fix in an issue to update the other modals as well!

_Originally posted by @stevekuznetsov in https://github.com/NWACus/avy/pull/638#discussion_r1430448845_

rossobrien:
### Description

Adds `onRequestClose` handlers to modals where it is missing

Closes #640

### Testing Steps
1. Open the Observations list and open the filter modal. You should be able to exit back to the list by swiping from the right edge of the screen
2. Open any forecast page and open the forecast zone dropdown. You should be able to close the dropdown by swiping from the right edge of the screen

### DIFF
--- a/components/content/Dropdown.tsx
+++ b/components/content/Dropdown.tsx
@@ -57,7 +57,7 @@ export const Dropdown: React.FC<DropdownProps> = ({items, selectedItem, onSelect
           </HStack>
         </View>
       </TouchableOpacity>
-      <Modal visible={dropdownVisible} transparent animationType="none">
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
Amazing work, thank you @rossobrien ! Looks great to me.

PR #936, opened 2025-04-10
Base Video Support

kevinherdez:
## Overview
This PR is the first of at least 4 PRs related to video support in the media viewer. It contains the base files for the new, custom media carousel shown when users click on an image in an observation. A couple of the files are largely copy and pasted from existing files (which I did to keep the code as separate as possible), and the bulk of the new code coming from the files related to `MediaViewModal`. I want to get feedback on it so that the future PRs are more focused and manageable.

## The Issue
The current media viewer relies on `react-native-image-viewing` to show the media images in a modal. It does not support displaying media that is not an image. 

There are other similar solutions to this problem, like react-native-zoom-toolkit, but they didn't support showing a WebView very well which is needed because we're displaying videos that are hosted on YouTube. 

Standard video viewing libraries also don't support YouTube videos which makes our scenario more unique.

## Solution
Create a custom media viewer that will allow us to be flexible with media types if anything changes in the future.

The media viewer is a FlatList that displays one item at a time, and conditionally shows an Image component or a WebView component depending on the media type

Right now, this only supports YouTube videos as I wasn't sure how to test other media types.

This solution also fixes the issue where the header and caption in the media view would be mismatched from what was shown. Images also look better on iPad, but only in portrait. I still need to fix landscape.

## Where to begin
`MediaCarousel.tsx` which is in the new `carouselV2` folder

## Testing
An observation from April 4th done at Steven's Pass (NWAC) by a forecaster has 3 images and 1 video

## Known Issues
Images can no longer be zoomed in on.

Swiping on a video to change items while it's playing doesn't work. This was also a bug while using YouTube video specific packages like `YoutubeiFrame`

## Future Plan
- The next PR will focus on bringing the image viewing back up to parity with what's currently there. This includes pinch/double tap to zoom
- PR after that will be hooking it up to all the carousel entry points behind a ramp
- Last PR will be removing the ramp
- Obviously this isn't counting any PRs for fixing bugs that come through testing/production

Demo: 
VideoSupportDemo

### DIFF
--- a/components/content/carousel/carouselV2/MediaViewerModal/MediaContentView.tsx
+++ b/components/content/carousel/carouselV2/MediaViewerModal/MediaContentView.tsx
@@ -2,7 +2,8 @@ import {ImageView} from 'components/content/carousel/carouselV2/MediaViewerModal
 import {WebVideoView} from 'components/content/carousel/carouselV2/MediaViewerModal/WebVideoView';
 import {View} from 'components/core';
 import {BodySm} from 'components/text';
-import React from 'react';
+import {usePostHog} from 'posthog-react-native';
+import React, {useEffect} from 'react';
 import {Dimensions} from 'react-native';
 import {MediaItem, MediaType} from 'types/nationalAvalancheCenter';
 
@@ -16,14 +17,25 @@ interface MediaContentProps {
 }
 
 export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({item, isVisible}: MediaContentProps) => {
+  const postHog = usePostHog();
+
   let content: React.JSX.Element;
+  let isMediaSupported = true;
 
   if (item.type === MediaType.Image) {
     content = <ImageView item={item} />;
   } else if (item.type === MediaType.Video) {
     content = <WebVideoView item={item} isVisible={isVisible} />;
   } else {
+    isMediaSupported = false;
     content = <BodySm>{'Unsupported Media Type'}</BodySm>;
   }
+
+  useEffect(() => {
+    if (postHog && !isMediaSupported) {
+      postHog.capture('mediaContentView-UnsupportedMedia', {mediaType: item.type});
+    }
+  }, [postHog, isMediaSupported, item]);
+
   return <View style={{width: SCREEN_WIDTH, height: SCREEN_HEIGHT}}>{content}</View>;
 };
--- a/components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
+++ b/components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
@@ -1,5 +1,6 @@
 import {View} from 'components/core';
 import {BodySm} from 'components/text';
+import {usePostHog} from 'posthog-react-native';
 import React, {useCallback, useEffect, useRef} from 'react';
 import {ActivityIndicator, Dimensions} from 'react-native';
 import WebView from 'react-native-webview';
@@ -14,10 +15,11 @@ interface WebVideoViewProps {
   isVisible: boolean;
 }
 
-const youtubeLink = (videoId: string) => `https://youtube.com/embed/${videoId}/?`;
+const youtubeLink = (videoId: string) => `https://youtube.com/embed/${videoId}`;
 
 export const WebVideoView: React.FunctionComponent<WebVideoViewProps> = ({item, isVisible}: WebVideoViewProps) => {
   const webRef = useRef<WebView>(null);
+  const postHog = usePostHog();
 
   useEffect(() => {
     if (!isVisible) {
@@ -37,6 +39,18 @@ export const WebVideoView: React.FunctionComponent<WebVideoViewProps> = ({item, 
       </View>
     );
   }, [maxHeight]);
+
+  useEffect(() => {
+    if (postHog) {
+      let properties: {[key: string]: string} = {};
+      if (typeof item.url === 'string') {
+        properties = {url: item.url};
+      } else if ('external_link' in item.url) {
+        properties = {url: item.url.external_link};
+      }
+      postHog.capture('webVideoView-UnsupportedVideoUrl', properties);
+    }
+  }, [postHog, item.url]);
 
   let source = '';
   if (typeof item.url !== 'string' && 'video_id' in item.url) {
--- a/components/screens/menu/DeveloperMenu.tsx
+++ b/components/screens/menu/DeveloperMenu.tsx
@@ -464,6 +464,15 @@ export const DeveloperMenu: React.FC<DeveloperMenuProps> = ({staging, setStaging
               },
             },
             {
+              label: '9: with video',
+              data: null,
+              action: () => {
+                navigation.navigate('observation', {
+                  id: '7b1f595d-312a-42f8-adb1-2f1886b7802b',
+                });
+              },
+            },
+            {
               label: 'NWAC pro observation with avalanches',
               data: null,
               action: () => {


components/content/carousel/carouselV2/ThumbnailList.tsx
kevinherdez, line 1: This file is also largely copy and pasted from `ImageList.tsx`. The biggest difference is parsing the `MediaItems` into `ThumbnailListItems` to more easily show both image and videos in the list

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
```diff
@@ -0,0 +1,23 @@
+import {View} from 'components/core';
+import {Image, useImage} from 'expo-image';
```
kevinherdez, line 2: I decided to bring in `expo-image` to make this a lot easier. This could also be done by refactoring `NetworkImage` which right now does a similar thing but only for thumbnail size images. If we'd rather not take on another package, I can work on refactoring `NetworkImage`

components/content/carousel/NetworkImage.tsx
```diff
@@ -53,8 +64,15 @@ export const NetworkImage: React.FC<NetworkImageProps> = ({uri, width, height, o
...
     image = (
-      <TouchableOpacity activeOpacity={0.8} onPress={onPressHandler} disabled={status !== 'success'}>
-        {image}
+      <TouchableOpacity activeOpacity={0.8} onPress={onPressHandler}>
+        <Center>
+          {image}
+          {showVideoIndicator && (
+            <View style={{zIndex: 1, position: 'absolute'}}>
+              <FontAwesome name="youtube-play" color={'red'} size={48} />
+            </View>
```
kevinherdez, line 73: Besides the color that I added for the modal background, this is the only change that would be live in prod. However, it will always be false so users should never see this until we launch the new media viewer

components/AvalancheCenterLogo.tsx
```diff
@@ -37,7 +37,7 @@ export const AvalancheCenterLogo: React.FunctionComponent<AvalancheCenterLogoPro
 
   const {data: uri} = useCachedImageURI(source[avalancheCenterId].uri);
   if (!uri) {
-    return <ActivityIndicator style={style} />;
+    return <ActivityIndicator />;
```
kevinherdez, line 40: This was throwing a typescript error because the `style` here is of type `ImageStyle`. The `style` seems to relate to the logo so I don't think it should be related to the `ActivityIndicator`, but if that's wrong I can find a better fix for it.

components/content/carousel/carouselV2/ThumbnailList.tsx
```diff
@@ -1,47 +1,98 @@
...
+    .map(item => {
+      if (item.type === MediaType.Image) {
+        return imageToThumbnailListItem(item);
+      } else if (item.type === MediaType.Video) {
+        return videoToThumbnailListItem(item);
+      }
+    })
+    .filter(item => item !== undefined);
+
+  return thumbnailItems;
```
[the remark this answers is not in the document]
kevinherdez: Typescript doesn't like the filter to remove `undefined` that I had before. I rewrote it to explicitly have `thumbnailItems` have a concrete type.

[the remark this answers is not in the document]
stevekuznetsov: I think only recent versions of `tsc` have learned how to infer type narrowing from functional pipelines - I've just run things through a good ol' fashioned `for` loop in cases like this ...

.gitignore
```diff
@@ -12,6 +12,7 @@ web-build/
 
 # macOS
 .DS_Store
+ios/
```
stevekuznetsov, line 15: FWIW you can also add this to your global `git config` with `git config --global core.excludesFile`  to add some file like `~/.gitignore` and not need to worry about updating every project. [ref](https://git-scm.com/docs/gitignore)

components/content/carousel/carouselV2/MediaViewerModal/MediaContentView.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({item, isVisible}: MediaContentProps) => {
+  let content: React.JSX.Element;
+
+  if (item.type === MediaType.Image) {
+    content = <ImageView item={item} />;
+  } else if (item.type === MediaType.Video) {
+    content = <WebVideoView item={item} isVisible={isVisible} />;
+  } else {
+    content = <BodySm>{'Unsupported Media Type'}</BodySm>;
```
stevekuznetsov, line 26: Perhaps in a follow-up, do you think we can add a PostHog event here to track what type it is, so we have some signal that we're not capturing what folks are actually trying to look at?

components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const SCREEN = Dimensions.get('screen');
+const SCREEN_HEIGHT = SCREEN.height;
+
+interface WebVideoViewProps {
+  item: VideoMediaItem;
+  isVisible: boolean;
+}
+
+const youtubeLink = (videoId: string) => `https://youtube.com/embed/${videoId}/?`;
```
stevekuznetsov, line 17: Are we adding query parameters at a later point? If not, probably OK to omit the `?` suffix

components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <ActivityIndicator size={'large'} />
+      </View>
+    );
+  }, [maxHeight]);
+
+  let source = '';
+  if (typeof item.url !== 'string' && 'video_id' in item.url) {
+    source = youtubeLink(item.url.video_id);
+  } else {
+    return <BodySm>{'There was an error loading the video'}</BodySm>;
```
stevekuznetsov, line 45: Similarly a good place to add error reporting in the future.

components/observations/ObservationDetailView.tsx
```diff
@@ -341,9 +342,9 @@ export const ObservationCard: React.FunctionComponent<{
                   )}
                 </VStack>
               </Card>
-              {images(observation.media).length > 0 && (
+              {(observation.media ?? []).length > 0 && (
```
stevekuznetsov, line 345: Could this be

```
{ observation.media && (
```

.gitignore
stevekuznetsov: (not something you need to change)

stevekuznetsov:
@kevinherdez are you using specific observations to test this out? Could we add some links to them in [here](https://github.com/NWACus/avy/blob/f2999d8cefbd33d3a3a5a7b71e48debefa5fa14d/components/screens/menu/DeveloperMenu.tsx#L389)?

components/AvalancheCenterLogo.tsx
```diff
@@ -2,11 +2,20 @@ import {QueryClient} from '@tanstack/react-query';
...
 import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
 
+interface AvalancheCenterLogoStyle {
+  resizeMode?: ImageResizeMode | undefined;
+  width?: DimensionValue | undefined;
+  height?: DimensionValue | undefined;
+  marginTop?: DimensionValue | undefined;
+  flex?: number | undefined;
+  flexGrow?: number | undefined;
+}
```
kevinherdez, line 15: `expo-image` seems to have changed the definition of `ImageStyle` which makes it not incompatible with the `ViewStyle` expected in the `ActivityIndicator`. This solution makes so the logo can only accept these styles, and then the `ImageStyle` specific style, `resizeMode`, is removed before passing it in to the `ActivityIndicator`

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
kevinherdez: Sounds good!

.gitignore
kevinherdez: Oh nice! I can't believe I didn't know about that before cause that's super helpful

components/content/carousel/carouselV2/MediaViewerModal/MediaContentView.tsx
kevinherdez: Yup I can go ahead and add that

components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
kevinherdez: I can remove it. To be honest, I might have just copy and pasted this without really thinking about it

components/observations/ObservationDetailView.tsx
kevinherdez: I was going to remove this line entirely before checking the code in so that it wouldn't be exposed in prod

stevekuznetsov:
Strong work @kevinherdez !!! Let's get this in

rustynwac:
@kevinherdez @yuliadub we had this working in Preview but in the most recent version, it looks like it no longer works on either device. Here's what I'm seeing. https://github.com/user-attachments/assets/a7ada286-b65a-4563-b325-cc4a12d4799f

PR #937, opened 2025-04-12
Fix Typescript Errors

kevinherdez:
This PR fixes Typescript errors that resulted in a build error in #936 

What's weird about these errors is that Typescript check in the CI has only failed in the above PR, but these changes have been in the codebase for months/years now. What's even weirder is that when I run the Typescript compiler script locally on `main` it fails, but in Github it's passing. We're going to investigate why the compiler is unreliable on Github.

### DIFF
--- a/components/AvalancheCenterLogo.tsx
+++ b/components/AvalancheCenterLogo.tsx
@@ -37,7 +37,7 @@ export const AvalancheCenterLogo: React.FunctionComponent<AvalancheCenterLogoPro
 
   const {data: uri} = useCachedImageURI(source[avalancheCenterId].uri);
   if (!uri) {
-    return <ActivityIndicator style={style} />;
+    return <ActivityIndicator />;
   }
 
   const images: Record<AvalancheCenterID, {(s: ImageStyle): ReactElement}> = {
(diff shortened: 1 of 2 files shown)


Rory-Sullivan:
@kevinherdez If there are discrepancies between running the TypeScript compiler locally and in CI that could mean that you have different versions of dependencies locally. Some dependencies were recently updated in the repo, have you updated those locally?

Also I know when I first cloned the repo I had issues with using npm to install dependencies, I would advise using yarn if you are not already. Yarn is used in CI so using it locally should ensure compatibility.

kevinherdez:
> @kevinherdez If there are discrepancies between running the TypeScript compiler locally and in CI that could mean that you have different versions of dependencies locally. Some dependencies were recently updated in the repo, have you updated those locally?
> 
> Also I know when I first cloned the repo I had issues with using npm to install dependencies, I would advise using yarn if you are not already. Yarn is used in CI so using it locally should ensure compatibility.

Sorry for the late reply here, and thanks! Steve was telling me that it's probably an issue with what I have locally, too. My inexperience with this led me to power through some of the issues I was having with `npm install`, and that's probably what put me in a bad state. I'll abandon this PR and give yarn a try to hopefully fix the other PR I have open.

Issue #777, opened 2024-08-24
Infinite Scroll Broken When First Data Doesn't Match Filters


stevekuznetsov:
Right now, `fetchMoreData()` in the `ObservationsListView` keeps fetching more pages until *some* data is returned. However, when a filter or filters are enabled, if we're able to get *some* data but none of it matches the filter, a user will see an infinite loading bug. We would be better off checking to see if there's any data being displayed, but need to track "more" data being displayed for when we trigger the fetching due to user scroll behavior.

yuliadub:
@stevekuznetsov I feel like this has been for the most part fixed so closing. reopen if you think otherwise

PR #942, opened 2025-04-19
Fixes observations list view

### LINKED ISSUE #871: Touch Interactions on Map Cards Ignored on Android
On Android, when the user opens a card by tapping on a forecast zone on the map, the card is not responsive to touch inputs. In the debugger we see the `<TouchableOpacity>` exists and has the correct `onPress` handler, but interactions don't trigger any of that behavior - nothing we put into the `onPress` fires, and we don't even see the opacity change. If the user swipes to a new card, there's some high likelihood that the card interactions work correctly again, but it's not 100%. Tapping on the zone another time always brings up the forecast. It seems like we might have a z-index issue or something when programatically scrolling the drawer, but I don't know how to debug this. The obvious culprit is our bump to React Native 0.76 and Expo SDK 52 in https://github.com/NWACus/avy/commit/dea4ccc706a7f5de871509896041d558fe4dd113.

The mapping library we're using does not yet support the New Architecture (see https://github.com/react-native-maps/react-native-maps/issues/5236) so we've turned off the New Architecture. It's not clear to me if turning this feature off while remaining on the new React Native version would still trigger some of the pathological behavior here.

https://github.com/user-attachments/assets/e6f045f3-9cf3-4265-a274-09497ac528e9

Rory-Sullivan:
Fixes handling of case when there is no items in the list. Should display a loading spinner if the app is currently fetching data, otherwise displays a warning about filtering out all data. Fixes issue #777.

image

Fixes display of list footer. Should display a loading spinner if the app is currently fetching more data, otherwise the number of observations in the list. Fixes #871.

image

### DIFF
--- a/components/observations/ObservationsListView.tsx
+++ b/components/observations/ObservationsListView.tsx
@@ -167,8 +167,6 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
     })();
   }, [observationsResult]);
 
-  const moreDataAvailable = observationsResult.hasNextPage;
-
   // the displayed observations need to match all filters - for instance, if a user chooses a zone *and*
   // an observer type, we only show observations that match both of those at the same time
   const resolvedFilters = useMemo(() => (mapLayer ? filtersForConfig(mapLayer, filterConfig, additionalFilters) : []), [mapLayer, filterConfig, additionalFilters]);
@@ -251,36 +249,30 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
     [],
   );
   const renderListFooter = useCallback(() => {
-    if (!moreDataAvailable) {
-      if (observationsSection.data.length === 0) {
-        return null;
-      }
-      return (
-        <Center height={OBSERVATION_SUMMARY_CARD_HEIGHT} paddingBottom={48}>
-          <Body>
-            <FormattedMessage
-              description="How many observations were found."
-              defaultMessage="{count, plural,
-  =0 {No observations match the filters.}
-  one {One observation matches the filters.}
-  other {# observations match the filters.}}"
-              values={{
-                count: observationsSection.data.length,
-              }}
-            />
-          </Body>
-        </Center>
-      );
-    } else if (observations.length > 0 && observationsResult.isFetchingNextPage) {
+    if (observationsResult.isFetchingNextPage) {
       return (
         <Center height={OBSERVATION_SUMMARY_CARD_HEIGHT} paddingBottom={48}>
           <ActivityIndicator size={'large'} color={colorLookup('text.secondary')} />
         </Center>
       );
-    } else {
-      return <View height={OBSERVATION_SUMMARY_CARD_HEIGHT} />;
     }
-  }, [moreDataAvailable, observationsSection.data.length, observations.length, observationsResult.isFetchingNextPage]);
+    return (
+      <Center height={OBSERVATION_SUMMARY_CARD_HEIGHT} paddingBottom={48}>
+        <Body>
+          <FormattedMessage
+            description="How many observations were found."
+            defaultMessage="{count, plural,
+  =0 {No observations match the filters.}
+  one {One observation matches the filters.}
+  other {# observations match the filters.}}"
+            values={{
+              count: observationsSection.data.length,
+            }}
+          />
+        </Body>
+      </Center>
+    );
+  }, [observationsSection.data.length, observationsResult.isFetchingNextPage]);
   const renderSectionHeader = useCallback(
     ({section: {title}}: {section: {title: string}}) =>
       hasPendingObservations ? (
@@ -376,7 +368,7 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
         onEndReached={fetchMoreData}
         ListFooterComponent={renderListFooter}
         ListEmptyComponent={
-          moreDataAvailable ? (
+          observationsResult.isFetching ? (
             <Center height={'100%'}>
               <VStack alignItems="center" space={8}>
                 <ActivityIndicator size={'large'} color={colorLookup('text.secondary')} />


Rory-Sullivan:
So as best I can tell in the current version of the app if you try to filter for data behind the first page that data never actually loads. For example if you filter for data in November 2024 it never loads in despite their being valid observations for that time frame (unless you first scroll back that far without filters and then filter). Unless there is something I'm missing here?

In the current version it will show the spinner forever but as best I can tell the app is not actually loading in any more data. This update I think correctly shows this to the user in that it stops showing the spinner to the user when the app stops fetching data and just says there is no data for the filter conditions.

I think there is a separate piece of work here to make it so that the app does load in data for filters past the initial page. I'm wondering if the user sets a date range if we should be passing that into our API query, to skip the step of loading in all of the data between now and whatever date range they set?

I also feel like if the user sets a query like the 'Olympics, professional' one that you described above, we should show them a message that says 'there is no observations that match those filters in the last x weeks' and then give them the option to go further back and load more data. I think this fits in well with what you said above;

> Or maybe we really do want fetchMoreData() to bail out when we get the first data at all - and then we let the proximity of the viewport to the end of the list trigger the next fetches?

I do think that showing a spinner forever is a bad user experience and definitely something we should avoid. What do you think about these ideas?

stevekuznetsov:
>In the current version it will show the spinner forever but as best I can tell the app is not actually loading in any more data. 

If you run with debug logging turned on, can you confirm that? My understanding was that `fetchMoreData()` determined if we would continue to run backwards in the time set, and explicitly did the calls with `const pageResult = await fetchNextPage();`. The intended path was that we'd put _something_ on screen by walking backwards until we found data to show, then let the built-in integration with the list take over, so more data would be fetched if/when the user scrolled close to the bottom of the list.

> I'm wondering if the user sets a date range if we should be passing that into our API query, to skip the step of loading in all of the data between now and whatever date range they set?

There's some thorny UX questions here about how to do an infinitely-scrolling backwards-looking feed while enforcing edges. I'd want to look at what other experiences exist and (or if) they handle explicit date ranges - do you know?

> I also feel like if the user sets a query like the 'Olympics, professional' one that you described above, we should show them a message that says 'there is no observations that match those filters in the last x weeks' and then give them the option to go further back and load more data. 

This seems reasonable - we'd want to somehow add more state to the list such that we could track "we did N fetches, instead of fetching more because of internal logic, we want the user to make an explicit choice to opt into more data" so we're not just firing off requests forever. This snarls up the "fetch more data before the user is able to scroll to the bottom of the list so they get a seamless experience" UX that we're getting from `useInfiniteQuery` but I think we might be able to provide the user a nice middle ground specifically in the case where we're churning through queries on their behalf more than we think we ought to.

Rory-Sullivan:
> If you run with debug logging turned on, can you confirm that?

Yes I was able to confirm this with logging today. In the while loop inside `fetchMoreData` there is a break statement;

```javascript
const fetchCount = pageResult.data?.pages[pageResult.data?.pages.length - 1].data?.length ?? 0;
if (fetchCount > 0) {
  break;
}
```

I confirmed that this `fetchCount` is the count of _unfiltered_ observations. Hence this exits after any data is returned not necessarily after there is some data to show to the user. From what you're saying this is not the intended behavior?

> I'd want to look at what other experiences exist and (or if) they handle explicit date ranges - do you know?

I have done something similar to this before but it was kind of a different use case. Any time the user changed a date range we would just discard all the data we had and fetch from the backend server. In this case I feel like we would want to keep track of what ranges we have fetched and then fill in gaps as the user scrolls or something like that, I feel like that could get tricky to handle though.

>  This snarls up the "fetch more data before the user is able to scroll to the bottom of the list so they get a seamless experience" UX that we're getting from useInfiniteQuery but I think we might be able to provide the user a nice middle ground specifically in the case where we're churning through queries on their behalf more than we think we ought to.

Yeah I do think this is really a question of what the intended UX is, and I don't feel like I can really answer that question. One thing I would say though is that if users are using the app in the backcountry, I think there is a good chance they will not have great reception (if any) and so I would lean towards giving them more control of what requests we are making.

On a similar note to the above, at the moment I don't believe the app shows request failures to the user in anyway. Something that I also think is quite likely in the backcountry. Perhaps this is a separate issue though.

Rory-Sullivan:
>If we can keep the existing UX (fetch backwards until we have something to show you), though, I would prefer that I think. WDYT?

Yeah I agree with this. I have a much better understanding of this now. I will make the adjustment so that we keep fetching data until we have something to show the user post filtering. I will probably still put some arbitrarily high stopping condition so that we don't actually fetch forever. Then I think we should play around with it in preview and make a call on if we want to try anything else we have been talking about here, or if we like the behavior as is.

Rory-Sullivan:
I have just pushed a commit with the above changes. So `fetchMoreData` takes filters into account when counting the returned data. I set an arbitrary date limit of 6 months in the past to stop the query running forever, I figure that should cover everything in a season.

Something I noticed when testing this without the date limit, is that if you set filters that don't return anything this query will really keep running forever. This includes if you change your filters or navigate away from the observations page or the app. Force closing the app was the only way I found of stopping the query in this case.

components/observations/ObservationsListView.tsx
```diff
@@ -151,27 +157,33 @@ export const ObservationsListView: React.FunctionComponent<ObservationsListViewP
         return;
       }
 
+      // Date limit is set to 6 months in the past, 1.5552e10 is 180 days in milliseconds
```
stevekuznetsov, line 160: The one thing to be wary of is pre-season testing (looking at the app in November, often folks are doing this to validate things before the season starts and need to see the latest data which could be from April). Let's set this to ~1y or something and use the `date-fns` [helper](https://date-fns.org/v4.1.0/docs/sub) instead of magic numbers.

stevekuznetsov:
> This includes if you change your filters or navigate away from the observations page or the app. Force closing the app was the only way I found of stopping the query in this case.

Hmm - up to you on whether or not you tackle it in this PR or not, but I think we should be able to handle this with an explicit [cancellation](https://tanstack.com/query/v3/docs/framework/react/guides/query-cancellation).

stevekuznetsov:
Dang, that's a real shame. We might be in a better-enough state to pull this in as-is, I'll give it a review. Perhaps there is some other way to trigger when it's not big enough to scroll.

PR #950, opened 2025-06-19
Add gesture Support to ImageView.tsx

kevinherdez:
### Overview

This adds gesture support to `ImageView.tsx` to bring the custom MediaViewer functionality back to what it was with the old  image viewer. Three gestures are added to the view: pinch, pan, and double tap.

These gestures use the `react-native-gesture-handler` package while the animations use `react-native-reanimated`. These packages are built to work together, and they were packages that are already being included in the repo.

The design decisions I made with passing in an optional native gesture and an option fullScreenWidth into the `ImageView` was to hopefully make it as reusable as possible.

### Overview of `react-native-gesture-handler`

I decided to go with this package over the built in `GestureResponder` that's provided by React Native because `react-native-gesture-handler` has better performance and platform support. It's built to use the iOS/Android native touches and runs the gesture/animation code at the native level, instead of `GestureResponder` which runs at the Javascript level. This leads to less frame dropping, and code that works on both platforms with little to no need to write platform specific code to handle the gestures.

#### Explanation of Edge Calculation

The equation in `ImageView.tsx` on line 80 pulls the image back to the edge of the screen after the gesture completes. This is based on the observation that no matter what the scale of the image is if the `translationX.value` is equal to half of the screen width, then the edge of the image will be in the middle of the screen. This seems to be true for both iOS and Android. Using this I calculate the edge of the screen as follows:

`screen_edge = 0.5 * screen_width - ((0.5 * screen_width) / scale.value)`

By setting the `translateX.value` to this, the image goes back to the edge of the screen

### Where to Start
- `ImageView.tsx`

### Testing

Tested on both the iOS simulator and Android emulator. Since we still support Expo 52 I wasn't able to try it on Expo Go on device. 

GestureSupport

### DIFF
--- a/components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
+++ b/components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
@@ -1,15 +1,113 @@
 import {View} from 'components/core';
 import {Image, useImage} from 'expo-image';
-import React from 'react';
+import React, {useState} from 'react';
 import {ActivityIndicator} from 'react-native';
+import {
+  Gesture,
+  GestureDetector,
+  GestureStateChangeEvent,
+  GestureUpdateEvent,
+  NativeGesture,
+  PanGestureHandlerEventPayload,
+  PinchGestureHandlerEventPayload,
+} from 'react-native-gesture-handler';
+import Animated, {runOnJS, useAnimatedStyle, useSharedValue, withTiming} from 'react-native-reanimated';
 import {ImageMediaItem} from 'types/nationalAvalancheCenter';
 
 interface ImageViewProps {
   item: ImageMediaItem;
+  nativeGesture: NativeGesture | null;
+  fullScreenWidth: number | null;
 }
 
-export const ImageView: React.FunctionComponent<ImageViewProps> = ({item}: ImageViewProps) => {
+const MAX_SCALE = 8.0;
+const MIN_SCALE = 1.0;
+const DOUBLE_TAP_SCALE = 1.5;
+const RESET_TRANSLATION_VALUE = 0.0;
+
+export const ImageView: React.FunctionComponent<ImageViewProps> = ({item, nativeGesture, fullScreenWidth}) => {
   const image = useImage(item.url.original);
+
+  const scale = useSharedValue(1);
+  const startScale = useSharedValue(0);
+
+  const translateX = useSharedValue(0);
+  const translateY = useSharedValue(0);
+  const prevTranslateX = useSharedValue(0);
+  const prevTranslateY = useSharedValue(0);
+
+  const [panEnabled, setPanEnabled] = useState(false);
+
+  const setPanOnJS = (value: boolean) => {
+    setPanEnabled(value);
+  };
+
+  const pinchGesture = Gesture.Pinch()
+    .onStart(() => {
+      startScale.value = scale.value;
+    })
+    .onUpdate((event: GestureUpdateEvent<PinchGestureHandlerEventPayload>) => {
+      scale.value = startScale.value * event.scale;
+    })
+    .onEnd(() => {
+      if (scale.value < MIN_SCALE) {
+        scale.value = withTiming(MIN_SCALE);
+        translateX.value = withTiming(RESET_TRANSLATION_VALUE);
+        translateY.value = withTiming(RESET_TRANSLATION_VALUE);
+        runOnJS(setPanOnJS)(false);
+        return;
+      }
+
+      if (scale.value > MAX_SCALE) {
+        scale.value = withTiming(MAX_SCALE);
+      }
+
+      runOnJS(setPanOnJS)(true);
+    });
+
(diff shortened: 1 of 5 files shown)


components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
```diff
@@ -1,16 +1,114 @@
...
+  const translateX = useSharedValue(0);
+  const translateY = useSharedValue(0);
+  const prevTranslateX = useSharedValue(0);
+  const prevTranslateY = useSharedValue(0);
+
+  const [panEnabled, setPanEnabled] = useState(false);
+
+  const setPanOnJS = (value: boolean) => {
+    setPanEnabled(value);
+  };
```
kevinherdez, line 43: This is needed to correctly set the React Native state back on the JS thread instead of directly on the UI thread

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
```diff
@@ -1,16 +1,114 @@
...
+    .onUpdate((event: GestureUpdateEvent<PinchGestureHandlerEventPayload>) => {
+      scale.value = startScale.value * event.scale;
+    })
+    .onEnd(() => {
+      if (scale.value < MIN_SCALE) {
+        scale.value = withTiming(MIN_SCALE);
+        translateX.value = withTiming(RESET_TRANSLATION_VALUE);
+        translateY.value = withTiming(RESET_TRANSLATION_VALUE);
+        runOnJS(setPanOnJS)(false);
+        return;
```
kevinherdez, line 58: I was having a hard time breaking this out into a function since the code in here is run directly on the UI thread and that seems to cause issues when the function is defined in TS. I can keep trying to find a way to do this if we feel strongly about breaking out this repeated code into a function.

yuliadub: out of my random reading here is a way to solve it - I tired it with "worklet" and it seems to work! let me know if that does the trick on your end too (i only tried in expo go)

`const resetTransform = () => {
  "worklet";
  scale.value = withTiming(MIN_SCALE);
  translateX.value = withTiming(RESET_TRANSLATION_VALUE);
  translateY.value = withTiming(RESET_TRANSLATION_VALUE);
  runOnJS(setPanEnabled)(false);
};`

yuliadub:
@kevinherdez  Looks good! Thanks for bringing these gestures back 😄 

> Since we still support Expo 52 I wasn't able to try it on Expo Go on device.

I was able to run this inside expo go on an emulator -- so I am a bit confused by this note. Can you please elaborate?

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
```diff
@@ -1,16 +1,114 @@
...
+      prevTranslateX.value = translateX.value;
+      prevTranslateY.value = translateY.value;
+    })
+    .onUpdate((event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
+      translateX.value = prevTranslateX.value + event.translationX / (1.5 * scale.value);
+      translateY.value = prevTranslateY.value + event.translationY / (1.5 * scale.value);
+    })
+    .onEnd(() => {
+      if (fullScreenWidth != null) {
+        const maxXTranslate = 0.5 * fullScreenWidth - (0.5 * fullScreenWidth) / scale.value;
```
yuliadub, line 79: trying to think if this can ever fail for some weird reason and I dont think so -- but just writing my thoughts out, like the image is too small (or too big) and not on the edges to begin with (that is not really possible since we pretty much will stretch it to the edges no matter what size it is, I think. I wonder what happens if someone uploads a pano shot? ), or what if the image is not scaled to center (again, this shouldn't happen) or if after many many pans it could get confused for some reason. I played around with this and couldnt get it to freak out inside expo go but would be interesting to do more testing on actual device

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
kevinherdez: Oh nice! I saw that that might work, but I think I was messing up the syntax or something because it wasn't working as expected when I tried it. I'll try it out and push a new change

components/content/carousel/carouselV2/MediaViewerModal/ImageView.tsx
kevinherdez: I wish I knew more about how the relationship between the scale and the panning worked so that we knew for sure that this would be true no matter what for fullscreen images that are scaled to fit in the view. I tried it on different screen sizes and both Android/iOS and it seems to hold up, but I agree I think we need to test on different devices to be sure.

For smaller images (not fullscreen) I added the if check to prevent the snapping from happening because those images no longer hold up to the assumptions made in this code. Hopefully, that prevents weird things from happening at least in certain cases.

PR #952, opened 2025-06-23
WeatherStationsDetail: handle unknown fields

stevekuznetsov:
We were previously not handling unknown field names, so we'd crash when we got new field types from the SnowObs API. This commit handles the unknown field case and adds some short names for battery voltage.


rustynwac:
Thanks Steve!

PR #955, opened 2025-08-11
package.json: bump to Expo 53.0.20

yuliadub:
Updating expo and dependecies. Ran into issues with md5 module not being found but after a clean install of all packages everything booted as expected.


yuliadub:
looks like CI is unhappy, looking into it

stevekuznetsov:
If you can, don't blow away node_modules and re-install, it will change so many dependencies it's very hard to guarantee that we didn't regress things. In the past we've introduced a dozen bugs that way :|

yuliadub:
> > If you can, don't blow away node_modules and re-install, it will change so many dependencies it's very hard to guarantee that we didn't regress things. In the past we've introduced a dozen bugs that way :|
> 
> good to know! I will try again without blowing it away

@stevekuznetsov  ok, not blowing away yarn.lock but the changes still look massive. Wonder if that is because I am updating a lot of packages? Seems like that is good to do since we are behind on a few. There are at least 2 packages I found where the newer version will require some code refactor, I will make issue for those but right now upped them to the version that still works with code as is. 

still need to fix yarn test

yuliadub:
ok. to follow up on this - something weird is happening with one of my Windows machines that is running WSL. When I went to run yarn install on my different windows machine, the changes in yarn.lock look a lot more reasonable and what we would expect. Still need to figure out why the tests are failing but getting close!

yuliadub:
alright, got all this resolved but now if you look at the expo go preview - when clicking into a map zone, looks like the app crashes. Gonna keep debugging whats up (the map doesnt load on the emulator on my machine so gotta get creative 😄 )

kevinherdez:
> alright, got all this resolved but now if you look at the expo go preview - when clicking into a map zone, looks like the app crashes. Gonna keep debugging whats up (the map doesnt load on the emulator on my machine so gotta get creative 😄 )

Was that testing on iPhone or Android? I tested on iPhone real quick and I hit weird issues where random zones would disappear when clicking on other ones. Eventually it crashed but it took a bit. I'll see if I can't get them to load locally to help investigate.

IMG_0605

yuliadub:
> The "new architecture" breaks the map entirely - IIRC in Expo Go, they always run the new architecture, regardless of whether your project's config has it turned on or not. So, I think that's expected behavior and won't happen in the preview app.

well its actually interesting - on an android emulator inside expo go no maps loads for me period. When running expo preview here from the PR, with the app also opening in expo go, the map does load and it crashes when I go to it, so I do think this is a real crash (plus Kevin did confirm that he is seeing this on iOS as well)

components/content/NavigationHeader.tsx
```diff
@@ -28,7 +28,7 @@ export const NavigationHeader: React.FunctionComponent<
     if (!back) {
       firstOpen = true;
       // set back to not be null since we want a shared obs to have a back button
-      back = {title: 'Observations'};
+      back = {title: 'Observations', href: undefined};
```
yuliadub, line 31: need to look into what href should be

stevekuznetsov:
Sure, of course chase real breaks, but keep that in mind since you will get weird behavior with the zones, etc, and there's nothing to be done.

kevinherdez:
> The "new architecture" breaks the map entirely - IIRC in Expo Go, they always run the new architecture, regardless of whether your project's config has it turned on or not. So, I think that's expected behavior and won't happen in the preview app.

Ah yeah I had forgot that you mentioned that awhile ago. Good callout

PR #990, opened 2025-11-04
Fix Video Viewer

kevinherdez:
It looks like something changed recently where Youtube started requiring that the referer header property needs to be defined for embedded videos. I tested it on both the Android emulator and the iOS simulator and this seems to work.

**NOTE:** While fixing this, I found another bug where the index of the header and footer on Android isn't correct when the modal initially loads. As far as I can tell, it's not updating the state with the correct initial value when the modal is shown and then it fixes itself when the user swipes to a different image. This doesn't repo in Expo 53 which is strange. I'll keep looking into it, but I didn't want that to hold up getting this fix out

### DIFF
--- a/components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
+++ b/components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
@@ -1,9 +1,11 @@
-import {View} from 'components/core';
+import {Center, View} from 'components/core';
 import {BodySm} from 'components/text';
+import Constants from 'expo-constants';
 import {usePostHog} from 'posthog-react-native';
 import React, {useCallback, useEffect, useRef} from 'react';
-import {ActivityIndicator, Dimensions} from 'react-native';
+import {ActivityIndicator, Dimensions, Platform} from 'react-native';
 import WebView from 'react-native-webview';
+import {WebViewSource} from 'react-native-webview/lib/WebViewTypes';
 import {colorLookup} from 'theme';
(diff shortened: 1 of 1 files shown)


components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
```diff
@@ -16,6 +16,7 @@ interface WebVideoViewProps {
 }
 
 const youtubeLink = (videoId: string) => `https://youtube.com/embed/${videoId}`;
+const headers = {Referer: 'https://us.nwac.forecast'};
```
stevekuznetsov: Does this need to be set to a valid domain, or something specific to the app? If you wanted this exact string you could use the Expo configuration package - we have access to identifiers there

components/content/carousel/carouselV2/MediaViewerModal/WebVideoView.tsx
```diff
@@ -15,7 +17,37 @@ interface WebVideoViewProps {
...
+
+const getBundleID = () => {
+  if (Platform.OS === 'ios') {
+    return Constants.expoConfig?.ios?.bundleIdentifier;
+  } else if (Platform.OS === 'android') {
+    return Constants.expoConfig?.android?.package;
+  } else {
+    return undefined;
+  }
+};
```
kevinherdez: Is there a more general spot to put this that's already created, or should I make a file in hooks called `expoConfigHelpers` or something to put this in? As of right now, it's not needed anywhere else, but maybe it'll be helpful in the future.

stevekuznetsov (approved):

Issue #1004, opened 2025-11-19
Add HPAC to Prod


rustynwac:
We might need to do something funky here for the weather station data. I don't see it in Preview but they have a widget with it on their website. 

About HPAC [https://hpavalanche.org/about/]
Donate [https://hpavalanche.org/donate/]
Upcoming Events [https://hpavalanche.org/about/events/]
Scholarships [https://hpavalanche.org/about/scholarship-fund/]
Maps and Resources [https://hpavalanche.org/about/resources/]
Monthly Newsletter [https://mailchi.mp/hpavalanche/news]

kevinherdez:
Sounds good! I've already experimented with it and I got it working. I'll clean up the code and get it up today

rustynwac:
I like making the experience better as long as we can keep it clean and not add technical debt we may need to debug later!

PR #1010, opened 2025-11-21
Add HPAC to Prod

kevinherdez:
This add HPAC to Prod #1004. HPAC has an agreement with CNFAIC to use their weather data widget. Unfortunately, this means that we can't use the HPAC center id to fetch the stations information as they have no token.

This means that the components `WeatherStationPage`, `WeatherStations`, and `WeatherStationDetails` all need to use 2 different center ids to properly fetch the information. The reason using both the center ids is to make sure the analytics accurately capture which AC is showing the page, even if the data isn't coming from them.

This change will also provide a slightly different experience on mobile than web for HPAC. On the web, the weather station widget shows the CNFAIC zones on the map. On mobile, the HPAC zone will be shown on the map with the CNFAIC data.

Tested on simulator dev build on iOS and Expo Go on Android

Simulator Screenshot - iPhone 17 Pro - 2025-11-21 at 12 22 51

Simulator Screenshot - iPhone 17 Pro - 2025-11-21 at 12 24 28

### DIFF
--- a/components/weather_data/WeatherStationDetail.tsx
+++ b/components/weather_data/WeatherStationDetail.tsx
@@ -10,7 +10,7 @@ import {incompleteQueryState, NotFound, QueryState} from 'components/content/Que
 import {Center, Divider, HStack, View, VStack} from 'components/core';
 import {BodyBlack, bodySize, BodyXSm, BodyXSmBlack} from 'components/text';
 import {formatDateTime} from 'components/weather_data/WeatherStationsDetail';
-import {usePartnerCenterIdForStationFetch} from 'components/weather_data/WeatherUtils';
+import {centerOrPartnerCenter} from 'components/weather_data/WeatherUtils';
 import {compareDesc} from 'date-fns';
 import {useAvalancheCenterMetadata} from 'hooks/useAvalancheCenterMetadata';
 import {useWeatherStationTimeseries} from 'hooks/useWeatherStationTimeseries';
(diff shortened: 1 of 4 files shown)


components/weather_data/WeatherUtils.test.ts
```diff
@@ -0,0 +1,18 @@
+import {usePartnerCenterIdForStationFetch} from 'components/weather_data/WeatherUtils';
+import {AvalancheCenterID, avalancheCenterIDSchema} from 'types/nationalAvalancheCenter';
+
+describe('usePartnerCenterIdForStationFetch', () => {
+  it('correctly updates the center id when needed', () => {
+    for (const center_id in avalancheCenterIDSchema) {
+      const centerIdValue = center_id as AvalancheCenterID;
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
kevinherdez: ESLint doesn't like hooks being called in a loop. I think this is cleaner than doing 1 test for each center but I can change this if it's bad practice

components/weather_data/WeatherStationDetail.tsx
```diff
@@ -32,8 +33,9 @@ interface columnData {
 }
 
 export const WeatherStationDetail: React.FC<Props> = ({center_id, stationId, source, requestedTime}) => {
+  const partnerCenterId = usePartnerCenterIdForStationFetch(center_id);
   const [days, setDays] = useState(1);
-  const avalancheCenterMetadataResult = useAvalancheCenterMetadata(center_id);
+  const avalancheCenterMetadataResult = useAvalancheCenterMetadata(partnerCenterId);
```
stevekuznetsov: This is only a partner center in the case of HPAC, so it's a bit misleading to name it as such - maybe keep the identifier and frame it as an override? 

```
const avalancheCenterMetadataResult = useAvalancheCenterMetadata(centerOrPartnerCenter(center_id));
```

We're just doing some logic (as opposed to using an API to drive the partner center designation) so we don't need to do a hook for it.

components/weather_data/WeatherUtils.test.ts
stevekuznetsov: We're not closing over some external system effect in the logic, so it can just be a helper function, not a hook.

stevekuznetsov (approved):

PR #1012, opened 2025-11-24
docs: add app deployment instructions

yuliadub:


yuliadub:
ok. going to just create a new branch, not sure what happened - i dont understand why the gahs are stuck

Issue #1014, opened 2025-11-24
Update Sierra (SAC) logo


rustynwac:
Sierra updated their logos. Logos [here](https://drive.google.com/drive/folders/1jOnreWSf8rwv38aRzYXgJXKTKVKa3B67?usp=drive_link)

kevinherdez:
Do you know if they have a preference for what they want to see on mobile?

Primary Logo Asset
|Selection|Header|
|---|---|
|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/33ec813c-ad2b-41a7-9297-72c525345662" />|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/a6a49c2d-749a-46bc-8955-06777eb62dc4" />|

Vertical Logo Asset
|Selection|Header|
|---|---|
|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/11230771-d331-4a2f-954d-a3839c214cbd" />|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/f4d913ae-9861-45c6-9d1c-fb7e87ab0fc0" />|

Icon Logo Asset
|Selection|Header|
|---|---|
|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/644744ad-189d-43ae-96d2-66bf7fa3409e" />|<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/ec4212cf-1a33-498d-a93a-219ca9f8e507" />|

PR #1016, opened 2025-11-24
Update SAC Logo

kevinherdez:
#1014. SAC logos needed to be updated with their new branding. The file was provided as a .webp instead of a .png, but I decided to leave it since React Native `Image` supports that file and it seems like it supports resizing nicely. It's only only 13 KB big which puts it on the smaller size compared to the other AC logo assets we have


rustynwac:
SAC_PrimaryIcon_FullColor_Dark
SAC_PrimaryLogo_FullColor-Dark

here ya go

Issue #1009, opened 2025-11-21
YouTube videos not rendering in forecast


rustynwac:
We upgraded the media viewer to include Youtube embeds but we only did that for obs and need to do update for forecasts.

<img width="678" height="701" alt="Image" src="https://github.com/user-attachments/assets/d3b746e9-282c-419d-a5f3-3f5bced57e6f" />

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/78558010-6756-43e6-889f-c3c7006a268e" />

Issue #1001, opened 2025-11-17
Remove old Media Viewer code


kevinherdez:
Now that we have a new media viewer that supports viewing YouTube videos. We need to remove the code for the old viewer. This will remove the `react-native-image-viewing` package that is unmaintained and currently being excluded from Expo Doctor checks.

PR #1019, opened 2025-11-26
Remove old carousel code

kevinherdez:
#1009 
#1001

This removes the old `Carousel` code that was image only and replaces it with the new `MediaCarousel` that supports both images and Youtube videos. By doing this, we can remove `react-native-image-viewing` from the repo which was no longer being supported

I added a section to the dev menu to view a forecast with a video and tested it on an iOS dev build and Expo Go on Android.

I was having a hard time finding a weather synopsis that had any media associated with it. With how the media viewer has been working in other areas, I'm confident the change is safe there too. 

For the Avalanche Problem Cards, we're filtering out videos when we receive the media from the backend. I didn't want to change that code as a part of this, but if that's something that we need to update, I'll make an issue to track it.

Simulator Screen Recording - iPhone 17 Pro - 2025-11-26 at 11 29 12


components/content/carousel/MediaCarousel.tsx
```diff
@@ -1,8 +1,21 @@
-import {MediaViewerModal} from 'components/content/carousel/carouselV2/MediaViewerModal/MediaViewerModal';
-import {ThumbnailList} from 'components/content/carousel/carouselV2/ThumbnailList';
+import {MediaViewerModal} from 'components/content/carousel/MediaViewerModal/MediaViewerModal';
+import {ThumbnailList} from 'components/content/carousel/ThumbnailList';
 import {View, ViewProps} from 'components/core';
 import React, {PropsWithChildren, useCallback, useState} from 'react';
-import {MediaItem} from 'types/nationalAvalancheCenter';
+import {ImageMediaItem, MediaItem, MediaType} from 'types/nationalAvalancheCenter';
+
+export const images = (media: MediaItem[] | null | undefined): ImageMediaItem[] => {
```
kevinherdez: This is directly copied from the deleted `index.tsx`

PR #1038, opened 2025-12-16
Avalanche Observations Part 2: Image Upload

kevinherdez:
This PR adds the ability to upload images with the avalanche details.

The idea for this directly piggybacks off how the image upload for normal observations work. Images are added to the form data, those images are used to create image upload tasks, and the result of the image upload is used to update the form metadata to include the images.

The key differences are:
- The avalanche id used to later update the image metadata is the index from the array
- The form data is being directly manipulated instead of the `extraData`

The decision was made to use the index to not have to add more unnecessary fields to the `avalancheObservationSchema`. Looking at the schema coming in from the backend, it doesn't look like there's the concept of an avalanche obs having an id. To properly set this up, we would need to create the id on the schema, generate an id, and then delete it before upload. The index approach uses what we already have, but admittedly is much less robust

I copied everything from `AvalancheImagePicker` from `ObservationImagePicker`. I'm going to go back in and make all of this code generic, like how we have for the other form fields, but for the sake of time I decided to make that change a follow up.

A follow up PR will have avalanche observation editing. Some of that code was changed to hide the edit button as not to confuse external testers. The same follow up PR will introduce the comments and Problem Type fields.

Simulator Screenshot - iPhone 17 Pro - 2025-12-15 at 12 07 31

### DIFF
--- a/components/observations/uploader/ObservationsUploader.ts
+++ b/components/observations/uploader/ObservationsUploader.ts
@@ -4,7 +4,7 @@ import {AxiosError} from 'axios';
 import _ from 'lodash';
 import uuid from 'react-native-uuid';
 
-import {ObservationFormData} from 'components/observations/ObservationFormData';
+import {ImagePickerAssetSchema, ObservationFormData} from 'components/observations/ObservationFormData';
 import {
   ImageTaskData,
   ObservationTask,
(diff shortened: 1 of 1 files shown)


components/observations/AvalancheObservationForm.tsx
```diff
@@ -41,6 +43,8 @@ export const AvalancheObservationForm: React.FC<{
     shouldUnregister: true,
   });
 
+  const insets = useSafeAreaInsets();
```
kevinherdez: It's recommended to not use `SafeAreaView` directly as this is more customizable and works better across platforms

stevekuznetsov: TIL, we have a lot of that in the codebase - maybe let's write up a card to track the fixes

components/observations/uploader/ObservationsUploader.ts
```diff
@@ -197,6 +197,43 @@ export class ObservationUploader {
...
+                exif: image.exif
+                  ? {
+                      Orientation: typeof image.exif.Orientation === 'string' || typeof image.exif.Orientation === 'number' ? image.exif.Orientation : undefined,
+                      DateTimeOriginal: typeof image.exif.DateTimeOriginal === 'string' ? image.exif.DateTimeOriginal : undefined,
+                    }
+                  : undefined,
+              },
+              caption,
+              name: name ?? '',
+              title: `Public Observation: ${avalanche.location}`,
```
stevekuznetsov: Is this title copy-pasta from the obs form? Is it still correct in this context?

components/observations/uploader/ObservationsUploader.ts
```diff
@@ -197,6 +197,43 @@ export class ObservationUploader {
         });
       });
 
+      observationFormData.avalanches.forEach((avalanche, index) => {
+        avalanche.images?.forEach(({image, caption}) => {
+          tasks.push({
```
stevekuznetsov: If the shape of the task being pushed is 99% identical to the other one, a refactor to a common helper with a callout for the diff would be nice

components/observations/AvalancheImagePicker.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const missingImagePermissions = imagePermissions !== null && !imagePermissions.granted && !imagePermissions.canAskAgain;
+
+  const {field} = useController<AvalancheObservationFormData, 'images'>({name: 'images', defaultValue: []});
+  const images = field.value;
+  const imageCount = images?.length ?? 0;
+
+  const isDisabled = imageCount === maxImageCount || disable || missingImagePermissions;
+
+  const {logger} = React.useContext<LoggerProps>(LoggerContext);
+  const pickImage = useCallback(() => {
```
stevekuznetsov: (general comment) if these are common/shared logic between this picker and the other picker, let's see if we can refactor to DRY it out - if there are substantial differences that make that difficult, then let's not

PR #1039, opened 2025-12-17
Use distance to sort the stations

kevinherdez:
Update to use the distance from the center of the AC region when sorting the stations within their clusters. The previous update sorted them by North to Sound, and West to East which still caused some farther away centers as being shown first because the clustering can be rather large. This should fix that issue at the tradeoff that it could feel a little jumpy using the cards to swipe between the stations


components/weather_data/WeatherStationMap.tsx
stevekuznetsov: If we're no longer using this, let's stop calculating it

PR #1050, opened 2026-01-05
Fix Alaska time issues

kevinherdez:
The fix that I made earlier actually wasn't adding any buffer to account for the time zone like I had thought. This fixes the issue by setting the time zone as Alaska instead of trying to add any buffer to the time. 

This will effectively set the time on the date to midnight Alaska time. Anyone who looks at an observation east of Alaska will still continue to see the correct date. The only difference is under the hood it might be anywhere from 01:00 to 04:00 depending on where they're looking at this from.

This should be a better fix, and maybe we can use the lessons learned here to try to get this changed on the backend.

### DIFF
--- a/utils/date.ts
+++ b/utils/date.ts
@@ -108,7 +108,9 @@ export const utcDateToLocalShortDateString = (date: Date | string | undefined | 
 };
 
 // This is used when dealing with observation start dates. These dates are given to us with no time zone or timestamp.
-// By giving it a timezone and setting it to miday, the date can be accurately represented in local time regardless of time zone shifts.
+// By setting the time zone to Anchorage, we're creating a buffer for the user so that localized date for the user anywhere east of Anchorage will show the same day.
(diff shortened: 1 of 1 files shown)


utils/date.ts
```diff
@@ -124,8 +123,8 @@ export const observationStartDateToLocalDateString = (date: Date | string | unde
   if (date == null) {
     return 'Unknown';
   }
-  const d = typeof date === 'string' ? toDate(date, {timeZone: 'America/Los_Angeles'}) : date;
-  d.setHours(12);
+
+  const d = typeof date === 'string' ? toDate(date, {timeZone: 'America/Anchorage'}) : date;
```
kevinherdez: This will still show an incorrect date if anyone is looking at an observation west of Alaska (ex Hawaii). I'm happy to change this to Honolulu to at least account for everyone in the United States time zones, but I figured the likelihood of that is small so I left it as Alaska

kevinherdez: Spoke offline and added comments explaining this. Closing.

PR #1080, opened 2026-02-03
Refactor to use mapbox and remove react-native-maps

kevinherdez:
This PR brings in mapbox, refactors the views that heavily used the old `<MapView/>`, and removes `react-native-maps`
- #1071 
- #1075
- #1072

### Where to start
- `ZoneMap.tsx` and `AvalancheForecastZonePolygon.tsx`
    - These two files contain the bulk of the changes that relate to Mapbox. The rest of the files that were changed are mainly a result of these two

### Differences between `@rnmapbox` and `react-native-maps`
- Mapbox has the idea of a `Camera` component that controls what the user is looking at. The `Camera` is used to programmatically change what the user is seeing on the map. In `react-native-maps` this is done directly on the `MapView`
- Mapbox uses layers to create the polygons that are shown on the map. They can directly receive GeoJSON data so a lot of the preprocessing that we were doing is not longer needed
- The `z-index` of the layers is determined by the order in which they are rendered in. The last layer has the highest `z-index`
    - This is why there's a conditional render for the `SelectedAvalancheForecastZonePolygon` component so that it's always the last one rendered
 - Mapbox does not have the concept of `Region` that is used heavily with `react-native-maps`. `Region` is refactored to be `AvalancheCenterRegion`. This allows a lot of the old calculations to remain the same and includes `CameraBounds` so that the view can be correctly positioned
 - Mapbox also heavily uses the GeoJSON concept of `Position` which is just a `number[]` where the first index is the longitude and the second is the latitude
- A lot of the `onPresses` that were passed to the `MapView` to handle different events are no longer required

### What's next
- Switch to the New Architecture
    - #1073

### DIFF
--- a/components/helpers/geographicCoordinates.ts
+++ b/components/helpers/geographicCoordinates.ts
@@ -1,55 +1,82 @@
+import {CameraBounds} from '@rnmapbox/maps';
 import booleanPointInPolygon from '@turf/boolean-point-in-polygon';
-import {LatLng, Region} from 'react-native-maps';
-import {MapLayerFeature, Position} from 'types/nationalAvalancheCenter';
+import {MapViewZone} from 'components/content/ZoneMap';
+import {AvyPosition, Geometry, MapLayerFeature, Position} from 'types/nationalAvalancheCenter';
 
 export interface RegionBounds {
-  topLeft: LatLng;
-  bottomRight: LatLng;
+  topRight: AvyPosition;
+  bottomLeft: AvyPosition;
 }
 
-export const regionFromBounds = (bounds: RegionBounds): Region => {
+const defaultAvalancheCenterMapRegionBounds: RegionBounds = {
+  topRight: {longitude: 0, latitude: 0},
+  bottomLeft: {longitude: 0, latitude: 0},
+};
+
+export interface AvalancheCenterRegion {
+  centerCoordinate: {longitude: number; latitude: number};
+  longitudeDelta: number;
+  latitudeDelta: number;
+  cameraBounds: CameraBounds;
+}
+
+export const toGeoJSONPosition = (avyPosition: AvyPosition): Position => [avyPosition.longitude, avyPosition.latitude];
+export const toAvyPosition = (geoJsonPosition: Position): AvyPosition => {
   return {
-    latitude: (bounds.topLeft.latitude + bounds.bottomRight.latitude) / 2.0,
-    latitudeDelta: Math.abs(bounds.topLeft.latitude - bounds.bottomRight.latitude),
-    longitude: (bounds.topLeft.longitude + bounds.bottomRight.longitude) / 2.0,
-    longitudeDelta: Math.abs(bounds.bottomRight.longitude - bounds.topLeft.longitude),
+    longitude: geoJsonPosition[0],
+    latitude: geoJsonPosition[1],
   };
 };
 
-export const updateBoundsToContain = (previous: RegionBounds, coordinates: LatLng[]): RegionBounds => {
+export const avalancheCenterRegionFromRegionBounds = (bounds: RegionBounds): AvalancheCenterRegion => {
+  const centerlong = (bounds.topRight.longitude + bounds.bottomLeft.longitude) / 2.0;
+  const centerLat = (bounds.topRight.latitude + bounds.bottomLeft.latitude) / 2.0;
+  const longDelta = Math.abs(bounds.bottomLeft.longitude - bounds.topRight.longitude);
+  const latDelta = Math.abs(bounds.topRight.latitude - bounds.bottomLeft.latitude);
+  const cameraBounds = {
+    ne: [centerlong + longDelta / 2.0, centerLat + latDelta / 2.0],
+    sw: [centerlong - longDelta / 2.0, centerLat - latDelta / 2.0],
+  };
+  return {
+    centerCoordinate: {
+      longitude: centerlong,
+      latitude: centerLat,
+    },
+    longitudeDelta: longDelta,
+    latitudeDelta: latDelta,
+    cameraBounds: cameraBounds,
+  };
+};
+
+export const updateBoundsToContain = (previous: RegionBounds, coordinates: Position[]): RegionBounds => {
   if (!coordinates) {
     return previous;
   }
   const bounds: RegionBounds = {
-    topLeft: {latitude: previous.topLeft.latitude, longitude: previous.topLeft.longitude},
-    bottomRight: {latitude: previous.bottomRight.latitude, longitude: previous.bottomRight.longitude},
+    topRight: {longitude: previous.topRight.longitude, latitude: previous.topRight.latitude},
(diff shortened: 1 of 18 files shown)


components/observations/ObservationDetailView.tsx
```diff
@@ -142,51 +143,21 @@ export const WeatherCard = ({observation, ...props}: {observation: Observation}
...
+  [-122.33426358631552, 47.6009139228834],
+  [-121.398011481564, 47.43050191839139],
+  [-121.09839553823569, 47.75708705098087],
+  [-121.41689869612796, 48.20857063251787],
+  [-121.60378434771066, 48.83086824211633],
+  [-121.45393919813452, 46.93280523754054],
 ];
 const isPlaceholder = (latitude: number, longtiude: number): boolean => {
-  return placeholders.map(point => point.latitude === latitude && point.longitude === longtiude).reduce((current, accumulator) => current || accumulator, false);
+  return placeholders.map(position => position[1] === latitude && position[0] === longtiude).reduce((current, accumulator) => current || accumulator, false);
```
yuliadub: instead of calling these position[1] and position[0] can we put them in corresponding latitude and longitude variables? easier to follow when reading the code that way

yuliadub: as I wrote that I realized that made sense in my brain but we are literally comparing it to "latitude" and 'longitude' but it did help me notice a typo in longtiude  :)

components/weather_data/WeatherStationMap.tsx
```diff
@@ -264,46 +237,32 @@ export const WeatherStationMap: React.FunctionComponent<{
...
 
+  const selectedKey = selected ? 'selected' : '';
+
   return (
-    <MapMarker
-      ref={markerRef}
-      key={station.properties.stid}
-      coordinate={coordinate}
+    <ShapeSource
+      key={`${station.properties.stid}-source-${selectedKey}`}
```
yuliadub: does matter that there will be a trailing '-' if selected is empty?

components/helpers/geographicCoordinates.ts
```diff
@@ -1,55 +1,73 @@
...
+
+export interface AvalancheCenterRegion {
+  centerCoordinate: {longitude: number; latitude: number};
+  longitudeDelta: number;
+  latitudeDelta: number;
+  cameraBounds: CameraBounds;
+}
+
+export const avalancheCenterRegionFromRegionBounds = (bounds: RegionBounds): AvalancheCenterRegion => {
+  const centerlong = (bounds.topRight[0] + bounds.bottomLeft[0]) / 2;
```
yuliadub: was there a particular reason we used to use floats (i.e. 2.0) and now we dont?

components/helpers/geographicCoordinates.ts
```diff
@@ -1,55 +1,73 @@
...
 
-export const updateBoundsToContain = (previous: RegionBounds, coordinates: LatLng[]): RegionBounds => {
+export const updateBoundsToContain = (previous: RegionBounds, coordinates: Position[]): RegionBounds => {
   if (!coordinates) {
     return previous;
   }
   const bounds: RegionBounds = {
-    topLeft: {latitude: previous.topLeft.latitude, longitude: previous.topLeft.longitude},
-    bottomRight: {latitude: previous.bottomRight.latitude, longitude: previous.bottomRight.longitude},
+    topRight: [previous.topRight[0], previous.topRight[1]],
```
yuliadub: probably evening brain, but why the switch here -- i.e. topLeft to topRight etc

components/map/AvalancheForecastZonePolygon.tsx
```diff
@@ -91,22 +52,24 @@ export const AvalancheForecastZonePolygon: React.FunctionComponent<AvalancheFore
...
-    const fillOpacity = renderFillColor ? zone.fillOpacity : 0;
-    const fillColor = colorFor(zone.danger_level).alpha(fillOpacity).string();
-    return (
-      <>
-        {polygonProps.map((prop, idx) => (
-          <Polygon key={idx} fillColor={fillColor} {...prop} />
-        ))}
-      </>
-    );
+    fillColor = colorFor(zone.danger_level).alpha(fillOpacity).string();
```
yuliadub: have you tested the flashing avalanche warning case? (just came to mind as I was looking at this line since that behavior differs the most from any other ones on the map that just show the color)

yuliadub:
> * Mapbox uses layers to create the polygons that are shown on the map. They can directly receive GeoJSON data so a lot of the preprocessing that we were doing is not longer needed

> * A lot of the `onPresses` that were passed to the `MapView` to handle different events are no longer required

nice! ❤️ 

looked through everything, left a few questions - will take another look tomorrow morning/early afternoon on more fresh brain. Thanks for working on this, nothing jumped out at me as not making sense thus far! Excited!

components/helpers/geographicCoordinates.ts
kevinherdez: I wanted to keep it consistent with the new `CameraBounds` concept that's coming from Mapbox. `CameraBounds` has 2 properties with are a Northeast position and a Southwest position that are basically top right and bottom left. This also makes the math a little easier because the top right value will have the biggest longitude and biggest latitude values and bottom left is the smallest values.

components/observations/ObservationDetailView.tsx
```diff
@@ -224,6 +196,10 @@ export const ObservationCard: React.FunctionComponent<{
   }, [postHog, observation.center_id, observation.id]);
   useFocusEffect(recordAnalytics);
 
+  const nePosition: Position = [(observation.location_point.lng ?? 0) + 0.075 / 2, (observation.location_point.lat ?? 0) + 0.075 / 2];
+  const swPosition: Position = [(observation.location_point.lng ?? 0) - 0.075 / 2, (observation.location_point.lat ?? 0) - 0.075 / 2];
```
stevekuznetsov: Did we previously accept null/undefined, or why `?? 0` now but not before? Naively seems like not showing a map is better than showing whatever lat/long 0/0 is, but the obs should always have the data - maybe schema for NWAC obs compatibility (which we no longer use) is why this is possibly null? Maybe good follow-up.

kevinherdez: I have no problem moving this down into the `ZoneMap` directly so that we don't have to null check anymore

components/weather_data/WeatherStationMap.tsx
```diff
@@ -178,74 +184,39 @@ export const WeatherStationMap: React.FunctionComponent<{
...
-      if (topElements.current) {
-        topElements.current.measureInWindow((x, y, width, height) => {
-          controller.animateUsingUpdatedTopElementsHeight(y, height);
-        });
-      }
-    }, 50);
-  }, [controller]);
-
-  const isRunningInExpoGo = Constants.appOwnership === AppOwnership.Expo;
+  const emptyHandler = useCallback(() => {}, []);
```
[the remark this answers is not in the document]
kevinherdez: Good call. For some reason I had it as a required parameter. I'll push a change to make this optional

Issue #82, opened 2023-01-18
Pick a single icon set if possible, and preload it


floatplane:
1) Right now we use a mixture of icons from **five** icon sets. This isn't good from a memory usage or a visual consistency standpoint

```
❯ rg "import.*vector-icons"
components/AvalancheForecastZoneMap.tsx
20:import {FontAwesome5} from '@expo/vector-icons';
 
components/TelemetryStationMap.tsx
4:import {FontAwesome5} from '@expo/vector-icons';
 
components/icons/nac-icons.tsx
1:import {createIconSet} from '@expo/vector-icons';
 
components/content/Carousel.tsx
5:import {AntDesign, FontAwesome5} from '@expo/vector-icons';
 
components/content/Card.tsx
6:import {FontAwesome} from '@expo/vector-icons';
 
components/DangerScale.tsx
3:import {MaterialIcons} from '@expo/vector-icons';
 
components/Observations.tsx
8:import {FontAwesome, MaterialCommunityIcons} from '@expo/vector-icons';
 
components/Observation.tsx
5:import {FontAwesome5, MaterialCommunityIcons, Fontisto} from '@expo/vector-icons';
 
App.tsx
7:import {AntDesign} from '@expo/vector-icons';

```

2) We should preload icons so they're ready to go when we need them - https://docs.expo.dev/archive/classic-updates/preloading-and-caching-assets/#pre-loading-and-caching-assets

stevekuznetsov:
@floatplane looks like we pre-load fonts these days but not icons, this seems relevant still.

stevekuznetsov:
@beaucollins I'm not sure if our startup/loading time is at all an issue, anecdotally it does not seem to be, but this may be something we can improve on for bundle size etc.

kevinherdez:
#1103 was pushed to Release. Closing issue

PR #1103, opened 2026-02-20
Use 1 Icon Set

kevinherdez:
This unifies the icons we're using in the app as we were pulling in icons from 8 different icon sets. I decided to go with Ionicons because it seemed to have all of the icons that we needed and it looked the best (in my opinion).

- #82 

I tested on both Android and iOS, and adjusted icon sizes to try to match the original sizes as best as possible.

I used this website to find the Ionicon equivalent from the other icon sets.
https://icons.expo.fyi/Index


yuliadub:
noticed there are two patterns

```javascript
import {Ionicons} from '@expo/vector-icons';
``` 

```javascript
import Ionicons from '@expo/vector-icons/Ionicons';
``` 

any specific reason for that? I saw lets standardize to one way to import if we can!

Issue #851, opened 2024-12-16
Logo is super small upon opening on Android


rustynwac:
Noted by Charlotte. Rusty not getting this behavior on iOS
![1000006061](https://github.com/user-attachments/assets/3ecec58c-288d-441f-a0f2-965024bc924e)

yuliadub:
As far as I know we are using the same assets on both iOS and android for the loading screen so this is likely a something something Android gotcha

PR #1110, opened 2026-02-25
Fix Android Splash Screen

kevinherdez:
- #851 

The splash screen image we were using was the size of a screen instead of the recommended 1024x1024. I cropped the splash screen image to center the logo in a 1024x1024 box and that seems to have done the trick. What seems to have been happening was because the `resizeMode` was `contain` in the splash screen settings on Android it was taking the image and trying to make it fit into some max size that they must have. iOS just seems to use the image as is.

The splash screen cannot be tested in a dev build connected to metro (or Expo Go). I tested this using a dev build on EAS for Android but this has not been tested on iOS since we need to use Preview to test it. Once this is checked in and pushed to Preview, I'll make sure that this doesn't change the splash screen looks there.

Guideline docs
https://docs.expo.dev/develop/user-interface/splash-screen-and-app-icon/

Before and after video
https://github.com/user-attachments/assets/b4cf3a4d-4116-45c5-b9ab-7ed7ae3e6da9


yuliadub (approved):
I bet apple should be fine but thanks for calling this out. lets merge this and test on preview to make sure :)

kevinherdez:
This is in Preview

Issue #1118, opened 2026-03-06
Move the Menu into a side drawer


kevinherdez:
- Move the menu tab into a side drawer
- Update the UI to match the designs

Issue #1101, opened 2026-02-20
Update the navigation header


kevinherdez:
Update the navigation header to match the designs

https://www.figma.com/design/Bai2w7x0vUakL5WRjPGRg7/AVY-Native-App?node-id=3150-5579&p=f&t=qJXxchehJMUHrDPB-0

PR #1137, opened 2026-03-17
Drawer Navigation, Navigation Stack Refactor, and Nav Header Branding

kevinherdez:
This PR address the following issues:
- #1118 
- #1102 
- #1101
- #1136 

## Overview

This pulls in `@react-navigation/drawer` to move the `Menu` in the drawer, and refactors the navigation stack to correctly hide the bottom tab in the views where it needs to be hidden according to the designs.

## Where to Start

`App.tsx`
- Just to see the starting point for the stack changes

`Drawer.tsx`
- The new top level navigator. From here you'll be able to follow all of the new nested stack navigators

### Refactor  

The screen definitions that use to be in `ObservationsScreen.tsx` and `WeatherScreen.tsx` have been moved into `MainStack.tsx`

## Navigation Stack Overview and Refactor

To achieve the behavior we want there are now 3 levels to the Navigation Stack

`DrawerNavigator`
|
`MainStackNavigator`
|
`BottomTabNavigator`

### DrawerNavigator

This only contains the `MainStackNavigator` as its main screen and it contains the UI for the drawer itself.

### MainStackNavigator

This contains the `BottomTabs` and any screen that is presented as a part of the "main" app experience. This moves a lot of views like the `ForecastScreen`,  `ObservationDetailScreen`, etc out of the bottom tab navigator so that the bottom tabs are not shown in these views. This also contains views that are navigated to from the drawer itself like the `AboutScreen` and the `AvalancheCenterSelector`. This is the based on React Navigation docs as best way to hide the bottom tab in views: https://reactnavigation.org/docs/hiding-tabbar-in-screens/

### BottomTabs

This is the old TabNavigator that was in App.tsx. However, the `HomeTabScreen`, `ObservationsTabScreen`, and the `WeatherScreen` have all been refactored to remove their own navigation stack. `BottomTabs` now only has 3 screens, none of which are their own stack navigators.

### Navigation Headers

`BottomTabNavigationHeader` was created for use in the 3 screens contained in `BottomTabs`. This is because the `header` definition in `BottomTab.Navigator` passes `BottomTabHeaderProps` instead of `NativeStackHeaderProps`. For simplicity, I broke out the different headers into 2 different components instead of trying to combine them into 1. 

### Branding
`BottomTabNavigationHeader` displays the branding for the center. Currently, that is just the logo with the acronym and no special colors

### ForecastScreen
This PR removes the ability to change zones when viewing the full avalanche forecast as per designs

### SafeArea

Many of the views that were a part of `BottomTabs` needed to be updated to have the correct bottom padding now that they no longer have a bottom tab

## Recording with the changes in action

https://github.com/user-attachments/assets/84d0f748-3c9d-43db-9720-8cfa46acce6c


components/screens/navigation/MainStack.tsx
```diff
@@ -0,0 +1,135 @@
...
+        component={ObservationSubmitScreen}
+        options={{
+          title: 'Submit an Observation',
+        }}
+      />
+      <MainStack.Screen
+        name="observationsPortal"
+        component={ObservationsPortalScreen}
+        options={{
+          title: 'Observation Portak',
```
yuliadub: portak has a cool sound to it 🙂

components/content/NavigationHeader.tsx
```diff
@@ -76,16 +75,14 @@ export const NavigationHeader: React.FunctionComponent<
   }, [shareUrl]);
 
   return (
-    // On phones with notches, the insets.top value will be non-zero and we don't need additional padding on top.
-    // On phones without notches, the insets.top value will be 0 and we don't want the header to be flush with the top of the screen.
-    <View style={{width: '100%', backgroundColor: 'white', paddingTop: Math.max(8, insets.top)}}>
-      <HStack justifyContent="space-between" pb={8} style={options.headerStyle} space={8} pl={3} pr={16}>
+    <View style={{width: '100%', backgroundColor: colorLookup('white'), paddingTop: insets.top, justifyContent: 'center', alignContent: 'center'}}>
```
yuliadub: I know we talked about this on slack - but looking at the deleted comment I think context here on why we decided to change paddingTop would be useful!

components/screens/navigation/BottomTabs.tsx
```diff
@@ -0,0 +1,149 @@
...
+          )
+        }
+      </BottomTabNavigator.Screen>
+    </BottomTabNavigator.Navigator>
+  );
+};
+
+// MARK: Bottom Tab Screens
+
+const MapScreen = ({route}: NativeStackScreenProps<TabNavigatorParamList, 'Map'>) => {
```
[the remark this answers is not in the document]
yuliadub: ok monday brain, this view does actually return the forecast zone up. Ignore me 🤦

components/screens/navigation/RootStack.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <RootStack.Screen name="buttonStylePreview" component={ButtonStylePreview} options={{title: `Button Style Preview`, headerBackButtonDisplayMode: 'minimal'}} />
+      <RootStack.Screen name="textStylePreview" component={TextStylePreview} options={{title: `Text Style Preview`, headerBackButtonDisplayMode: 'minimal'}} />
+      <RootStack.Screen
+        name="avalancheComponentPreview"
+        component={AvalancheComponentPreview}
+        options={{title: `Avalanche Component Preview`, headerBackButtonDisplayMode: 'minimal'}}
+      />
+      <RootStack.Screen name="toastPreview" component={ToastPreview} options={{title: `Toast Preview`, headerBackButtonDisplayMode: 'minimal'}} />
+      <RootStack.Screen name="timeMachine" component={TimeMachine} options={{title: `Time Machine`, headerBackButtonDisplayMode: 'minimal'}} />
+      {/* <RootStack.Screen
```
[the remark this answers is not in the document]
yuliadub: actually this part of the bottomTabNavigator so maybe not. but I think stuff is set up now we do need it for the developer menu only for viewing test cases like active warnings/etc so that may need some refactoring

routes.ts
```diff
@@ -49,72 +36,23 @@ export type HomeStackParamList = {
...
-    requestedTime: RequestedTimeString;
-  };
-  observation: {
-    id: string;
-  };
-  nwacObservation: {
-    id: string;
-  };
+export type DrawerParamList = {
+  MainStack: MainStackNavigationProps;
```
yuliadub: similarly here for type-safe nexted navigation:

`MainStack: NavigatorScreenParams<MainStackParamList> | undefined;`

components/map/AnimatedCards.tsx
```diff
@@ -216,6 +216,10 @@ export class AnimatedMapWithDrawerController {
     }
   }
 
+  forceAnimateMapRegion() {
```
yuliadub: why do we need this force?

routes.ts
```diff
@@ -1,34 +1,21 @@
...
 
-type WeatherStationsDetailPageProps = {
-  center_id: AvalancheCenterID;
-  zoneName: string;
-  name: string;
-  stations: Record<string, WeatherStationSource>;
-  requestedTime: RequestedTimeString;
-};
+export type MainStackParamList = {
+  bottomTabs: TabNavigationProps;
```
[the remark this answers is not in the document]
kevinherdez: Oh nice I didn't know that. I'll make the changes for all of these.

yuliadub (approved):
looks awesome! thanks for all your work on this!

2 things but im signing off: 

1. This is claude suggestion given the new navigation stack: 
 
The ObservationForm.tsx comment says "SafeAreaView shouldn't inset from bottom edge because TabNavigator is sitting there" — but in the new structure, [observationSubmit] lives in MainStack, not inside BottomTabs. There's no bottom tab below the observation form anymore, so the bottom content will render behind the device's home indicator. The bottom edge should likely be added:

The same logic applies to any screen that moved from inside the old per-tab stack navigators to MainStack (observation detail views, weather station details, forecast screen). Each of these has SafeAreaView configs that were tuned for the old hierarchy. Worth a quick audit of all moved screens.

2. I meant to comment about this last time but forgot, but do we still need nwacObservation? i know for a while NWAC obs were not on the NAC but I thought they have all moved there. Maybe worth just flagging as an issue, in no way a blocker for this PR

Issue #1162, opened 2026-04-01
Create ObservationDetail Modal


kevinherdez:
Present the ObservationDetail view as a modal when we receive a deep link.

kevinherdez:
Released in Preview on iOS build 64 and Android build 78

Issue #1161, opened 2026-04-01
Fix Stack Config


kevinherdez:
The `linking` config needs to be updated to respect the new stack

kevinherdez:
Released in Preview on iOS build 64 and Android build 78

Issue #1160, opened 2026-03-31
Audit UX of Observation Detail Card


rustynwac:
When viewing an observation, is it clear where the observation is located? The AFP has delivered center-less observation functionality. In the app you can be in one center and open a shared link from another center. Is it clear where you are looking? 

Video [here](https://drive.google.com/file/d/1OSZiCwoq9rnNJ7tEzl8MRb0jj6wqLy7M/view?usp=sharing)

PR #1165, opened 2026-04-02
Fix deep linking config based on stack changes

kevinherdez:
- #1161
- #1162

This fixes the `linking config` for the new stack and introduces `ObservationDetailViewModal`. 

The reason for the addition is that we want to show the `ObservationDetailView` from a modal instead of pushing it onto the stack. I decided to break this out into a separate view because the header that we show depending on if we're showing the modal or showing the stack is different, and I didn't want to overload `NavigationStackHeader` for this one use case.

Simulator Screen Recording - iPhone 17 Pro - 2026-04-02 at 16 25 09

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -424,11 +424,15 @@ const BaseApp: React.FunctionComponent<{
     filter: (url: string) => url.includes('/observations/'), // Only handle observation links
     config: {
       screens: {
-        Observations: {
+        MainStack: {
           path: 'observations/#/view/observations',
           screens: {
-            observationsList: '',
-            observation: ':id',
+            bottomTabs: {
+              screens: {
+                Observations: '',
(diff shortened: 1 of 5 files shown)


yuliadub:
this flow makes sense and looks very clean! my only thought here is that this modal doesnt have a "share" in there. not a big deal since someone shared it with me and I can go back to the text I got it from and copy and paste the link to share with someone else but it woudn't be an uncommon flow for someone to send me a link to an observation and then me wanting to send it someone else.

kevinherdez:
Makes sense. We have this issue #1160 tracking a bigger discussion around how to make the details page better in terms of letting a user know where the observation is coming from. That could help this flow because it might feel a little crowded to have both the logo and the share button in the header

Issue #1169, opened 2026-04-13
Center map correctly in No Center State


kevinherdez:
When a user closes the app and reopens it while in the 'No Center State', the map will center on their preferred Avalanche Center that they have selected under the hood.

We need to figure out what the experience should be around this:
- Should the map recenter on their last known location and zoomed out?
- Should we save their last map camera state to try to reload the map as it was when they closed the app?
- Should we fetch the user's current location without them having to hit the location button?

rustynwac:
My expectation would be:

> Should we save their last map camera state to try to reload the map as it was when they closed the app?

But maybe it's worth discussing

kevinherdez:
The map centering is released in Preview on iOS build 64 and Android build 78. I'm leaving this issue open for now in case we want to tweak the design

kevinherdez:
I'm going to go ahead and close this since it seems like the behavior we want. This is in Preview v80

Issue #1168, opened 2026-04-13
Match No Center State to designs


kevinherdez:
The 'No Center State' uses and app logo and a togo SVG that don't exist in the app. These need to be added to match the designs

kevinherdez:
This is in Preview v80

Issue #1166, opened 2026-04-08
Fix Drawer opening on screens where it shouldn't


kevinherdez:
You can open the drawer by swiping from the left from views that get pushed onto the stack. 

The expected behavior here should be that the pushed view comes off the stack when swiped, and only the views that have the menu button should open the drawer when swiped from the left

kevinherdez:
Released in Preview on iOS build 64 and Android build 78

Issue #1122, opened 2026-03-06
User location and no center state


kevinherdez:
- Allow the user to go into a no center state

kevinherdez:
Released in Preview on iOS build 64 and Android build 78

PR #1170, opened 2026-04-13
No Center State

kevinherdez:
- #1122 
- #1089 
- #1166
- #1169  

Introduces the 'no center state' to the experience as well as fixes a couple polish bugs

## Overview

When a user zooms out to a set threshold on the map, they enter the 'no center experience' meaning that all of the branding is removed and the bottom tab bar goes away to allow the user to focus on the map. They can come back into the 'center focused experience' by tapping on forecasting zone on the map or by selecting a center in the menu

### Changes to `Preferences`

Introduces the `isInNoCenterExperience` and the `lastMapCamera` preferences to control whether the user should see the 'no center' experience, and the last saved camera position respectfully. Putting this in preferences allows us to persist this across app launches, and by making it a separate preference only the views that need to handle this state need to be changed. By saving `lastMapCamera` we're able to reload the map where the user last left off when they're looking at the no center experience.

Based on discussions in the design doc, I decided not to have this experience based on setting the user's preferred center to `undefined` (i.e no center). That would have eliminated the need for another preference, but it would've required us to handle that state in most of our hooks/views

### Changes to `AvalancheForecastZoneMap`

I broke this file out into 3 files to make it easier to reason about the code.
1. `AvalancheForecastZoneMap`
    - I left the original name for this file because we're used to it. It now handles showing the avalanche selector modal, fetching the user's location, and  fetching/passing necessary information to the map
2. `AvalancheForecastMapView`
    - This file takes the information given from `AvalancheForecastZoneMap` and create the map the user sees along with handling all the interactions on the map
3. `AvalancheForecastZoneCards`
    - Nothing really changed here. I just broke this one out into its own file

### Custom bottom tab bar and new `ForecastMapNavigationHeader`

To show/hide the bottom tab bar and remove the branding from the navigation header, new custom components where created to handle this. These are `AnimatedBottomTabBar` and `ForecastMapNavigaitonHeader`

The biggest change is that the tab bar is now using `AnimatedBottomTabBar` instead of the built in `react-navigation` one. This allows us to control the animation of the bar which isn't supported in the built in one. The downside to this is that the tab bar is absolutely positioned so that the map extends beneath it. This means that we have to handle this in all the views that show the tab bar as `react-navigation` will no longer do this for us. Thankfully the new map project experience only has 3 views showing the tab bar

The custom `ForecastMapNavigationHeader` is only shown on the `AvalancheForecastZoneMap` and is used to nicely transition from showing the branded header, and showing the 'no center' header

### New packages

`expo-location` and `expo-linear-gradient` were added to fetch the user's location and have the new custom header match the designs repectfully

`expo-location` is used instead of `rnmapbox`'s built in location fetching because it allows us to fetch the user's location 1 time whereas in `rnmapbox` they have it set up to continuously fetch the user's location without a great way to just fetch it 1 time

## Screen recording of it in action

https://github.com/user-attachments/assets/076d36fb-e021-4969-be30-dfc4ddd88e4d

## Known issues
- The Drawer header for the 'No Center State' and the logo in the custom navigation header don't exactly match the designs.
    - #1168


components/map/AnimatedCards.tsx
```diff
@@ -357,16 +357,18 @@ export interface AnimatedCardsProps<T, U> {
...
 export const AnimatedCards = <T, U>(props: AnimatedCardsProps<T, U>) => {
-  const {center_id, date, items, getItemId, selectedItemId, setSelectedItemId, controllerRef, renderItem, buttonOnPress} = props;
+  const {center_id, date, items, getItemId, selectedItemId, setSelectedItemId, controllerRef, renderItem, buttonOnPress, bottomOffset = 0} = props;
   const {logger} = React.useContext<LoggerProps>(LoggerContext);
 
   const {width} = useWindowDimensions();
 
   const [previouslySelectedItemId, setPreviouslySelectedItemId] = useState<U | null>(null);
   const [programmaticallyScrolling, setProgrammaticallyScrolling] = useState<boolean>(false);
+  const isUserScrolling = useRef(false);
```
kevinherdez, line 371: This was added because navigations triggered `onScroll` events that would fire `handleScroll`. 

This was an issue when opening up the `Drawer`. Opening the `Drawer` would trigger the `handleScroll` which would then change the `selectedItemId` and show the cards even though they're supposed to be hidden

components/map/ZoneMap.tsx
kevinherdez, line 1: I tried to move the files in the first commit, but it looks like I did this one after. The big change here is that `ZoneMap` no longer is a `forwardRef` and instead directly accepts the `cameraRef` as a prop

components/map/AvalancheForecastZoneCards.tsx
kevinherdez, line 1: This file was moved in the second commit. The only change here is on line 64 where `{(zone.start_date || zone.start_date) ....` was changed to `{(zone.start_date || zone.end_date) ...`

components/map/AvalancheForecastZoneMap.tsx
```diff
@@ -0,0 +1,277 @@
...
+            results={[allMapLayersResult, metadataResult, ...forecastResults, ...warningResults]}
+            terminal
+            customMessage={{
+              notFound: () => ({
+                headline: 'Missing forecast',
+                body: 'There may not be a forecast available for today.',
+              }),
+            }}
+          />
+        </Center>
```
kevinherdez, line 252: This needed to be moved inside of the return instead of it being separate because if we were in a loading state, then the custom header wouldn't show

Preferences.tsx
```diff
@@ -71,9 +72,9 @@ export const PreferencesProvider: React.FC<PreferencesProviderProps> = ({childre
 
   const setPartialPreferences = useCallback(
     (newPreferences: Partial<Preferences>) => {
-      setPreferences(merge({}, preferences, newPreferences));
+      setPreferences(current => merge({}, current, newPreferences));
```
kevinherdez, line 87: This is a fix where if multiple `setPreferences` are fired in other areas, then the correct state is captured

components/map/AvalancheForecastMapView.tsx
```diff
@@ -0,0 +1,183 @@
...
+  }, [bottomTabBarHeight, controller]);
+
+  React.useEffect(() => {
+    controller.current.animateUsingUpdatedTopElementsHeight(topElementMeasurements.yPos, topElementMeasurements.height);
+  }, [controller, topElementMeasurements]);
+
+  const onCameraChanged = useCallback(
+    (mapState: MapState) => {
+      if (mapState.gestures.isGestureActive) {
+        if (mapState.properties.zoom < 6 && controller.current.state !== AnimatedDrawerState.Hidden) {
```
yuliadub: what is the reason the number here is 6, but then below on line 145 it is 5?

yuliadub: I assume there is a specific reason so lets put them in a const and explain why they are what they are

components/map/AvalancheForecastMapView.tsx
```diff
@@ -0,0 +1,183 @@
...
+          controller.current.setState(AnimatedDrawerState.Hidden, false);
+          setSelectedZoneId(null);
+        }
+
+        // There's a difference threshold for the no center experience so that the transition between the center focused state and the no center state is smoother
+        if (!isInNoCenterExperience && mapState.properties.zoom < 5) {
```
yuliadub: with how bad people can be at zooming and unzooming depending on their patience, are we worried about transitions to and from no center to center and flooding the stack with lots of events if someone doesnt quite get how close or far they need to be to go into a center experience (likely a non issue and just something we need to test but wanted to put it as a comment as a general thought)

App.tsx
```diff
@@ -454,6 +454,12 @@ const BaseApp: React.FunctionComponent<{
         }
       }
 
+      // When receiving a deep link, we need to make sure we're in the center focused experience so that the tab bar is shown when
+      // navigating to the observations tab
+      if (preferences.isInNoCenterExperience) {
+        setPreferences({isInNoCenterExperience: false});
```
yuliadub: I am not sure I follow this logic - this triggers a state update directly, could this cause issues with the overall preference (say if a user was in a different center before receiving the link) maybe instead it would make sense to do a useEffect instead that triggers when a deep link is detected?

components/weather_data/WeatherStationMap.tsx
```diff
@@ -197,8 +203,8 @@ export const WeatherStationMap: React.FunctionComponent<{
   return (
     <>
       <ZoneMap
-        style={StyleSheet.absoluteFillObject}
-        ref={mapCameraView}
+        style={{flex: 1, paddingBottom: bottomTabBarHeight}}
```
yuliadub: Claude tells me: 

Applying paddingBottom to a MapView may not work as expected on all platforms. Maps typically need padding applied via their internal camera/padding APIs rather than CSS-style padding. Verify this works correctly on both iOS and Android.

components/map/AvalancheForecastZoneMap.tsx
```diff
@@ -0,0 +1,277 @@
...
+    setTimeout(() => {
+      if (topElements.current) {
+        topElements.current.measureInWindow((x, y, width, height) => {
+          setTopElementMeasurements({yPos: y, height: height});
+        });
+      }
+    }, 50);
+  }, [setTopElementMeasurements]);
+
+  const onFetchLocation = useCallback(() => {
```
yuliadub, line 181: f both getLastKnownPositionAsync and getCurrentPositionAsync fail (e.g., airplane mode, GPS unavailable), there's no try/catch around the location fetch itself after the permissions check. Would we able to recover if this happens or do we need to add logic to handle that case?

components/content/navigation/AnimatedBottomTabBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const TabItem: React.FunctionComponent<TabItemProps> = ({routeKey, routeName, label, isFocused, tabBarIcon, onTabPress}) => {
+  const onPress = useCallback(() => {
+    onTabPress(routeKey, routeName, isFocused);
+  }, [onTabPress, routeKey, routeName, isFocused]);
+
+  const color = isFocused ? (colorLookup('primary') as string) : (colorLookup('text.secondary') as string);
+
+  return (
+    <TouchableOpacity style={styles.tab} onPress={onPress} accessibilityRole="button" accessibilityState={isFocused ? {selected: true} : {}}>
```
yuliadub: Another Clause suggestion (since we are using accessibilityRole="button") : 

Consider also including accessibilityLabel with the tab name for screen reader users, since the icon-only rendering doesn't describe the tab's purpose to VoiceOver/TalkBack.

components/content/navigation/ForecastMapNavigationHeader.tsx
```diff
@@ -0,0 +1,101 @@
...
+import {Title3Black} from 'components/text';
+import {LinearGradient} from 'expo-linear-gradient';
+import React, {useCallback, useEffect} from 'react';
+import {Image, StyleSheet} from 'react-native';
+import Animated, {useAnimatedStyle, useSharedValue, withTiming} from 'react-native-reanimated';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+import {colorLookup} from 'theme';
+import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
+
+interface ForecastNavigaitonHeaderProps {
```
yuliadub: Forecast**Navigation**HeaderProps  🙂

yuliadub:
Sorry it took me so long, excited to see all this coming together! Just two comments that are more general: 

Since the AnimatedBottomTabBar is absolutely positioned
- All scrollable content needs manual paddingBottom to avoid being obscured.
- This PR adds paddingBottom in ObservationsListView and NWACWeatherStationList, but any new views added in the future will need to remember this - can we de-risk that?
- maybe something like using a context (like BottomTabBarHeightContext) consistently, which is already partially done via useBottomTabBarHeight().

______

I wont lie I am not the fan of the way re-renders feel right now that we have when we go into the center experience, it feels almost like the app is about to crash but then it re-loads with the new center context. When I was explicitly clicking on the zone in the previous version, that experience felt natural but now with zooming it feels a bit disruptive. I am not sure if we could keep the map in the background as is and re-render the stuff on top but something like that may feel less disruptive in my opinion. Or could we slide the top nac with the logo in but not actually reload until the user clicks on an actual zone? (anyways, not a blocker by any means but just something that I noticed was throwing me off a bit)

components/map/AvalancheForecastMapView.tsx
kevinherdez: That's a good call out. Right now once you're out of the no center state the only way to get back in is to actively tap on a zone. However, we could play around with debouncing this function if we see that the performance is deteriorating

App.tsx
kevinherdez: That's probably the better approach here. I'll update this to put it in a `useEffect`, and I'll add extra logic to only do this, if the link we received is going to take them to the Observation List. For those links, we need to bring the tab bar back so that we don't put them in a state where they're stuck in the Observation List with no way to change tabs. Now that the observation detail modal is checked in, we don't need to show the tab bar for those links.

Under the hood the user always have a center that they're tied to so we wouldn't be changing the center on them, but I could see this being a confusing experience for them. There's a larger discussion that we need to have about what the experience should be when receiving links from different centers especially is those links take you to the Obs list

kevinherdez: That's being tracked here, #1163

components/weather_data/WeatherStationMap.tsx
kevinherdez: It seems to work correctly on both, but I can look into changing how the padding is adjusted with the camera

kevinherdez:
>- This PR adds paddingBottom in ObservationsListView and NWACWeatherStationList, but any new views added in the future will need to remember this - can we de-risk that?
>- maybe something like using a context (like BottomTabBarHeightContext) consistently, which is already partially done via useBottomTabBarHeight().

There's definitely an opportunity here to make this better. I can wrap the views that need it in a `BottomTabBarHeightContext` to make it more clear that those views need to handle it. 

However, this is partially mitigated by the new  app design. Now, only the main tab views show the tab bar and all other views have the tab bar hidden as a part of the navigation stack refactor. I'm hoping that it's more common that we add a view that will be shown outside of the context of the bottom tab instead of adding new views to the tab bar.

components/map/AvalancheForecastZoneMap.tsx
kevinherdez: We need to add logic to handle that.  Only `getCurrentPositionAsync` needs to be wrapped in a try/catch. I'm not too sure what we would want to do when we get in that state, but my initial thoughts would be to show an alert saying it failed with a try again button

components/weather_data/WeatherStationMap.tsx
kevinherdez: I actually needed to adjust this to `marginBottom`. Changing the camera padding has the map view going under the tab bar which isn't what we want here

kevinherdez:
In the new commit, I added `preferencesLoaded` to the `PreferencesContext` so that `isInNoCenterState` correctly sets to `false` when the app is launched through a deep link. By moving that code into a `useEffect` it was actually setting the preference while `AsyncStorage` was will reading the value so `AsyncStorage` was setting it back to `true` after we navigated to the Obs tab

components/map/AvalancheForecastMapView.tsx
kevinherdez: I wrapped `isInNoCenterExperience` in a ref to try to limit the amount of cascading changes that happen when it's used in a `useCallback`

components/map/AvalancheForecastZoneMap.tsx
kevinherdez: I just have it logging an error until we decide what to do here

PR #1175, opened 2026-04-15
Replace CorniceFall problem type with Cornice

kevinherdez:
A Zod Error came in that uncovered an issue with the `observation -> advanced_fields -> avalanche_problems -> type` schema where if the type didn't match the specific `AvalancheProblemName` enum, then we failed to parse.

The problem was that one of the avalanche problems had a type of 'Cornice' instead of 'Cornice Fall'.

Both the AFP widget to create an observation and the AFP dashboard to create a forecast have this value set to `Cornice` so the best fix seems to be to update our schema to match.

I updated `AvalancheProblemType` to match as well just for consistency

Obs that had the issue:
https://nwac.us/observations/#/view/observations/9801b523-356b-440c-9f30-1fe0607ce9fa

Zod Error:
```
ZodError: [
  {
    "code": "too_big",
    "maximum": 0,
    "type": "string",
    "inclusive": true,
    "exact": true,
    "message": "String must contain exactly 0 character(s)",
    "path": [
      "advanced_fields",
      "avalanche_problems",
      2,
```

### DIFF
--- a/components/AvalancheProblemIcon.tsx
+++ b/components/AvalancheProblemIcon.tsx
@@ -19,7 +19,7 @@ const icons: Record<AvalancheProblemType, ImageSourcePropType> = {
   [AvalancheProblemType.DeepPersistentSlab]: require('../assets/problem-icons/DeepPersistentSlab.png'),
   [AvalancheProblemType.WetLoose]: require('../assets/problem-icons/WetLoose.png'),
   [AvalancheProblemType.WetSlab]: require('../assets/problem-icons/WetSlab.png'),
-  [AvalancheProblemType.Cornice]: require('../assets/problem-icons/CorniceFall.png'),
+  [AvalancheProblemType.CorniceFall]: require('../assets/problem-icons/CorniceFall.png'),
   [AvalancheProblemType.Glide]: require('../assets/problem-icons/Glide.png'),
 };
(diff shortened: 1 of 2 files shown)


yuliadub:
if we are 100% sure CorniceFall will never make another debut, this change looks good (but if we are not, maybe we keep both?) referencing a similar issue we have run into in the past if you want to cross check if us keeping both was not the right solution (and maybe we can remove one of them too: https://github.com/NWACus/avy/pull/1175/changes)

kevinherdez:
I asked the NAC about what to do here, and Aaron's response was to keep both. I pushed a change adding Cornice Fall back to the name, and one to undo the changes that renamed the type `Cornice`.

"You should keep both for Cornice and Glide, at least for now.

Forecasts store an FK to national.avalanche_problems, the name column is literally 'Cornice Fall' / 'Glide Avalanches'. So even though the dropdown labels it "Cornice", the persisted value is an FK (id=8 → 'Cornice Fall'), and that's what the API emits. 1,273 forecast rows with Cornice or Glide.
Obs stores type as a raw string inside JSONB. The widget writes Cornice / Glide today, but the database has 6 Cornice Fall and 2 Glide Avalanches. Both live in the same column."

Issue #1125, opened 2026-03-07
Fix MediaPreview Error State


kevinherdez:
Rusty found a bug that when the data contains a bad media item the `<InternalError/>` component we were trying to load completely messed up the forecast tab

Issue #1092, opened 2026-02-14
PDFs don't work in MediaViewer


kevinherdez:
While investigating a Sentry issue in PostHog, I noticed that we were getting `MediaViewer-UnsupportedMediaType` events from PDF files. These need to be supported

kevinherdez:
PR out for this that I need to update against the latest Map Project/Expo updates

PR #1179, opened 2026-04-29
Media Viewer Updates

kevinherdez:
- #1092 
- #1125 

The first commit is the fix for the MediaPreview error state. It just needed to be wrapped in a view that was the same height and width of the thumbnail to render correctly
Simulator Screenshot - iPhone 17 Pro - 2026-04-28 at 14 53 37

## PDF Viewing
The second commit introduces viewing PDFs in the app. This experience works in the same way that the `WebVideoView` does which is presenting the PDF inside of a `WebView` since both `WebViews` on iOS and Android support PDF viewing.

### Important Changes
A lot of the code for `WebVideoView` was broken out into `WebMediaView` that now both `WebVideoView` and the new `PDFView` utilize. The `WebView` code was also cleaned up to better handle centering the view

A new `PDFThumbnail` view was added. This mimics the web experience in that we don't show the preview of the PDF itself. Instead we just show that we have a document in the list and show an icon.

|PDF Thumbnail List|PDF in the media viewer|
|---|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-04-28 at 16 10 57|Simulator Screenshot - iPhone 17 Pro - 2026-04-29 at 12 06 15|

### DIFF
--- a/components/content/carousel/MediaViewerModal/MediaContentView.tsx
+++ b/components/content/carousel/MediaViewerModal/MediaContentView.tsx
@@ -1,16 +1,13 @@
 import {ImageView} from 'components/content/carousel/MediaViewerModal/ImageView';
+import {PDFView} from 'components/content/carousel/MediaViewerModal/PDFView';
 import {WebVideoView} from 'components/content/carousel/MediaViewerModal/WebVideoView';
 import {View} from 'components/core';
 import {BodySm} from 'components/text';
 import {usePostHog} from 'posthog-react-native';
 import React, {useEffect} from 'react';
-import {Dimensions} from 'react-native';
+import {useWindowDimensions} from 'react-native';
 import {NativeGesture} from 'react-native-gesture-handler';
 import {MediaItem, MediaType} from 'types/nationalAvalancheCenter';
-
-const SCREEN = Dimensions.get('screen');
-const SCREEN_WIDTH = SCREEN.width;
-const SCREEN_HEIGHT = SCREEN.height;
 
 interface MediaContentProps {
   item: MediaItem;
@@ -20,14 +17,17 @@ interface MediaContentProps {
 
 export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({item, isVisible, nativeGesture}) => {
   const postHog = usePostHog();
+  const dimensions = useWindowDimensions();
 
   let content: React.JSX.Element;
   let isMediaSupported = true;
 
   if (item.type === MediaType.Image) {
-    content = <ImageView item={item} nativeGesture={nativeGesture} fullScreenWidth={SCREEN_WIDTH} />;
+    content = <ImageView item={item} nativeGesture={nativeGesture} fullScreenWidth={dimensions.width} />;
   } else if (item.type === MediaType.Video) {
     content = <WebVideoView item={item} isVisible={isVisible} />;
+  } else if (item.type === MediaType.PDF) {
+    content = <PDFView item={item} />;
   } else {
     isMediaSupported = false;
     content = <BodySm>{'Unsupported Media Type'}</BodySm>;
@@ -39,5 +39,5 @@ export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({it
     }
   }, [postHog, isMediaSupported, item]);
 
-  return <View style={{width: SCREEN_WIDTH, height: SCREEN_HEIGHT}}>{content}</View>;
+  return <View style={{width: dimensions.width, flex: 1}}>{content}</View>;
 };
--- a/components/content/carousel/MediaViewerModal/WebMediaView.tsx
+++ b/components/content/carousel/MediaViewerModal/WebMediaView.tsx
@@ -1,0 +1,104 @@
+import {onlineManager} from '@tanstack/react-query';
+import {MediaLoadErrorView} from 'components/content/carousel/MediaViewerModal/MediaLoadErrorView';
+import {isOfflineErrorCode} from 'components/content/carousel/MediaViewerModal/webViewOfflineError';
+import {View} from 'components/core';
+import React, {Ref, useCallback, useEffect, useImperativeHandle, useRef, useState} from 'react';
+import {ActivityIndicator, LayoutChangeEvent, StyleSheet} from 'react-native';
+import WebView from 'react-native-webview';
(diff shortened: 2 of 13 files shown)


components/content/carousel/MediaViewerModal/webViewOfflineError.ts
```diff
@@ -0,0 +1,8 @@
+import {Platform} from 'react-native';
+
+const ANDROID_OFFLINE_ERROR_CODE = -2;
+const IOS_OFFLINE_ERROR_CODE = -1009;
+
+export const isOfflineErrorCode = (errorCode: number) => {
+  return (Platform.OS === 'android' && errorCode == ANDROID_OFFLINE_ERROR_CODE) || (Platform.OS === 'ios' && errorCode == IOS_OFFLINE_ERROR_CODE);
```
yuliadub: he == vs === caught my eye here, at first I thought it was maybe because errorCode is a str, but looks like its a number so why not use ===?

components/content/carousel/MediaViewerModal/WebVideoView.tsx
```diff
@@ -127,34 +78,14 @@ export const WebVideoView: React.FunctionComponent<WebVideoViewProps> = ({item,
...
-      bounces={false}
-      style={{maxHeight: maxHeight, transform: [{translateY: yOffset}], backgroundColor: colorLookup('modal.background')}}
+    <WebMediaView
+      ref={handleRef}
       source={sourceData}
-      renderLoading={onRenderLoading}
-      onError={onError}
-      startInLoadingState
+      heightFraction={0.33}
+      errorMessage="An error occured loading the video. Please try again"
```
yuliadub: should be "occurred"

components/content/carousel/ThumbnailList.tsx
```diff
@@ -29,31 +32,40 @@ const thumbnailListItems = (mediaItems: MediaItem[]): ThumbnailListItem[] => {
...
     title: item.title,
   };
 };
 
+export const pdfToThumbnailListItem = (_item: PDFMediaItem): ThumbnailListItem => {
+  return {
+    kind: 'pdf',
+    uri: '',
+    caption: null,
+    title: undefined,
```
yuliadub: do we ever expect that pdfs will get a caption/title?

components/content/carousel/MediaViewerModal/WebMediaView.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    return onlineManager.subscribe(() => {
+      if (loadError && isOfflineErrorCode(loadError.nativeEvent.code) && onlineManager.isOnline()) {
+        webRef.current?.reload();
+        setLoadError(null);
+      }
+    });
+  }, [loadError]);
+
+  const maxViewHeight = containerHeight * heightFraction;
```
yuliadub, line 62: not sure how much this matters, but containerHeight will be briefly 0 (its initial state before it gets set), could cause a weird ui render but if you havent spotted anything maybe this is a non issue?

components/content/carousel/MediaViewerModal/MediaContentView.tsx
```diff
@@ -39,5 +39,5 @@ export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({it
     }
   }, [postHog, isMediaSupported, item]);
 
-  return <View style={{width: SCREEN_WIDTH, height: SCREEN_HEIGHT}}>{content}</View>;
+  return <View style={{width: dimensions.width, flex: 1}}>{content}</View>;
```
yuliadub, line 42: is there a specific reason for dropping the set height here? flex:1 could mislead us to potentially end up with a deformed image

components/content/carousel/MediaViewerModal/MediaContentView.tsx
```diff
@@ -20,14 +17,17 @@ interface MediaContentProps {
 
 export const MediaContentView: React.FunctionComponent<MediaContentProps> = ({item, isVisible, nativeGesture}) => {
   const postHog = usePostHog();
+  const dimensions = useWindowDimensions();
```
yuliadub, line 20: curious why you went with this change vs keeping the Dimensions.get('screen')?

components/content/carousel/MediaViewerModal/webViewOfflineError.ts
kevinherdez: I think that was oversight on my part

components/content/carousel/ThumbnailList.tsx
kevinherdez: No, the `pdfMediaSchema` only has `type` and `url:{original: string}` properties defined on it. The `title` and `caption` properties in the `ThumbnailList` are holdovers from before the `MediaViewer` rewrite. There's not used in the list itself anymore, so I'll go ahead and remove them from the `ThumbnailListItem` definition

components/content/carousel/MediaViewerModal/WebMediaView.tsx
kevinherdez: I can put a comment in there pointing that out, but I haven't seen any issues with it so it could be fine

components/content/carousel/MediaViewerModal/MediaContentView.tsx
kevinherdez: Not really a specific reason. I didn't love that that we were passing in a set height/width to get the images to render correctly since all I'm trying to do here is make the view fullscreen. This seemed like a more flexible approach, but I don't mind changing it back

components/content/carousel/MediaViewerModal/MediaContentView.tsx
kevinherdez: This is potentially a more "correct" approach that I didn't know about when I first wrote the component. The hook will report changes to the window dimensions available to the app which is more important for apps that support rotation/foldable screens. `Dimensions.get('screen')` doesn't update once it returns the values. For our purposes, there's really no difference. This just feels cleaner to use the hook rather than the static constant

components/content/carousel/ThumbnailList.tsx
kevinherdez: They are being used in `MediaPreview` so I'm going to change the object type since it doesn't make a lot of sense to reuse the `ThumbnailListItem` for that component

Issue #1171, opened 2026-04-14
Improve Animation for Navigation Banner


kevinherdez:
Fade out or swipe up the animation for the `ForecastMapNavigationHeader` when transitioning from the Center State to the No Center State and vice versa

kevinherdez:
This is in Preview v80

Issue #1151, opened 2026-03-26
Create a popup for centers that are not opted into Avy


rustynwac:
There's a handful of centers not using Avy. A couple in AK + Gallatin, Utah, Colorado, Canada. We want to redirect to their sites rather than surface them in the app. We need a nice way to do that with some messaging like "This avalanche center is not optimized to display data within Avy. Visit the website here"

kevinherdez:
Should the 6 centers that are on Avy, but are not supported in production (Haines, Gallatin, Kachina Peaks, etc.) be shown the same experience?

rustynwac:
> Should the 6 centers that are on Avy, but are not supported in production (Haines, Gallatin, Kachina Peaks, etc.) be shown the same experience?

I'll make sure they get on for this year.

kevinherdez:
This is in Preview v80

Issue #1121, opened 2026-03-06
Create a first run experience


kevinherdez:
- Show a pop up after initial avalanche selection explaining the new UI

kevinherdez:
This is in Preview v80

PR #1180, opened 2026-05-05
Map Project Improvements

### LINKED ISSUE #1177: Observation page moves around in x and y axis when scrolling (in Preview 62)
One of my testers got this. Waiting to hear what type of iOS device.

Image
Image
Image
Image

Avy version 1.0.0 (62)
Git revision f6c29ec7ad8e91944da1c12eb3cd367f0b5bc994
Update group ID e (channel: preview)
Update ID b72685d0-d528-4fc5-9ccb-19ed61ffe347 (platform: ios)
User ID 2b9cfb00-cee9-4fc2-a9e1-ff5155182239

kevinherdez:
This PR fixes bugs, introduces better UI, and improves the no center experience architecture. I tried to keep all of the commits focused in what they address and the rest of the description goes in detail about what each commit does to hopefully make it easier to review.

## Commit 1: Update UI to better match designs

- New logos were added from the figma and introduced in the `NoCenterDrawerHeader` and the `ForecastMapNavigationHeader`
   - #1168 
- Introduced a new CenterFocusedDrawerHeader that matches the figma on how the drawer should look
    -  - #1168 
- Changed the animation header when going into/out of the no center experience. Now the center focused header sits on top of the non-branded header and slides up when going into the no center experience and slide down when focusing on a center. The branded header now uses the center display ID
    - #1171
    
|Drawer Before|Drawer After|
|---|---|
|Screenshot 2026-05-05 at 10 01 58 AM|Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 15 12 31|

Header animation
https://github.com/user-attachments/assets/f2057661-68d9-4a82-b7c2-465bb6baa7b6

## Commit 2: Put isInNoCenterExperience and lastCameraStop in their own context

- Introduces the `MapPersistence` that saves the `isInNoCenterExperience` and `lastCameraStop` outside of `preferences`. The benefit of this is that only the views that need to update the map experience need to pull this in and it would trigger refreshes in views that rely on `usePreferences`
- Fixes #1177 by locking the `ScrollView` direction. This issue was mainly an iOS issue where `UIScrollView` is by-default orthogonal so if the view, for some reason, calculated the width to be bigger than the screen, then we would hit this issue

## Commit 3: Add FRE and Center Not Supported Modals

- Introduces the First Run Experience explaining the map project that shows on first launch after a user selects a center
    - #1121 
- Introduces the `CenterNotSupportedModal` that shows when a user taps on a center not on the NAC. The modal has a button that will take the user to that center's forecast page.
    - #1151

|FRE|
|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 15 13 20|

|Center not supported modal|
|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 14 58 51|

## Commit 4: Fix for map recentering

On certain devices a race condition was happening between going into the no center experience and the `AnimatedMapWithDrawerController` triggering an `animateToRegion` which would bring the user back to the center they were just at

## Commit 5: Address issues found by Claude

Claude did a PR review locally and these were the issues fixed as a result of that


components/map/AnimatedCards.tsx
```diff
@@ -221,11 +222,22 @@ export class AnimatedMapWithDrawerController {
     this.animateMapRegion();
   }
 
+  // When in the no-center experience the map should remain wherever the user has panned/zoomed it.
+  // Suppressing here is the single choke point that covers every caller of animateMapRegion (setState, layout-driven calls, focus effects, etc.)
+  setSuppressMapCentering(suppress: boolean) {
+    if (this.suppressMapCentering === suppress) return;
+    this.suppressMapCentering = suppress;
+    if (suppress) {
+      this.animateMapRegion.cancel();
```
stevekuznetsov, line 231: I was not expecting a setter to actually call this - maybe update the method name to be a bit more clear?

Issue #1150, opened 2026-03-25
CAIC vs CBAC


rustynwac:
They forecast for the same area. How do we display both if we need to?

kevinherdez:
I'm going to go ahead and close this issue. CAIC is being shown on the map with CBAC. If we need to change the design, we can either reopen or create a new issue

kevinherdez:
This is still an ongoing issue as the original design is being changed to better reflect the preferences of both CAIC and CBAC. Reopening

rustynwac:
Design [here](https://www.figma.com/design/Bai2w7x0vUakL5WRjPGRg7/AVY-Native-App?node-id=4193-671&p=f&t=EzZqphw3YNQ1g050-0)

PR #1185, opened 2026-05-19
Show CAIC on the map

kevinherdez:
- #1150

Adds CAIC onto the map and carves the CBAC polygons out of it. `@turf` provides a `difference` function that we use to modify the CAIC polygons so that they don't include the CBAC ones. This prevents the polygons from overlapping.

|No center experience|CBAC centered|
|---|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-05-19 at 15 53 58|Simulator Screenshot - iPhone 17 Pro - 2026-05-19 at 15 53 47|


busbyk (approved):
This is awesome. `turf` is such a cool package. 

I tested this out on ios simulator and it's working as expected for me. I like the tests and I like the approach. 

I also had Claude do a code review and I like it's suggestions - all optional but worth considering: 

1. Comment is now stale — the old comment at line 84 said "CAIC is unsupported in the app." That's removed, which is correct, but there's no replacement comment explaining why carving is needed. A one-liner like // Carve CBAC zones out of CAIC so polygons don't overlap on   
the map would help future readers of useAllMapLayers.ts.
2. The @turf/helpers version mismatch — @turf/difference and @turf/union pull in @turf/helpers@7.3.5, while the project's direct dependency is @turf/helpers@^7.3.2. This works fine (compatible range), but you now have two copies of @turf/helpers in the lockfile (7.3.2 and   
7.3.5). Consider bumping the direct dep to ^7.3.5 to deduplicate, or it'll sort itself out on the next broad upgrade.                                                                                                                                                              
3. Missing test for error paths — the catch blocks and the union returning null path aren't tested. These are defensive and unlikely to fire, but if you want full confidence in the fallback-to-uncarved behavior, a test that mocks @turf/difference to throw would cover it. Not blocking, just a gap to be aware of.                                                                                                                                                                                                                                              

Good stuff!

yuliadub (approved):

PR #1191, opened 2026-06-04
Undo Xcode Image EAS

kevinherdez:
Undoes PR #1176 now that the project is on Expo 56. The original changes was needed because we had to build the project on Xcode 26+ but Expo 53 defaulted to Xcode 18 on EAS. Expo 56 defaults to Xcode 26.4 on EAS

### DIFF
--- a/eas.json
+++ b/eas.json
@@ -32,19 +32,13 @@
     "preview": {
       "channel": "preview",
       "autoIncrement": true,
-      "ios": {
-        "image": "macos-sequoia-15.6-xcode-26.2"
-      },
       "env": {
         "APP_VARIANT": "preview"
       }
     },
     "release": {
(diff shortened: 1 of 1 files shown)


eas.json
```diff
@@ -32,19 +32,13 @@
...
         "APP_VARIANT": "preview"
       }
     },
     "release": {
       "channel": "release",
-      "autoIncrement": true,
-      "ios": {
-        "image": "macos-sequoia-15.6-xcode-26.2"
-      }
+      "autoIncrement": true
```
yuliadub, line 41: @kevinherdez this is missing a comma that was there before (but maybe it doesnt matter)

yuliadub: looking at it closer it does not, ignore me 😄

Issue #1186, opened 2026-05-19
Fix Time Machine


kevinherdez:
We need to fix the time machine functionality so that the map layer call uses it. Right now, only the forecast fetching is respecting the time machine date

kevinherdez:
This is in Preview

PR #1192, opened 2026-06-04
Pipe requestedTime to useAllMapLayers

kevinherdez:
- #1186 

Pipes `requestedTime` to `useAllMapLayers` so that when the `requestedTime` is changed through the `TimeMachine` the map will show older data so we can better test the new Map experience


yuliadub (approved):
🚢

PR #1204, opened 2026-07-02
Replace react-native-toast-view with custom toast

kevinherdez:
Rusty reported an issue on Android with Expo 56 that the expired forecast toast doesn't seem to respond to taps and the animation is laggy. The new architecture for React Native introduced a new UI rendering framework, Fabric, that changes how RN views are translated to native and these differ between iOS and Android. Since `react-native-toast-view` hasn't been updated in a year I decided to create our own custom toast view that mimics what we had and addressing these issues.

The two big design descisions with the toast is that it used `React Native Gesture Handler` instead of `Responders` so that it captures touches better without blocking other touches, and it **does not** use `react-native-reanimated`. The reason being is that Reanimated commits the animation to Android's ShadowTree and on the JS thread instead of the main UI thread like the legacy `Animated` library from React Native does.

I also changed the `ForecastScreen` to lazily load the tabs so that it feels more responsive on Android, with the tradeoff that the weather tab and observations tab need to fetch their information on the first load so users will see a loading spinner. Pushing and popping on the nav stack on Android is more expensive than on iOS so this greatly helps performance. However, there's an argument to be made that we only do this on Android.

Before video:
https://github.com/user-attachments/assets/2b06fb15-d2d3-486e-a393-18ae54542378

After video:
https://github.com/user-attachments/assets/ec83afc6-565f-4e98-a950-f39176152d06


rustynwac:
The after vid certainly looks better to me!

Issue #1205, opened 2026-07-03
Add title sponsor Nokian Tyres!


rustynwac:
Add logo to splash page and pop up modal to settings menu. Copy being confirmed by Nokian now. 
Designs [here](https://claude.ai/design/p/019dfe3a-f60e-76a8-96fd-1c08e135ddbb?file=Refined+Direction.html)
Logos [here](https://company.nokiantyres.com/news-and-media/media-bank/)

<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/03ece86a-d0c9-4b33-b7dc-bd1a6c1c1e1c" />
<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/63a8c89c-ac99-4e21-aa42-a2d29ec0346a" />
<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/790d7379-21ca-438d-b115-11c90e20af19" />

rustynwac:
One thing to note, would be if we could update the URL? It actually went to our global webpage, our local webpage is [na.nokiantyres.com](http://na.nokiantyres.com/). For tracking purposes, if you could use this: https://na.nokiantyres.com/?utm_medium=referral&utm_source=3rd%20Party&utm_campaign=Nokian_Tyres_Avy

kevinherdez:
Perfect I'll get that changed and out for a PR this week

kevinherdez:
@rustynwac what do you think Nokian would want to see for the splash screen? Right now, I created a mirrored splash that does a little animation showing the sponsor. In the video the Nokian splash screen is a little quick, but we can adjust that.

https://github.com/user-attachments/assets/49a8adb4-86e2-4f2c-9815-835e5dba3091

We can also add this to the static splash screen that is created for us by Expo. However, there is a limitation where we can only provide Expo with 1 image asset and a color to create the splash screen which means that we would need to create a custom image assets showing the Avy app logo, the divider, and the Nokian sponsor all in one image (highlighted below).

<img width="230" height="500" alt="Image" src="https://github.com/user-attachments/assets/84f2ca41-0b73-41d5-9563-fe4632d056cc" />


Just let me know what they would prefer

rustynwac:
Honestly, whatever is easy will work great (easy to revert back too if they drop and we dont have a sponsor). The animation draws more attention to it which I like. As long as that doesn't bog down old phones!

kevinherdez:
[...]

PR #1219, opened 2026-08-06
Add Title Sponsor Nokian to Avy

kevinherdez:
- #1205 

The Nokian Tyre sponsorship is being added to 3 areas: the splash screen, the Drawer, and the FRE

## Splash Screen
The `splash.png` file was removed in favor of using `avy-logo-transparent.png` for the splash screen. The main reason for this is that full screen images were deprecated for splash screens, and, also, the smaller asset allows us to have more control over the image.

The new flow for the splash screen is that we're now always showing the "mirrored" splash screen view that we used to show only when load up was taking a long time. The view was broken out into the `SponsorSplashScreen` file. `SponsorSplashScreen` shows the Avy logo in the same exact spot as the built in splash screen, and animates in the Nokian logo.

This view was added directly into the `BaseAppView` to allow for a smooth transition between the `SponsorSplashScreen` and the app's content. 

`SplashScreenContext` was added to prevent the `AvalancheCenterSelectorModal` from showing while the `SponsorSplashScreen` was still visible.

Demo: https://github.com/user-attachments/assets/0af7eada-24d0-4c3f-a30f-1444d4597683

## Drawer

A new sponsor section was added to the top of the `Drawer` that predominantly shows Nokian as our sponsor. When a user taps on the section it opens the new `DrawerModal` to explain more about the partnership and to give users the option to go to the Nokian site using our campaign URL.

|Drawer Sponsorship|Drawer Modal|
|---|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-08-06 at 10 45 30|Simulator Screenshot - iPhone 17 Pro - 2026-08-06 at 10 45 32|

## FRE
The Nokian logo and an extra bullet point was added to the FRE that we're going to show with the release of the map project
Simulator Screenshot - iPhone 17 Pro - 2026-08-06 at 10 42 38

---
<a href="https://app.blacksmith.sh/NWACus/codesmith/avy/pr/1219"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788631740&installation_model_id=426131&pr_number=1219&repository=NWACus%2Favy&return_to=https%3A%2F%2Fgithub.com%2FNWACus%2Favy%2Fpull%2F1219&signature=3132b5bb439b1bf445f764ad2bacda8f575d00debffbdebf77975dad6172ca77"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


components/modals/FirstRunExperienceModal.tsx
```diff
@@ -28,8 +28,13 @@ export const FirstRunExperienceModal: React.FC<FirstRunExperienceModalProps> = (
...
         <View style={{backgroundColor: colorLookup('white'), borderRadius: 16, padding: 24, width: '100%', maxWidth: 340}}>
-          {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports*/}
-          <Image source={require('assets/avy-logo-transparent.png')} resizeMode="contain" style={{height: 90, width: '100%', alignSelf: 'center', paddingBottom: 12}} />
+          <HStack space={12} alignItems="center" justifyContent="center" paddingBottom={12}>
+            {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports*/}
+            <Image source={require('assets/avy-logo-transparent.png')} resizeMode="contain" style={styles.avyLogo} />
+            <Title3Black color={colorLookup('text.secondary')}>+</Title3Black>
+            {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports*/}
+            <Image source={require('assets/logos/Nokian_Tyres_Logo.jpg')} resizeMode="contain" style={styles.sponsorLogo} />
```
yuliadub: we have both jpg and png (and we are using jpg and not png it looks like), historically png is better to use, but since we dont support dark mode this wont be an issue for now most likely

components/screens/navigation/Drawer.tsx
```diff
@@ -185,6 +201,81 @@ const DrawerMenu: React.FunctionComponent<DrawerMenuProps> = ({navigation, avala
...
+        </VStack>
+      </TouchableOpacity>
+    </VStack>
+  );
+};
+
+const NOKIAN_TYRES_URL = 'https://na.nokiantyres.com/?utm_medium=referral&utm_source=3rd%20Party&utm_campaign=Nokian_Tyres_Avy';
+const SponsorDrawer: React.FunctionComponent<{visible: boolean; onDismiss: () => void}> = ({visible, onDismiss}) => {
+  const visitNokianTyres = useCallback(() => {
+    void WebBrowser.openBrowserAsync(NOKIAN_TYRES_URL);
```
yuliadub: maybe lets catch and show an error in case we cant open the browser for any reason?

components/screens/navigation/Drawer.tsx
```diff
@@ -185,6 +201,81 @@ const DrawerMenu: React.FunctionComponent<DrawerMenuProps> = ({navigation, avala
...
+            {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports */}
+            <Image source={require('assets/logos/Nokian_Tyres_Logo.jpg')} style={styles.logo} />
+          </Center>
+        </VStack>
+      </TouchableOpacity>
+    </VStack>
+  );
+};
+
+const NOKIAN_TYRES_URL = 'https://na.nokiantyres.com/?utm_medium=referral&utm_source=3rd%20Party&utm_campaign=Nokian_Tyres_Avy';
```
yuliadub: do we have a better place to put this URL so it doesnt get lost? maybe makes sense to make a sponsor.ts in data that would look something like this (if we add new sponsors this hopefully makes it easier too)

```ts
// data/sponsors.ts
export interface Sponsor {
  id: string;
  displayName: string;
  campaignUrl: string;
  // eslint-disable-next-line @typescript-eslint/no-require-imports
  logo: number;        // require('assets/logos/…')
  logoSnow: number;    // white-on-dark variant for splash
  logoAspectRatio: number; // width / height, drops the 416/1024 magic numbers
  drawerBlurb: string;
  splashBlurb: string; // "IN PARTNERSHIP WITH"
}

export const TITLE_SPONSOR: Sponsor = {
  id: 'nokian-tyres',
  displayName: 'Nokian Tyres',
  campaignUrl:
    'https://na.nokiantyres.com/?utm_medium=referral&utm_source=3rd%20Party&utm_campaign=Nokian_Tyres_Avy',
  logo: require('assets/logos/Nokian_Tyres_Logo.jpg'),
  logoSnow: require('assets/logos/Nokian_Tyres_Snow.png'),
  logoAspectRatio: 1024 / 416,
  drawerBlurb:
    'Nokian Tyres invented the winter tire in 1934…',
  splashBlurb: 'IN PARTNERSHIP WITH',
};
```

components/SponsorSplashScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+    logoTranslateY.value = withTiming(0, {duration: LOGO_SETTLE_DURATION_MS, easing: Easing.out(Easing.cubic)});
+    contentOpacity.value = withDelay(CONTENT_FADE_DELAY_MS, withTiming(1, {duration: CONTENT_FADE_DURATION_MS}));
+  }, [active, logoTranslateY, contentOpacity]);
+
+  useEffect(() => {
+    if (!dismissed) {
+      return;
+    }
+    opacity.value = withTiming(0, {duration: FADE_OUT_DURATION_MS});
+    const timeout = setTimeout(handleFadeComplete, FADE_OUT_DURATION_MS);
```
yuliadub: Have not verified this - so apply as you see this. 

claude tells me that setTimeout may cause a delay on slower devices where after you tap out of the sponsor logo, your first click for a zone may not go through. It suggests this instead

```ts
useEffect(() => {
  if (!dismissed) return;
  opacity.value = withTiming(
    0,
    {duration: FADE_OUT_DURATION_MS},
    (finished) => {
      if (finished) runOnJS(handleFadeComplete)();
    },
  );
}, [dismissed, opacity, handleFadeComplete]);
```

also it says to add pointerEvents so even during the fade the view lets touches through.

```ts
<Animated.View
  pointerEvents={dismissed ? 'none' : 'auto'}
  style={[StyleSheet.absoluteFill, {backgroundColor: SPLASH_BACKGROUND_COLOR}, rootStyle]}
>
```

yuliadub (approved):

components/modals/FirstRunExperienceModal.tsx
kevinherdez: Good point. I just downloaded their black and white logos from the Nokian asset catalog that they have. I'll see if there's a PNG for the black text one.

components/screens/navigation/Drawer.tsx
kevinherdez: Great point

components/SponsorSplashScreen.tsx
kevinherdez: Nice I removed the pointer events so that the touches wouldn't go through the new splash screen but this is better

components/screens/navigation/Drawer.tsx
kevinherdez: Here's what the error shown to the user will look like
Simulator Screenshot - iPhone 17 Pro - 2026-08-12 at 15 59 33

components/screens/navigation/Drawer.tsx
kevinherdez: I added this in the latest commit. There are some differences in the data shape, and used Claude's suggestion of having this accessed via a hook, but it should still allow us to easily add new sponsors later

components/SponsorSplashScreen.tsx
kevinherdez: The only difference what I committed is that `runOnJS` is deprecated so the code added uses the new `scheduleOnRN`

PR #1238, opened 2026-09-15
Update metrics to enforce passing centerId

### LINKED ISSUE #1167: Improve Telemetry Tracking
We have some blind spots in out event tracking that we should fix. A couple that I can think of are

- When a user presses the Submit Obs button
- When a user presses the Share button
- When someone switches a center

@rustynwac please feel free to add events and I'll add them all on 1 PR

kevinherdez:
This PR forces each `capture` and `screen` call to have a `center` property to be inline with our new shared conventions with AvyWeb. Because AvyApp supports switching centers in the same product, we cannot add the `center` as a super property that's included in the `register` call

## Main Change
`useAnalytics` was updated to include new `CenteredEventProperties` and `CenterlessEventProperties` types. By adding these new types, the compiler can make sure that `screen` and `capture` calls includes properties that contain a `center` key, and `register` and `identity` calls explicitly do not contain a `center` key. The `CenterlessEventProperties` type is more to prevent a future regression

## Additional changes
The majority of the changes are updates to our `capture` and `screen` calls to include the `center` property as well as typos in some of the event names.

## Open Questions
@rustynwac our `screen` calls capture the screen name in camel case while the new convention is to use snake case (ex, `avalancheForecastMap` instead of `avalanche_forecast_map`). Should we change these to match convention, or should we leave these alone so that we can more easily compare new data with data from past seasons? It doesn't necessarily have to be included in this PR, but we should decide what we want to do for the upcoming season.

---
<a href="https://app.blacksmith.sh/NWACus/codesmith/avy/pr/1238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792090749&installation_model_id=426131&pr_number=1238&repository=NWACus%2Favy&return_to=https%3A%2F%2Fgithub.com%2FNWACus%2Favy%2Fpull%2F1238&signature=10507a434502d3b6d611a4999ac6343261171e5a18e367bca835fb3cc925825b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


hooks/useAnalytics.ts
```diff
@@ -31,29 +31,35 @@ export const getPresentedFromForAnalytics = (navigation: NavigationWithState): s
...
-  capture: (...args) => {
-    postHog?.capture(...args);
+  screen: (name, properties, options) => fireAndForget(postHog?.screen(name, properties, options), `failed to capture ${name} screen view`, logger),
+  register: properties => fireAndForget(postHog?.register(properties), 'failed to register analytics super properties', logger),
+  capture: (event, properties, options) => {
+    postHog?.capture(event, properties, options);
   },
   captureCenterSwitch: (switchedFrom, switchedTo, origin) => {
-    postHog?.capture('center_switched', {switchedFrom: switchedFrom, switchedTo: switchedTo, eventOrigin: origin});
+    postHog?.capture('center_switched', {center: switchedFrom, switched_from: switchedFrom, switched_to: switchedTo, event_origin: origin});
```
yuliadub, line 59: Should center be switchedTo here since that would in theory be the active center at the time of the event? (as I say this either way thought seems confusing 😄, but since both centers are preserved in the event anyways maybe just a comment with the logic in case we forget why we decided one way or another)

yuliadub (approved):

hooks/useAnalytics.ts
kevinherdez: Yeah I went back and forth on that one for awhile, and I agree it really could go either way. @rustynwac do you feel strongly either way about whether this event should be attributed to the center where the switch came from or attributed to the center that the user just switched to?

Either way I'll add a comment explaining it.

rustynwac: I don't think I have a strong preference or maybe understand the context it'll be used in

rustynwac:
I think it's ok to change moving forward. We haven't done a ton of year over year comparison anyway

PR #1241, opened 2026-09-17
Add Avlanche CAN to the map

### LINKED ISSUE #1235: Add Canada to map view
With the bump out to their website like UAC + CAIC

kevinherdez:
This adds the Avalanche Canada polygons to AvyApp and treats it as an unsupported center

|AvCAN on the Map|
|---|
|Simulator Screenshot - iPhone 17 Pro - 2026-09-16 at 16 08 07|

## Main Changes
Three new hooks were added to fetch the information: `useCanadaMapLayer`, `useCanadaMetadata`, and `useCanadaZones`. `useCanadaZones` is what `AvalancheForecastMapView` uses to display the polygons by wrapping the necessary hooks.

A new polygon view was added to support the Canada zone information: `CanadaAvalancheForecastZonePolygons`

Because Haines Avalanche Center overlaps with part of Avalanche Canada we need to carve out the Haines zone like we do with CBAC and CAIC. A helper function, `canadaMapViewZones.ts`, was added and is called from `useCanadaZones` to carve out the zones.

Fetching and displaying the Canada zones was written to prioritize stability between renders so that we wouldn't unnecessarily have to recalculate carving out the HAC zone from Avalanche Canada

---
<a href="https://app.blacksmith.sh/NWACus/codesmith/avy/pr/1241"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792195352&installation_model_id=426131&pr_number=1241&repository=NWACus%2Favy&return_to=https%3A%2F%2Fgithub.com%2FNWACus%2Favy%2Fpull%2F1241&signature=0a0ca1f0224b33e2a5715ce4644b705c356314851ee8cc45024b8928f0eb8929"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


yuliadub:
2 comments

- do we want time machine work with Canada? (possible im missing it, but seems like right now it wont) could be good for testing warnings and other possible things we may want to show just on the map layer that may be different for canada vs US?

- I like that Canada uses a separate `CanadaMapViewZone` without requiring a US-style `center_id`, since Canadian forecasts are primarily organized around zones. One thing I'm wondering - would it be worth it to avoid representing `CAN` as an unsupported center in the tap, modal, and analytics paths, and instead identify the country/source and Canadian `zone_id` explicitly? This would keep the UI language more canada specific (i.e.  “This avalanche center” can instead say “This forecast is not available within Avy.”) and give us room for canadian zones to evolve independently without forcing them into the US center model in case we do one day show those forecasts in avy and they were to differ in any specific way. (one specific analytics case I can think of right now could be if we wanted to track who clicked on which specific zone on the map or something like that)

components/modals/CenterNotSupportedModal.tsx
```diff
@@ -1,27 +1,46 @@
...
+import {logger} from 'logger';
 import {colorLookup} from 'theme';
-import {AvalancheCenterID, AvalancheCenterWebsites} from 'types/nationalAvalancheCenter';
+import {AvalancheCenterID} from 'types/nationalAvalancheCenter';
 
 interface CenterNotSupportedModalProps {
   visible: boolean;
-  centerId: AvalancheCenterID | null;
+  centerId: AvalancheCenterID;
+  unsupportedCenterId: AvalancheCenterID | 'CAN' | null;
```
yuliadub: possible worth it to throw this string in a const? i'll let you decide

kevinherdez: With the other feedback, it makes sense to add this somewhere more formally so that it's already in place if we support Canada in the future

kevinherdez:
> 2 comments
> 
> * do we want time machine work with Canada? (possible im missing it, but seems like right now it wont) could be good for testing warnings and other possible things we may want to show just on the map layer that may be different for canada vs US?

Good point! They didn't lay it out in the API on how to do that so I skipped adding it, but I found the spot on their website where you can access it and figured it out. I'll get that added.

> * I like that Canada uses a separate `CanadaMapViewZone` without requiring a US-style `center_id`, since Canadian forecasts are primarily organized around zones. One thing I'm wondering - would it be worth it to avoid representing `CAN` as an unsupported center in the tap, modal, and analytics paths, and instead identify the country/source and Canadian `zone_id` explicitly? This would keep the UI language more canada specific (i.e.  “This avalanche center” can instead say “This forecast is not available within Avy.”) and give us room for canadian zones to evolve independently without forcing them into the US center model in case we do one day show those forecasts in avy and they were to differ in any specific way. (one specific analytics case I can think of right now could be if we wanted to track who clicked on which specific zone on the map or something like that)

I'm not exactly sure what you mean by
> and instead identify the country/source and Canadian `zone_id` explicitly

Is that in reference to how we know whether a user tapped on a Canada zone vs a zone coming from the NAC's `map-layer` call in general or just for analytics? I think we have to treat Canada similarly to how we treat the other unsupported centers, but we can deviate on what that means in terms of the language used in the modal and data we capture through analytics. That doesn't necessarily mean we need to use the `CAN` "center id" like I have in the PR, but with the app being so `center id` focused, there might be a bigger architecture change that would be worth investigating to make it easier to support Canada without pigeon-holing them into NAC data concepts

I need to add analytics for the taps on the unsupported centers which I'll do on this PR, and to your point it could make sense to keep those as separate events depending on if Canada was tapped or not.

